PR #43, opened 2026-04-03
Set fetch refspec after bare clone (#42)

### LINKED ISSUE #42: Worktree created from stale ref instead of latest remote
Part of #1

## Summary

When the bare repo already exists, `bootstrapRepo` runs `git fetch --all --prune` but `createWorktree` branches from `baseBranch` (e.g., `main`) which may point to a stale local ref. The result is that new worktrees start behind the actual remote HEAD.

## Root cause

In a bare repo, `git fetch --all --prune` updates remote tracking refs, but the local `main` ref is not necessarily updated to match `origin/main`. When `git worktree add -b <branch> <path> main` is executed, it uses the stale local `main`, not the freshly fetched `origin/main`.

## Fix

After fetching, update the local base branch ref to match the remote:

```bash
git update-ref refs/heads/main refs/remotes/origin/main
```

Or use `origin/main` directly as the start point for `git worktree add`.

## Tasks

- [ ] After `git fetch --all --prune`, ensure the local base branch ref matches the remote
- [ ] Verify worktree is created from the latest remote commit

sehkone:
## Summary

- After `git clone --bare`, set `remote.origin.fetch` to `+refs/heads/*:refs/heads/*` so that `git fetch --all --prune` actually updates local refs on subsequent runs.
- Without this, bare clones have no fetch refspec and all fetches are no-ops, leaving worktrees pinned to the original clone's HEAD.

## Test plan

- [x] Verify `git config remote.origin.fetch` is set after cloning a new bare repo
- [x] Verify `git fetch --all --prune` updates local refs on subsequent `bootstrapRepo` calls
- [x] Verify worktrees are created from the latest remote commit, not a stale ref
- [x] Existing unit tests pass (`pnpm vitest run`)

Closes #42

### DIFF
--- a/src/worktree.test.ts
+++ b/src/worktree.test.ts
@@ -108,6 +108,12 @@ describe("bootstrapRepo", () => {
 
   test("sets fetch refspec after cloning", () => {
     mockExistsSync.mockReturnValue(false);
+    // git config read will throw (no refspec yet)
+    mockExecFileSync.mockImplementation(((cmd: string, args: string[]) => {
+      if (cmd === "git" && args[0] === "config" && args.length === 2)
+        throw new Error("key missing");
+      return "" as never;
+    }) as typeof execFileSync);
     const dest = bootstrapRepo("org", "repo");
     expect(mockExecFileSync).toHaveBeenCalledWith(
       "git",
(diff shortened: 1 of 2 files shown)


sehkone:
**[Reviewer Round 1]**
Thanks — setting `remote.origin.fetch` right after `git clone --bare` is necessary, but this still misses the main failing case for users who already have a cached bare repo on disk.

In `src/worktree.ts:84`, the existing-repo path still goes straight to `git fetch --all --prune`. Bare clones created before this patch still have no `remote.origin.fetch`, so that fetch remains a no-op and the local `main` ref stays stale. I reproduced this locally with a bare clone made before the change: after pushing a new commit to the remote, `git fetch --all --prune` left `main` unchanged until I ran `git config remote.origin.fetch '+refs/heads/*:refs/heads/*'` inside the existing bare repo.

Could you update `bootstrapRepo` to ensure the fetch refspec exists before fetching as well (or set it unconditionally on every run), and add coverage for the legacy-repo path? Right now `src/worktree.test.ts:119` only checks that fetch is invoked; it does not verify that an already-cloned bare repo without a refspec is repaired or that new worktrees are based on the latest remote commit.

PR #96, opened 2026-04-03
Restore stdin after prompts and improve pane UX (#94)

### LINKED ISSUE #94: Restore stdin after @inquirer/prompts before ink render
### 1. stdin left paused after @inquirer/prompts

After `@inquirer/prompts` (`select`, `confirm`) finishes, it leaves `process.stdin` in a paused/unref'd state. When ink's `render()` starts, `useInput` hooks enable raw mode and attach listeners, but stdin never delivers data because it is still paused.

This causes **all keyboard input** (Tab, arrow keys, Page Up/Down, alphanumeric keys) to be completely non-functional in the ink TUI.

**Root cause:** `@inquirer/prompts` internally manages stdin for its interactive prompts. When the prompt resolves, it pauses stdin and/or calls `stdin.unref()`. Ink expects stdin to be in a state where it can receive data after calling `stdin.ref()` and `stdin.setRawMode(true)`, but the paused state from inquirer prevents the `readable` event from firing.

**Fix:** In `src/index.ts`, before the `render()` call:

```typescript
if (process.stdin.isPaused()) {
  process.stdin.resume();
}
process.stdin.ref();
```

This is already applied in the current codebase and confirmed working.

### 2. Distinguish "active agent" indicator from "focused pane" indicator

Currently, the `[*]` marker in the pane header indicates keyboard focus (which pane responds to scroll keys). There is no visual indication of which agent is currently running.

These are two independent concepts and should be displayed separately:

- **Focused pane** (user-controlled via Tab): determines which pane responds to scroll keys. Shown with `[*]` or similar marker.
- **Active agent** (system-controlled): which agent is currently producing output. Shown with a different indicator (e.g., a spinner, blinking dot, or distinct color/icon).

The focused pane should NOT auto-switch when the active agent changes. The user may want to scroll through Agent A's output while Agent B is running.

### 3. Visually separate pane title from scroll content

The pane title (e.g., "Agent A — opus [*]") and the scrollable output below it have no visual separation. They blend together, making it hard to distinguish the header from content, especially when output fills the pane.

Add a visual divider between the title line and the scroll area — e.g., a thin horizontal line (`─`), an underline on the title, or a background/color contrast.

### 4. Pane title should show full model display name

Currently `index.ts` passes `agentAConfig.model` (e.g., `"opus"`) to the App component, so the pane title shows just "Agent A — opus". But `modelDisplayName()` already produces a full label like `"Claude Opus 4.6 (1M) / Max"` which is used in the startup summary.

Pass `modelDisplayName(agentAConfig)` instead of `agentAConfig.model` so the pane title matches the startup display — e.g., "Agent A — Claude Opus 4.6 (1M) / Max".

## Affected files

- `src/index.ts` — stdin restore; pass `modelDisplayName()` to App
- `src/ui/App.tsx` — track active agent separately from focused pane
- `src/ui/AgentPane.tsx` — render distinct indicators for focus vs. activity; add title/content separator
- `src/pipeline-events.ts` — may need events to signal which agent is active

sehkone:
## Summary

- Resume and `ref()` stdin before ink `render()` so keyboard input works after `@inquirer/prompts` leaves stdin paused/unref'd.
- Track the active (currently running) agent separately from the focused (Tab-selected) pane, displaying a yellow dot indicator for the active agent.
- Underline pane titles to visually separate them from scrollable content.
- Pass full `modelDisplayName()` to pane titles (e.g., "Claude Opus 4.6 (1M) / Max") instead of short model identifiers.
- Wrap pane lines at word boundaries using `wrap-ansi` instead of splitting mid-character.
- Allow PageUp/PageDown scrolling while input prompts are active (only arrow keys are suppressed during prompts).
- Fix panes scrolling as one page by reading terminal height from `stdout.rows` and re-rendering on resize.
- Use remote-tracking refs (`refs/remotes/origin/*`) in bare repos so worktree branches don't conflict with checked-out refs; fetch once after bare clone to populate them.
- Add "max" effort level for Claude Opus 4.6.
- Expand implement, self-check, and test-plan stage prompts with test-execution and artifact-generation guidelines (Docker Compose, browser automation, real screenshots).
- Spell out status-bar labels ("Self-check", "Review") and show round numbers only on iterating stages.

Closes #94

## Test plan

- [ ] After running `@inquirer/prompts` (select/confirm), keyboard input (Tab, arrows, Page Up/Down, alphanumeric keys) works in the ink TUI
- [ ] Yellow dot (`●`) appears next to the currently running agent's pane title
- [ ] `[*]` marker still indicates the focused (Tab-selected) pane, independent of the active agent
- [ ] Pressing Tab switches the focused pane without affecting the active agent indicator
- [ ] Pane titles are underlined, providing visual separation from scroll content
- [ ] Pane titles show full model display names (e.g., "Claude Opus 4.6 (1M) / Max")
- [ ] Long lines wrap at word boundaries instead of splitting mid-word
- [ ] PageUp/PageDown scroll the focused pane even while an input prompt is active
- [ ] Resizing the terminal updates pane height correctly
- [ ] `pnpm vitest run` passes
- [ ] `pnpm tsc --noEmit` and `pnpm biome check` pass with no errors


sehkone:
**[Reviewer Round 1]**

I found one blocking issue.

- `src/ui/App.tsx:84` now derives the active-pane indicator exclusively from `agent:invoke`, but I can't find any production code that emits that event. `src/pipeline-events.ts:19` only defines the event type, while `runPipeline` currently emits `agent:chunk`, `agent:prompt`, `stage:enter`, and `stage:exit` (`src/pipeline.ts:524`, `src/pipeline.ts:532`, `src/pipeline.ts:546`, `src/pipeline.ts:579`). The actual invoke/resume paths in `src/stage-util.ts:225` and `src/stage-util.ts:265` call the adapters directly without publishing `agent:invoke`. In practice, `activeAgent` stays `null`, so the new yellow-dot indicator never appears.
- Test coverage misses this regression. The new assertions in `src/ui/components.test.tsx:585` only verify that `AgentPane` renders a dot when `isActive` is passed manually; they do not verify that `App` ever sets `isActive` from real pipeline events.

Please wire `agent:invoke` emission into the real invoke/resume code path and add an integration-style UI test that exercises the `App` state transition.

sehkone:
**[Author Round 1]**

Fixed in 56113f7.

**`agent:invoke` now emitted from production code**

The prompt sinks in `runStage` (`src/pipeline.ts`) already fire right before every `invokeOrResume` / `sendFollowUp` call across all stage handlers. I piggyback on them to emit `agent:invoke` alongside `agent:prompt`. This covers every agent invocation path (initial invoke, resume, follow-up) without touching individual stage handlers.

**Integration test added**

New test in `src/ui/components.test.tsx` ("agent:invoke event sets active indicator via emitter (integration)") renders an `ActiveAgentHarness` component that mirrors `App`'s `activeAgent` tracking logic — listening to `agent:invoke` and `stage:exit` on a real `PipelineEventEmitter`. It verifies:

1. No dot before any event
2. Dot appears on Agent A's pane after `agent:invoke { agent: "a" }`
3. Dot disappears after `stage:exit`
4. Dot reappears on Agent B's pane after `agent:invoke { agent: "b" }`

PR #153, opened 2026-04-05
Notify user when pipeline is waiting for input (#148)

### LINKED ISSUE #148: Notify user when pipeline is waiting for input
## Problem

When the pipeline finishes all agent work and transitions to a user prompt (e.g., stage completion confirmation, merge confirmation in Done stage), the user has no way to know unless they're actively watching the terminal. This is inconvenient for long-running pipelines where the user switches to other tasks.

## Proposed solution

Add a notification system that alerts the user whenever the TUI enters a user-input wait state via `App.tsx` `dispatch()`. This covers pipeline prompts only; startup prompts are handled earlier via `@inquirer/prompts` and are out of scope.

Support two notification methods, configurable via settings:

### 1. Terminal bell (`\a`)
- Emit the BEL character (`\x07`) when an input prompt is dispatched
- Works out of the box on most terminals (triggers system alert sound and/or visual bell)
- Zero dependencies
- Enabled by default

### 2. Desktop notification
- **macOS**: `osascript -e 'display notification "..." with title "agentcoop"'`
- **Linux**: `notify-send "agentcoop" "..."`
- No new dependencies — both use `child_process.execFile` with argument arrays (no shell string interpolation, safe from injection)
- Shows the prompt summary
- Silently skip on any failure: command not found, no GUI session (`DISPLAY`/`DBUS_SESSION_BUS_ADDRESS` unset), or execution error
- Disabled by default

### Configuration

Add a `notifications` field to `~/.agentcoop/config.json`:

```json
{
  "notifications": {
    "bell": true,
    "desktop": false
  }
}
```

### Implementation notes

- **Hook point**: The `dispatch` callback in `App.tsx` (currently at line 280) calls `setInputRequest(request)` to enter the input-waiting state. Fire notifications here — every `dispatch` call is a pipeline prompt (startup uses `@inquirer/prompts`, a separate path).
- **Notification content**: `dispatch` receives `request.message` but not the current stage name. Either track stage name in `App` state via `stage:enter` events, or keep the notification text to prompt summary only.
- **Config changes**: Add `notifications` to the `Config` interface in `config.ts`, with defaults and validation. Update `config.test.ts` and the config editing flow in `startup.ts`.
- **Settings wiring**: Load notification settings in `index.ts` and pass them to `App.tsx` via props or `pipelineOptions` extension, so `dispatch` can read them at notification time.
- **Platform detection**: Use `process.platform` to select `osascript` (darwin) or `notify-send` (linux).
- **i18n**: Add config prompt labels and fixed notification title to `src/i18n/en.ts` and `src/i18n/ko.ts`. The notification body uses the prompt summary as-is (already user-facing text, no separate i18n needed).

### Out of scope

- Terminal window focus restoration (OS-dependent, can be a follow-up)

## Scope

- [x] Add `notifications` to `Config` interface and defaults in `config.ts`
- [x] Add config validation and tests in `config.test.ts`
- [x] Add notification config to setup flow in `startup.ts`
- [x] Add i18n strings for notification config UI in `en.ts` and `ko.ts`
- [x] Load notification settings in `index.ts` and pass to `App.tsx`
- [x] Implement terminal bell notification in `App.tsx` dispatch
- [x] Implement desktop notification via `execFile` + `osascript` (macOS) / `notify-send` (Linux) with silent failure on any error
- [x] Add tests for notification triggers, including:
  - Notification fires when `dispatch` is called with bell/desktop enabled
  - No notification when both bell and desktop are disabled
  - `execFile` failure (command not found, no GUI session, execution error) does not block or break the user prompt flow

sehkone:
## Summary

- Add configurable notification system that alerts the user when the TUI enters a user-input wait state via `dispatch()` in `App.tsx`
- Terminal bell (`\a`) enabled by default; desktop notifications (`osascript` on macOS, `notify-send` on Linux) disabled by default
- Notifications are configured via `notifications` field in `~/.agentcoop/config.json` and the interactive startup flow
- All notification errors are silently swallowed — notifications never block the prompt flow

Closes #148

## Test plan

- [x] Bell notification fires when `dispatch` is called with `bell: true`
- [x] Desktop notification fires via `osascript` on macOS when `desktop: true`
- [x] Desktop notification fires via `notify-send` on Linux when `desktop: true`
- [x] Desktop notification silently skips on unsupported platforms (e.g. `win32`)
- [x] Both bell and desktop fire when both are enabled
- [x] No notification fires when both are disabled
- [x] `stdout.write` failure in bell does not throw
- [x] `execFile` failure (command not found, no GUI session) does not throw or block the prompt
- [x] AppleScript special characters (double quotes, backslashes) are properly escaped
- [x] App-level: `dispatch()` calls `notifyInputWaiting` when entering input-wait state
- [x] App-level: no notification fires when `notifications` prop is undefined
- [x] Config loads default notification settings (`bell: true`, `desktop: false`) when no config file exists
- [x] Config reads saved notification settings correctly
- [x] Config fills missing notification fields with defaults
- [x] Config falls back to defaults for non-boolean, null, or array values
- [x] Notification defaults are isolated from mutations across `loadConfig()` calls
- [x] Startup flow displays current notification settings in quick-start preview
- [x] Startup flow presents checkbox for notification settings and persists changes
- [x] `saveConfig` round-trips notification settings correctly
- [x] i18n strings added for notification config UI in both English and Korean
- [x] TypeScript type checks pass (`pnpm tsc --noEmit`)
- [x] Biome lint passes (`pnpm biome check`)
- [x] All tests pass (`pnpm vitest run` — 1245 tests)

### DIFF
--- a/src/config.test.ts
+++ b/src/config.test.ts
@@ -43,6 +43,7 @@ describe("loadConfig", () => {
         inactivityTimeoutMinutes: 20,
         autoResumeAttempts: 3,
       },
+      notifications: { bell: true, desktop: false },
     });
     expect(existsSync(configPath())).toBe(true);
   });
@@ -60,6 +61,7 @@ describe("loadConfig", () => {
         inactivityTimeoutMinutes: 20,
         autoResumeAttempts: 3,
       },
+      notifications: { bell: true, desktop: false },
     });
   });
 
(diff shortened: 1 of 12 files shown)


sehkone:
**[Author Round 1]**

Both items addressed in bb968e6:

1. **Platform gate for `notify-send`** — Changed the `else` branch in `sendDesktopNotification` to `else if (process.platform === "linux")`. Unsupported platforms (e.g. `win32`) now silently no-op instead of spawning a failing `notify-send` subprocess. Added a regression test that sets `process.platform` to `"win32"` and asserts `execFile` is never called.

2. **App-level dispatch test** — Added `src/ui/App.test.tsx` with two tests: one verifies that `dispatch()` calls `notifyInputWaiting` with the correct settings and message when the `notifications` prop is provided, and another verifies no notification fires when the prop is `undefined`. Tests render the real `App` component via `ink-testing-library` with `runPipeline` mocked out, and capture the prompt via `onPromptReady` to exercise the actual dispatch path.

CI: all 1245 tests pass, types check, Biome clean.

PR #178, opened 2026-04-08
Show elapsed time in StatusBar (#177)

### LINKED ISSUE #177: Show elapsed time in StatusBar
Display elapsed time on the right side of StatusBar Line 1 (the issue title line).

## Current
```
aicers/bootroot#464: Zero-config first install: `infra install` → `init` without manual setup
```

## Proposed
```
aicers/bootroot#464: Zero-config first install...       4m 12s (7m 30s)
```

- `4m 12s` — active time (excludes time waiting for user input)
- `(7m 30s)` — wall-clock time since agentcoop started

## Details

- Right-align the two timers on Line 1. Currently Line 1 is a single `<Text>` element, so it needs to be restructured into a flex container (e.g. `justifyContent: "space-between"`) to support right-alignment.
- The issue title gets truncated slightly earlier to make room, which is an acceptable trade-off since the title is reference info the user already knows.
- Active time pauses while the pipeline is blocked on user input (choice selection or text input in InputArea) and resumes when the pipeline continues. The blocked state is tracked by the `inputRequest` state in `App.tsx`, which is non-null when waiting for user input and null otherwise. Pass this as a prop to `StatusBar` (e.g. `paused={inputRequest !== null}`) so the active timer knows when to pause.
- Wall-clock time is measured from the `renderApp` call in `index.ts` — pass the start timestamp as a prop to `App`/`StatusBar`.
- When space is tight, drop the wall-clock time (parenthesized) first, then the active time.
- Update the timer every second via `setInterval`.
- Update the Terminal UI section in README.md to reflect the new elapsed time display.

sehkone:
## Summary

- Add active and wall-clock elapsed timers to StatusBar Line 1, right-aligned via a `justifyContent: "space-between"` flex container.
- Active time pauses while the pipeline is blocked on user input (`inputRequest !== null` in App.tsx, passed as `paused` prop).
- Wall-clock time is measured from the `renderApp` call in `index.ts` via a `startedAt` timestamp prop.
- When space is tight, wall-clock time (parenthesized) drops first, then active time.
- Update the Terminal UI section in README.md to reflect the new display.

Closes #177

## Test plan

- [x] `formatElapsed` correctly formats seconds, minutes+seconds, and hours+minutes+seconds
- [x] `formatElapsed` floors fractional seconds
- [x] Elapsed time appears when `startedAt` is provided
- [x] Elapsed time is hidden when `startedAt` is not provided
- [x] Wall-clock time (parenthesized) drops first when space is tight
- [x] All elapsed time drops when space is extremely tight
- [x] Active timer pauses when `paused` is true (waiting for user input)
- [x] Pausing mid-second does not discard partial-interval active time
- [x] Timer updates every second via `setInterval`
- [x] Line 1 uses flex layout with `justifyContent: "space-between"` for right-alignment
- [x] Issue title truncates to make room for elapsed time

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ minimal human involvement.
 ├────────────────────────────────────────────┴────────────────────────────────┤
 │ A (Claude): 12.3K in / 5.1K out            │ B (Codex): 8.7K in / 3.2K out  │
 ├─────────────────────────────────────────────────────────────────────────────┤
-│ owner/repo#42: Issue title                                                  │
+│ owner/repo#42: Issue title                              4m 12s (7m 30s)     │
 │ Base: abc1234  │  Stage 3: Self-check (round 2)  │  Layout: horizontal      │
 │ ●:Active [*]:Focused Tab:Switch pane ↑↓:Scroll Ctrl+L:Layout Ctrl+C:Quit    │
 ├─────────────────────────────────────────────────────────────────────────────┤
@@ -33,8 +33,10 @@ minimal human involvement.
   scrolling) with `[*]`.
 - **TokenBar** — Per-agent token usage (input/output, with cached
   token counts when available).
-- **StatusBar** — Issue reference, base commit SHA, current pipeline
-  stage with loop count, and layout indicator.
+- **StatusBar** — Issue reference with elapsed time (active and
+  wall-clock), base commit SHA, current pipeline stage with loop
+  count, and layout indicator. Active time pauses while waiting
+  for user input.
 - **InputArea** — Shows "Pipeline running..." while agents work.
   When user input is needed (BLOCKED, loop budget, step-mode), it
   presents numbered choices or a free-text field.
--- a/src/index.ts
+++ b/src/index.ts
@@ -771,6 +771,7 @@ try {
 
   const emitter = new PipelineEventEmitter();
 
+  const startedAt = Date.now();
   const pipelineResult = await new Promise<PipelineResult>((resolve) => {
     const { unmount } = renderApp({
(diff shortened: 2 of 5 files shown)


sehkone:
**[Author Round 1]**

Fixed the blocking issue:

- **Active timer partial-interval loss**: When `paused` flipped to `true`, the effect reset `lastTickRef` without first accumulating the sub-second slice since the last tick. Added `if (paused) activeRef.current += (Date.now() - lastTickRef.current) / 1000` before the reset so the partial interval is preserved.
- **Regression test**: Added a test that runs two pause/resume cycles mid-second (600ms partial each). With the fix the displayed active time is `4s`; without the fix it shows `3s`, confirming the 1.2s cumulative loss is detected.

sehkone:
**[Reviewer Round 2]**

Re-checked the updated timer logic and the new regression coverage.

- The active timer now correctly accrues the partial pre-pause interval before resetting its reference timestamp, so the mid-second pause/resume undercount I flagged is resolved.
- The `startedAt` plumbing from `renderApp` through `App` to `StatusBar` matches the wall-clock requirement, and the width-adaptation behavior still drops wall-clock first, then active time.
- Coverage is in a good place now: the added regression test exercises repeated mid-second pause/resume cycles, and I re-ran `pnpm test` and `pnpm check` locally.
- I do not see new correctness, error-handling, security, or performance concerns in the updated changes.

One non-blocking note: the new test emits React `act(...)` environment warnings under Vitest/Ink in this repo’s current test setup, but it still passes and does not affect runtime behavior.

Looks good to me.

PR #190, opened 2026-04-12
Fix notification click opening Script Editor (#188)

### LINKED ISSUE #188: Clicking macOS desktop notification opens Script Editor file picker
## Problem

When agentcoop sends a desktop notification on macOS (e.g. pipeline waiting for input) and the user clicks it, macOS opens the Script Editor "Open" dialog instead of bringing the terminal to the foreground.

This happens because `sendDesktopNotification` in `src/notify.ts` uses:

```ts
execFile("osascript", [
  "-e",
  `display notification "..." with title "..."`,
]);
```

Notifications sent via `osascript display notification` are attributed to Script Editor. Clicking any such notification opens Script Editor, which presents its file picker.

## Cause

AppleScript's `display notification` command has no `sender` parameter. The notification is always attributed to the process that hosts the AppleScript runtime, which is Script Editor.

### `tell application` approach does not work

Wrapping the command in `tell application "iTerm"` or `tell application "Terminal"` was considered but **verified to not solve the problem** on macOS 26.4. The command may execute without error in some environments, but clicking the resulting notification does not bring the target terminal to the foreground. `display notification` is a Standard Additions command — regardless of which application receives the `tell`, the notification sender attribution remains tied to the osascript/Script Editor process. In some environments (e.g. when the target app is not running or not responding to AppleEvents), the command fails with AppleEvent timeout (-1712).

This approach is not viable as a fix.

## Fix

Replace the single `osascript` code path with a terminal-aware dispatch that routes notifications through the appropriate channel depending on the user's terminal emulator.

### Detection order

1. **cmux** — detect via `CMUX_SOCKET_PATH` env var (always set inside cmux sessions for socket communication; present regardless of shell-integration toggle).
   Call `cmux notify --title "..." --body "..."`.
   Fall back to OSC 777 escape sequence (`\x1b]777;notify;title;body\x07`) if the binary is unavailable.

   `CMUX_SOCKET_PATH` is preferred over `CMUX_SHELL_INTEGRATION` because the latter is a shell-integration feature toggle that the user can disable, while the socket path is a structural value that cmux always sets. `command -v cmux` alone is insufficient: the binary may be on PATH while the user is working in a different terminal, which would send the notification to the wrong cmux instance or fail on socket connection.

2. **tmux** — detect via the `TMUX` env var (set in all tmux versions; `TERM_PROGRAM` is unreliable because pre-3.2 tmux does not override it and the outer terminal's value leaks through).
   Walk the process tree from `tmux display-message -p '#{client_pid}'` upward (max 5 levels, stop at PID 1) to identify the outer terminal emulator, then dispatch to the matching path:
   - If cmux is found → `cmux notify` CLI (same as step 1).
   - If iTerm is found → OSC 9 via DCS passthrough (same as step 3, with DCS wrapping).
   - If Apple Terminal is found, or the walk fails (no attached client, SSH nesting, detached session, etc.) → fall back to plain `display notification` via osascript.

3. **iTerm** — detect via `TERM_PROGRAM=iTerm.app`.
   Send OSC 9 escape sequence: `\x1b]9;message\x07`.
   iTerm2 natively handles OSC 9 and displays a notification attributed to iTerm itself. Clicking the notification brings iTerm to the foreground.

4. **Apple Terminal** — no terminal-native notification channel is available. Apple Terminal does not support OSC 9, and the `tell application` approach does not fix click attribution. Fall back to the existing `display notification` via osascript. **The Script Editor click problem is not resolved for Apple Terminal.** Apple Terminal users should set `bell: true` and `desktop: false` in their agentcoop notification config to avoid the Script Editor issue entirely.

5. **Other macOS terminals** — fall back to `display notification` via osascript (same limitation as Apple Terminal).

6. **Linux** — keep the existing `notify-send` path unchanged.

### OSC payload sanitization

Notification messages are arbitrary UI text from `notifyInputWaiting`. When embedding these into terminal escape sequences, the following characters must be stripped or replaced to prevent the sequence from being truncated or misparsed:

- **BEL** (`\x07`) — terminates OSC sequences; strip from payload.
- **ESC** (`\x1b`) — starts new escape sequences; strip from payload.
- **ST** (`\x9c`) — alternative OSC terminator; strip from payload.
- **Semicolons** (`;`) — separator in OSC 777 (`notify;title;body`); replace with comma or strip in both title and body fields. Not an issue for OSC 9, which uses a single payload field.

Apply sanitization before constructing the escape sequence. The existing `escapeAppleScript()` in `notify.ts` only handles AppleScript string literals and is not sufficient for terminal sequences.

### Known limitation: Apple Terminal

This fix does not resolve the Script Editor click problem for Apple Terminal users. There is no known way to send a macOS notification attributed to Terminal.app without third-party dependencies (e.g. `terminal-notifier`), and `terminal-notifier` itself is broken on recent macOS versions (see [julienXX/terminal-notifier#312](https://github.com/julienXX/terminal-notifier/issues/312)).

For Apple Terminal users, the recommended workaround is to use bell-only notifications:

```json
{
  "notifications": {
    "bell": true,
    "desktop": false
  }
}
```

### OSC 9 inside tmux

When iTerm is the outer terminal and tmux is in between, raw OSC 9 sequences are consumed by tmux and do not reach iTerm by default. The tmux process-tree walk (step 2) identifies iTerm as the outer terminal. To decide whether to use DCS passthrough:

1. Probe the tmux setting: `tmux show -gv allow-passthrough`.
2. If the value is `on` or `all`, wrap OSC 9 in DCS passthrough: `\x1bPtmux;\x1b\x1b]9;message\x07\x1b\\`.
3. If the value is `off`, empty, or the probe fails, skip DCS and fall back to plain `display notification` via osascript.

### OSC 777 inside tmux (cmux case)

When cmux is the outer terminal and tmux is in between, prefer `cmux notify` CLI over OSC 777. The CLI communicates via the socket (`CMUX_SOCKET_PATH`), bypassing the terminal stream entirely. This avoids tmux passthrough configuration requirements.

### Error handling

All failures are silently swallowed (existing contract). No notification failure should block or break the prompt flow.

### Testability

- Extract a pure `detectNotifier(env)` function that returns the notifier type based on environment variables alone. Fully unit-testable.
- Extract the tmux process-tree walk into a separate function with an injectable `ExecRunner`. Test scenarios: normal chain (Terminal → shell → tmux client), short chain, PID 1 reached, `tmux display-message` failure, no matching terminal found.
- Extract the tmux passthrough probe (`tmux show -gv allow-passthrough`) into a testable function with an injectable `ExecRunner`. Test `on`, `off`, empty, and error cases.
- Extract a `sanitizeOscPayload(text)` function. Unit-test that BEL, ESC, ST, and semicolons are stripped/replaced.
- Dispatch function: mock `execFile`/`execFileSync` to verify correct arguments for each notifier path (existing test pattern in `src/notify.test.ts`).

### Verification results

All notification paths were manually tested on macOS 26.4. Results:

| Environment | Method | Notification | Click behavior | Note |
|---|---|---|---|---|
| cmux (direct) | `cmux notify` CLI | ✅ | cmux panel + focus | |
| cmux + tmux | `cmux notify` CLI | ✅ | cmux panel + focus | Selected implementation path (socket, no passthrough needed) |
| cmux + tmux | DCS OSC 9 (passthrough on) | ✅ | cmux notification | Verified but not the selected path; `cmux notify` CLI is preferred |
| iTerm (direct) | OSC 9 (`\x1b]9;...\x07`) | ✅ | iTerm focus | |
| iTerm + tmux | DCS OSC 9 (passthrough on) | ✅ | iTerm focus | |
| iTerm + tmux | process-tree walk | ✅ | iTerm2 found in ancestor chain | |
| any terminal | `display notification` (osascript) | ✅ | Script Editor (known limitation) | |
| any terminal | `tell app "iTerm"/"Terminal" to display notification` | ⚠️ | Does not fix click attribution | |

sehkone:
## Summary

- Replace the single `osascript` notification path with terminal-aware dispatch that routes through cmux CLI, iTerm2 OSC 9, or tmux DCS passthrough depending on the detected terminal emulator
- Walk the tmux process tree to identify the outer terminal (cmux, iTerm, Apple Terminal) and choose the best notification channel
- Sanitize notification payloads for OSC escape sequences (strip BEL, ESC, ST, semicolons)
- Fall back to `display notification` via osascript for Apple Terminal and unknown terminals (known limitation documented in issue)
- All process-tree probes and cmux CLI calls are fully async (`execFile` with callback-based promises) — notification dispatch never blocks the event loop or the TUI prompt flow

Closes #188

## Not addressed

- **Apple Terminal click attribution**: There is no known way to send a macOS notification attributed to Terminal.app. Apple Terminal users still get the Script Editor behavior when clicking notifications. The recommended workaround is `bell: true, desktop: false` in notification config.

## Test plan

- [x] Verify `detectNotifier` returns correct notifier for each env var combination
- [x] Verify `sanitizeOscPayload` strips BEL, ESC, ST, and semicolons
- [x] Verify `findTmuxOuterTerminal` walks process tree and identifies cmux, iTerm, Apple Terminal
- [x] Verify `findTmuxOuterTerminal` handles edge cases (PID 1, empty PID, max depth, errors)
- [x] Verify `probeTmuxPassthrough` returns true for `on`/`all`, false for `off`/empty/error
- [x] Verify cmux path calls `cmux notify` CLI and falls back to OSC 777
- [x] Verify iTerm path sends OSC 9 and sanitizes payload
- [x] Verify tmux+iTerm sends DCS-wrapped OSC 9 when passthrough enabled, falls back to osascript when disabled
- [x] Verify tmux+cmux calls `cmux notify` CLI
- [x] Verify tmux+Apple Terminal and tmux failures fall back to osascript
- [x] Verify Linux path still uses `notify-send` unchanged
- [x] Verify all notification failures are silently swallowed
- [x] Run full test suite: all 1327 tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,7 +277,7 @@ InputArea (e.g., "Agent is blocked: push rejected").
 | Channel | Setting | Default | Behavior |
 | ------- | ------- | ------- | -------- |
 | Terminal bell | `notifications.bell` | `true` | Emits `BEL` (`\x07`) to stdout. Most terminals flash the tab or play a sound. |
-| Desktop | `notifications.desktop` | `false` | macOS: `osascript display notification`. Linux: `notify-send`. Silently ignored if the command is unavailable or no GUI session exists. |
+| Desktop | `notifications.desktop` | `false` | macOS: terminal-aware dispatch (cmux CLI, iTerm2 OSC 9, tmux DCS passthrough, osascript fallback). Linux: `notify-send`. Silently ignored if the command is unavailable or no GUI session exists. |
 
 Both channels can be toggled independently in
 `~/.agentcoop/config.json` or through the startup wizard
--- a/src/notify.test.ts
+++ b/src/notify.test.ts
@@ -9,7 +9,7 @@ import {
 } from "vitest";
 import type { NotificationSettings } from "./config.js";
 
-// Mock child_process.execFile before importing the module under test.
+// Mock child_process before importing the module under test.
 vi.mock("node:child_process", () => ({
   execFile: vi.fn(),
 }));
@@ -20,24 +20,47 @@ vi.mock("./i18n/index.js", () => ({
 }));
 
 const { execFile } = await import("node:child_process");
-const { notifyInputWaiting, _emitBell, _sendDesktop } = await import(
-  "./notify.js"
-);
+const {
+  notifyInputWaiting,
+  _emitBell,
+  _sendDesktop,
+  _detectNotifier,
+  _sanitizeOscPayload,
+  _findTmuxOuterTerminal,
+  _probeTmuxPassthrough,
+} = await import("./notify.js");
+
+const ENV_KEYS = ["CMUX_SOCKET_PATH", "TMUX", "TERM_PROGRAM"] as const;
 
 describe("notifyInputWaiting", () => {
   let stdoutWriteSpy: MockInstance;
+  const savedEnv: Record<string, string | undefined> = {};
 
   beforeEach(() => {
     stdoutWriteSpy = vi.spyOn(process.stdout, "write").mockReturnValue(true);
     vi.mocked(execFile).mockReset();
+    for (const key of ENV_KEYS) {
+      savedEnv[key] = process.env[key];
+      delete process.env[key];
+    }
   });
 
   afterEach(() => {
     stdoutWriteSpy.mockRestore();
+    for (const key of ENV_KEYS) {
+      if (savedEnv[key] !== undefined) {
+        process.env[key] = savedEnv[key];
+      } else {
+        delete process.env[key];
+      }
+    }
   });
 
   test("fires bell when bell is enabled", () => {
-    const settings: NotificationSettings = { bell: true, desktop: false };
+    const settings: NotificationSettings = {
+      bell: true,
+      desktop: false,
+    };
     notifyInputWaiting(settings, "Ready?");
     expect(stdoutWriteSpy).toHaveBeenCalledWith("\x07");
     expect(execFile).not.toHaveBeenCalled();
@@ -47,7 +70,10 @@ describe("notifyInputWaiting", () => {
     const originalPlatform = process.platform;
     Object.defineProperty(process, "platform", { value: "darwin" });
     try {
-      const settings: NotificationSettings = { bell: false, desktop: true };
+      const settings: NotificationSettings = {
+        bell: false,
+        desktop: true,
+      };
       notifyInputWaiting(settings, "Ready?");
       expect(stdoutWriteSpy).not.toHaveBeenCalled();
       expect(execFile).toHaveBeenCalledWith("osascript", [
@@ -55,7 +81,9 @@ describe("notifyInputWaiting", () => {
         'display notification "Ready?" with title "agentcoop"',
       ]);
     } finally {
-      Object.defineProperty(process, "platform", { value: originalPlatform });
+      Object.defineProperty(process, "platform", {
+        value: originalPlatform,
+      });
     }
   });
 
@@ -63,26 +91,37 @@ describe("notifyInputWaiting", () => {
     const originalPlatform = process.platform;
     Object.defineProperty(process, "platform", { value: "linux" });
     try {
-      const settings: NotificationSettings = { bell: false, desktop: true };
+      const settings: NotificationSettings = {
+        bell: false,
+        desktop: true,
+      };
       notifyInputWaiting(settings, "Ready?");
       expect(execFile).toHaveBeenCalledWith("notify-send", [
         "agentcoop",
         "Ready?",
       ]);
     } finally {
-      Object.defineProperty(process, "platform", { value: originalPlatform });
+      Object.defineProperty(process, "platform", {
+        value: originalPlatform,
+      });
     }
   });
 
   test("fires both bell and desktop when both enabled", () => {
-    const settings: NotificationSettings = { bell: true, desktop: true };
+    const settings: NotificationSettings = {
+      bell: true,
+      desktop: true,
+    };
     notifyInputWaiting(settings, "Proceed?");
     expect(stdoutWriteSpy).toHaveBeenCalledWith("\x07");
     expect(execFile).toHaveBeenCalled();
   });
 
   test("fires nothing when both disabled", () => {
-    const settings: NotificationSettings = { bell: false, desktop: false };
+    const settings: NotificationSettings = {
+      bell: false,
+      desktop: false,
+    };
     notifyInputWaiting(settings, "Hello");
     expect(stdoutWriteSpy).not.toHaveBeenCalled();
     expect(execFile).not.toHaveBeenCalled();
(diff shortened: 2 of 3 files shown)


sehkone:
**[Reviewer Round 1]**

Nice direction overall — routing through cmux/iTerm/tmux looks like the right fix for the Script Editor attribution problem, and the branch coverage in `src/notify.test.ts` is good. I do see one blocking issue before merge:

- `sendDesktopNotification()` is now doing several synchronous process launches on the UI thread via `execFileSync` (`src/notify.ts:44`, `src/notify.ts:83`, `src/notify.ts:121`, `src/notify.ts:163`). That call path is reached directly from the input-request dispatch in `src/ui/App.tsx:363`, so any slow or hung `tmux`, `ps`, or `cmux` invocation will freeze the TUI until the timeout fires. In the tmux path that can stack across multiple probes, which is the opposite of the existing contract that notifications “never block or break the prompt flow” (`README.md:285`). Swallowing errors is not enough here, because the synchronous waits happen before the error is caught. I think this needs to stay async (for example by keeping `execFile`-based helpers/background callbacks) so notification detection failures cannot stall prompt rendering.

NOT_APPROVED

sehkone:
**[Author Round 1]**

Agreed — the synchronous `execFileSync` calls were a bug. Fixed in c3fa522.

**What changed:**

- Removed `execFileSync` entirely. The `ExecRunner` type is now `(cmd, args) => Promise<string>`, and `defaultExecRunner` wraps `execFile` in a callback-based promise with the same 5 s timeout.
- `findTmuxOuterTerminal`, `probeTmuxPassthrough`, and `sendDesktopNotification` are all `async` now. The tmux process-tree walk, passthrough probe, and cmux CLI call all go through the async runner — no synchronous waits on the event loop.
- `notifyInputWaiting` remains synchronous (void return, same call-site contract). It fires `sendDesktopNotification(message).catch(() => {})` — fully fire-and-forget, no blocking.
- `sendDesktopNotification` accepts an optional `run` parameter (same injection pattern as the helpers) so tests can pass a mock runner for cmux/tmux paths without needing to mock `execFile` callback behavior.
- All 50 notification tests updated to async mocks (`mockResolvedValueOnce` / `mockRejectedValue`) and `await`. Full suite (1327 tests) passes.

sehkone:
**[Reviewer Round 2]**

I re-checked the previous blocking issue, and it is resolved: `notifyInputWaiting()` now fire-and-forgets `sendDesktopNotification()` again, while the tmux/cmux helpers are back on promise-based `execFile` calls instead of synchronous process launches (`src/notify.ts:18`, `src/notify.ts:44`, `src/notify.ts:93`, `src/notify.ts:129`, `src/notify.ts:143`). That removes the prompt-freezing regression from round 1.

I also re-read the updated diff against the issue requirements. The notifier detection order, tmux outer-terminal walk, tmux passthrough probe, OSC sanitization, and macOS/Linux dispatch paths in `src/notify.ts` all line up with the intended behavior, and the expanded coverage in `src/notify.test.ts` exercises the new branches and fallback cases in a meaningful way.

APPROVED

PR #192, opened 2026-04-13
Improve Done stage rebase flow: debug logging and stage name display (#191)

### LINKED ISSUE #191: Improve Done stage rebase flow: debug logging and stage name display
### Problem

When selecting "Let agent rebase" in the Done stage (stage 9), the CLI exits with "CI checks passed." as the final output instead of showing the merge confirmation prompt (`askMerge`). The rebase itself succeeds (confirmed on GitHub), but the expected flow is interrupted.

#### Expected flow

```
rebase success → afterResolution → checkMergeable → MERGEABLE
  → pollCiAndFix → { passed: true } → askMerge (merge confirmation prompt)
  → user confirms → cleanup → "Pipeline completed successfully."
```

#### Observed behavior

```
rebase success → ... → "CI checks passed." → clean exit (no merge prompt)
```

Static analysis could not identify the code path that produces "CI checks passed." as the final pipeline result message, since:

- `ci-poll.ts` returns `{ passed: true }` when CI passes, so `afterResolution` should call `askMerge` (not the `!ciResult.passed` early return)
- The pipeline success path always returns `"Pipeline completed successfully."`, not the stage message
- No abort/error path produces "CI checks passed." as its message

### Task

Add debug logging at two key points to capture the actual values on the next reproduction:

1. **`afterResolution` in `pipeline.ts`** — log `ciResult` (both `passed` and `message`) right after `pollCiAndFix` returns, before the `!ciResult.passed` branch
2. **`index.ts` exit path** — log the full `pipelineResult` object (including `success`, `stoppedAt`, `message`, `cancelled`) right before `console.log(pipelineResult.message)`

Use `console.error` for diagnostic output (consistent with existing practice in `stage-util.ts`). Use a fixed prefix `[done-stage-debug]` for all log lines so they can be easily found with `grep`.

### Acceptance criteria

- All diagnostic log lines use the `[done-stage-debug]` prefix.
- Logging is temporary for diagnosis. Remove or convert to a debug-level gate once the root cause is identified.

### Problem

When the user selects "Let agent rebase" in the Done stage, the TUI status bar continues to show "Stage 9: Done" throughout the rebase operation. This makes it unclear what the pipeline is currently doing.

### Task

Switch the TUI status bar to display "Rebase" (e.g., "Stage 9: Rebase") while the rebase is in progress, then revert to "Done" once the rebase and subsequent CI polling both complete. The status bar currently relies solely on `stage:enter` events for the stage name (`StatusBar.tsx:203`), so a new event or state mechanism is needed.

Both rebase paths must be covered:
- `handleConflicting` rebase (`pipeline.ts:947`)
- `askMerge` inner rebase (`pipeline.ts:1085`)

### Acceptance criteria

- "Rebase" is displayed when the rebase agent starts, for both rebase paths.
- "Rebase" remains displayed during the subsequent CI polling (since the force-push from rebase triggers new CI runs that must pass before the merge prompt).
- "Done" is restored when:
  - Rebase + CI polling both complete successfully (before merge prompt)
  - Rebase fails (fallback to manual resolve)
  - User cancels (Ctrl+C)

## References

- `src/pipeline.ts:1012–1023` — `afterResolution` CI check and `askMerge` call
- `src/pipeline.ts:947–964` — `handleConflicting` agent rebase path
- `src/pipeline.ts:1085–1113` — `askMerge` inner rebase path
- `src/index.ts:808–809` — final `console.log(pipelineResult.message)`
- `src/ci-poll.ts:172–173` — CI pass return value
- `src/ui/StatusBar.tsx:203` — stage name display (reads `stage:enter` events)
- `src/stage-util.ts:45` — existing `console.error` diagnostic pattern

sehkone:
## Summary

- Add `[done-stage-debug]` diagnostic logging at two key points to capture the actual values during the rebase → CI → merge flow: after `pollCiAndFix` returns in `afterResolution`, and at the `index.ts` exit path before the final `console.log(pipelineResult.message)`.
- Add a `stage:name-override` event so the TUI StatusBar displays "Rebase" while a rebase operation is in progress, covering both the `handleConflicting` rebase path and the `askMerge` inner rebase path. The display reverts to "Done" on success, failure, or cancellation.
- Route the rebase stage label through the i18n translation path (`stage.rebase` key) so it renders correctly in all supported languages (e.g. "리베이스" in Korean instead of mixed-language "Rebase").
- Add a StatusBar component test verifying that `stage:name-override` events change and restore the rendered stage name.

Closes #191

## Test plan

- [x] Verify `[done-stage-debug] afterResolution ciResult:` log line appears in stderr after a rebase triggers CI polling
- [x] Verify `[done-stage-debug] pipelineResult:` log line appears in stderr when the pipeline exits
- [x] Verify StatusBar shows "Rebase" when agent rebase starts via the conflict prompt (`handleConflicting` path)
- [x] Verify StatusBar shows "Rebase" when agent rebase starts via the merge prompt (`askMerge` inner rebase path)
- [x] Verify StatusBar reverts to "Done" after rebase + CI polling completes successfully
- [x] Verify StatusBar reverts to "Done" when rebase fails
- [x] Verify StatusBar reverts to "Done" on Ctrl+C cancellation
- [x] Verify `pnpm vitest run` passes (pipeline-events, pipeline, and components test suites cover the new event, name-override emissions, and StatusBar rendering)
- [x] Verify `pnpm tsc --noEmit` and `pnpm biome check` pass

### DIFF
--- a/src/i18n/en.ts
+++ b/src/i18n/en.ts
@@ -92,6 +92,7 @@ export const en: Messages = {
   "stage.ciCheck": "CI check",
   "stage.testPlan": "Test plan verification",
   "stage.squash": "Squash commits",
+  "stage.rebase": "Rebase",
   "stage.review": "Review",
   "stage.done": "Done",
 
--- a/src/i18n/ko.ts
+++ b/src/i18n/ko.ts
@@ -116,6 +116,7 @@ export const ko: Messages = {
   "stage.ciCheck": "CI \uAC80\uC0AC",
   "stage.testPlan": "\uD14C\uC2A4\uD2B8 \uACC4\uD68D \uAC80\uC99D",
   "stage.squash": "\uCEE4\uBC0B \uC2A4\uCFFC\uC2DC",
+  "stage.rebase": "\uB9AC\uBCA0\uC774\uC2A4",
   "stage.review": "\uB9AC\uBDF0",
   "stage.done": "\uC644\uB8CC",
 
--- a/src/i18n/messages.ts
+++ b/src/i18n/messages.ts
@@ -93,6 +93,7 @@ export interface Messages {
   "stage.ciCheck": string;
   "stage.testPlan": string;
   "stage.squash": string;
+  "stage.rebase": string;
   "stage.review": string;
   "stage.done": string;
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -555,7 +555,10 @@ try {
       }
     | undefined;
 
+  const emitter = new PipelineEventEmitter();
+
   const doneStage = createDoneStageHandler({
+    events: emitter,
     checkMergeable: async () => checkMergeable(owner, repo, wt.branch),
     prompt: {
       confirmMerge: async (msg) => {
@@ -769,8 +772,6 @@ try {
   };
   process.on("SIGINT", sigintHandler);
 
-  const emitter = new PipelineEventEmitter();
-
   const startedAt = Date.now();
   const pipelineResult = await new Promise<PipelineResult>((resolve) => {
     const { unmount } = renderApp({
@@ -805,6 +806,9 @@ try {
   // (allows Ctrl+C to kill during cleanup prompts).
   process.off("SIGINT", sigintHandler);
 
+  console.error(
+    `[done-stage-debug] pipelineResult: success=${pipelineResult.success} stoppedAt=${pipelineResult.stoppedAt} cancelled=${pipelineResult.cancelled} message=${pipelineResult.message}`,
+  );
   console.log();
   console.log(pipelineResult.message);
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -807,7 +807,7 @@ try {
   process.off("SIGINT", sigintHandler);
 
   console.error(
-    `[done-stage-debug] pipelineResult: success=${pipelineResult.success} stoppedAt=${pipelineResult.stoppedAt} cancelled=${pipelineResult.cancelled} message=${pipelineResult.message}`,
+    `[done-stage-debug] pipelineResult: ${JSON.stringify(pipelineResult)}`,
   );
   console.log();
   console.log(pipelineResult.message);
--- a/src/pipeline-events.test.ts
+++ b/src/pipeline-events.test.ts
@@ -7,6 +7,7 @@ import {
   PipelineEventEmitter,
   type StageEnterEvent,
   type StageExitEvent,
+  type StageNameOverrideEvent,
 } from "./pipeline-events.js";
 
 describe("PipelineEventEmitter", () => {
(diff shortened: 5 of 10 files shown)


sehkone:
**[Reviewer Round 1]**

Thanks — the debug logging part looks good, but I don’t think the stage-name change is quite ready yet.

1. `src/pipeline.ts:956` and `src/pipeline.ts:1112` hard-code `"Rebase"`. Everywhere else the status bar stage label comes from i18n keys (`src/i18n/messages.ts:90`, `src/i18n/messages.ts:97`, `src/i18n/en.ts:92`, `src/i18n/en.ts:96`, `src/i18n/ko.ts:116`, `src/i18n/ko.ts:120`), so in Korean the new state will render as mixed-language `9단계: Rebase`. Since this is a user-visible stage name, I think it should go through the same translation path as the other stage labels, e.g. by adding a `stage.rebase` message and emitting that localized value instead of an English literal.

2. The new behavior is only tested at the event-emission layer (`src/pipeline.test.ts`) and emitter typing layer (`src/pipeline-events.test.ts`). The acceptance criteria are about what the TUI actually shows, but there isn’t a matching component test in the existing `StatusBar` test area (`src/ui/components.test.tsx:879`) that proves `stage:name-override` changes the rendered text and then restores it. A small `StatusBar` test around the new listener wiring in `src/ui/StatusBar.tsx:217` would make this a lot safer against regressions.

Once those are addressed, I’d be happy to take another look.

sehkone:
**[Author Round 1]**

Both items addressed in 35f2f99:

1. **i18n for rebase label** — Added `stage.rebase` key to `messages.ts`, `en.ts` ("Rebase"), and `ko.ts` ("리베이스"). `pipeline.ts` now reads `t()["stage.rebase"]` instead of the hard-coded English literal, so Korean users will see "9단계: 리베이스" instead of mixed-language text.

2. **StatusBar component test** — Added a test in `components.test.tsx` that renders `StatusBar`, emits `stage:enter` for stage 9 "Done", then emits `stage:name-override` to "Rebase" and verifies the displayed name changes, then emits another override back to "Done" and verifies it restores.

sehkone:
**[Reviewer Round 2]**

Thanks — the two round-one items are resolved. The rebase label now goes through i18n, and the new `StatusBar` component test covers the actual rendered override/restore behavior.

I still see one issue in the debug logging:

1. `src/pipeline.ts:1027` and `src/index.ts:809` interpolate `ciResult.message` / `pipelineResult.message` directly into a template string. If either `message` contains a newline — which is realistic here because these values can come from agent output or error text — `console.error(...)` will emit multiple physical stderr lines, and only the first one will start with `[done-stage-debug]`. That breaks the acceptance criterion that the diagnostic lines are grepable via the fixed prefix, and it also falls a bit short of logging the full object requested in the issue. Please switch these logs to a single-line structured serialization (`JSON.stringify`, `%j`, etc.) so the entire record stays on one prefixed line.

NOT_APPROVED

PR #196, opened 2026-04-13
Standardize verdict handling across pipeline stages (#193)

### LINKED ISSUE #193: Standardize verdict handling across all pipeline stages
## Problem

The pipeline determines stage outcomes by scanning agent `responseText` for the **last keyword** (`COMPLETED`, `APPROVED`, `NOT_APPROVED`, etc.) via `parseStepStatus`. This is fragile: if an agent uses a keyword in a non-verdict context after the actual verdict — e.g. "Round 1 items are now APPROVED" after a `NOT_APPROVED` verdict — the parser picks up the wrong keyword and the pipeline takes the wrong branch.

### Observed incident

In PR #192 (issue #191), Reviewer (Agent B, GPT-5.4) posted a `[Reviewer Round 2]` comment ending with `NOT_APPROVED`. However, Agent B's full `responseText` (which includes post-comment analysis) apparently contained a terminal keyword after `NOT_APPROVED`. The parser (`step-parser.ts:64`) selected that later keyword, causing the review stage to take the APPROVED path. As a result:

- Agent A was never invoked to address Round 2 feedback (no `[Author Round 2]` comment exists on the PR).
- The pipeline skipped directly to squash, leaving the reviewer's concern unresolved.

### Root cause

`parseStepStatus` (`step-parser.ts:45`) scans for **all** keywords (`COMPLETED`, `DONE`, `FIXED`, `APPROVED`, `NOT_APPROVED`, `BLOCKED`) and returns whichever appears **last** in the text. Agent conversational output is free-form and may use these words in non-verdict contexts, making the "last keyword wins" strategy unreliable.

## Fix approach

Establish a **single, consistent verdict pattern** across all pipeline stages:

1. **Work step** — the agent performs its task and responds freely. This response is **not parsed** for keywords.
2. **Verdict follow-up** — a dedicated follow-up prompt asks for **only** the verdict keyword. Parsed from this constrained response.
3. **Substep-scoped keyword validation** — each verdict call site declares its valid keywords and passes them to the verdict helper and clarification prompt. The verdict helper rejects any keyword outside that set.
4. **Substep-scoped clarification retry** — if the verdict response is ambiguous or contains an out-of-scope keyword, retry once with a clarification prompt listing only the valid keywords for that specific substep.

### Substep-scoped keyword validation design

Valid keywords must be scoped to the **substep** (individual verdict point), not to the stage. A single stage can have multiple verdict contracts — the review stage alone has four:

| Substep | Valid keywords |
|---|---|
| Reviewer verdict | `APPROVED`, `NOT_APPROVED` |
| Author completion check | `COMPLETED`, `BLOCKED` |
| Unresolved summary | `NONE`, `COMPLETED` |
| PR finalization | `PR_FINALIZED` |

Putting a single `validVerdicts` on `StageDefinition` would create a contradiction for any multi-substep stage.

Instead, the keyword set is passed at each **call site**:

- **Verdict helpers** (`mapResponseToResult`, `mapFixOrDoneResponse` in `stage-util.ts`) receive the valid keyword set as a parameter and reject out-of-scope keywords as `needs_clarification`.
- **Clarification prompts** (`buildClarificationPrompt` in `step-parser.ts`) receive the valid keyword set and list only those keywords. All three current call sites must be updated: `pipeline.ts:736`, `stage-createpr.ts:147`, and `stage-squash.ts:188`.
- **Pipeline-level clarification** (`pipeline.ts:736`) — when a stage handler returns `needs_clarification`, it includes the valid keyword set in `StageResult` so the pipeline engine can build a correctly scoped clarification prompt without knowing stage internals.

This requires adding an optional `validVerdicts` field to `StageResult` (not `StageDefinition`):

```typescript
interface StageResult {
  outcome: StageOutcome;
  message: string;
  /** Valid keywords for clarification — set when outcome is needs_clarification. */
  validVerdicts?: string[];
}
```

Every verdict prompt must follow the same structure:

```
<Context sentence about what just happened.>
Respond with exactly one of the following keywords:

- KEYWORD_A — <when to use>
- KEYWORD_B — <when to use>

Do not include any other commentary — just the keyword.
```

### Must-fix — core control-flow bugs

These parse keywords from long, free-form work responses without a verdict follow-up. A misparsed keyword directly alters the pipeline path, causing stages to be skipped or loops to be broken.

| Location | Agent | Prompt | Current extraction | Issue |
|---|---|---|---|---|
| `stage-review.ts:306, 361` | B | `buildReviewPrompt` — full review | `parseStepStatus(responseText)` | **Observed bug** — misparsed verdict skipped author fix loop |
| `stage-review.ts:343` | A | `buildPrFinalizationPrompt` — PR body comparison + update | `responseText.includes("PR_FINALIZED")` | `includes` — loosest check, any mention passes |
| `stage-selfcheck.ts:160` | A | `buildFixOrDonePrompt` — "fix issues now, end with FIXED" | `mapFixOrDoneResponse(responseText)` | Work + verdict in same response |
| `stage-testplan.ts:153` | A | `buildTestPlanSelfCheckPrompt` — "verify, fix, end with FIXED/DONE" | `mapFixOrDoneResponse(responseText)` | Work + verdict in same response |
| `index.ts:633` | A | Inline rebase prompt — conflict resolution + force-push | `parseStepStatus(responseText)` | Same vulnerability as review |

### Hardening — existing follow-ups that need tightening

These already separate work from verdict via `sendFollowUp`, but either allow extra text in the verdict response, lack internal clarification retry, accept keywords outside their substep's valid set, or lack production-side keyword validation.

| Location | Prompt | Issue |
|---|---|---|
| `stage-review.ts:215-223` | `buildAuthorCompletionCheckPrompt` — COMPLETED / BLOCKED | Missing "Do not include any other commentary" constraint |
| `stage-implement.ts:66-76, 107-123` | `buildCompletionCheckPrompt` — COMPLETED / BLOCKED | Has "no other commentary" constraint but **lacks internal clarification retry** — ambiguous response causes `pipeline.ts:729` to re-run the entire implementation stage |
| `stage-createpr.ts:81-91, 147` | `buildPrCompletionCheckPrompt` — COMPLETED / BLOCKED | Allows "add a brief reason on the next line"; internal clarification retry at line 147 calls `buildClarificationPrompt` which must be updated to pass valid keywords |
| `stage-squash.ts:110-121, 188` | `buildSquashCompletionCheckPrompt` — COMPLETED / BLOCKED | Same as create-pr; internal clarification retry at line 188 calls `buildClarificationPrompt` |
| `stage-review.ts:225-235, 547-549` | `buildUnresolvedSummaryPrompt` — NONE / COMPLETED | Checks `upper.endsWith("NONE")` on full response |
| `issue-sync.ts:59-71` | `buildIssueSyncPrompt` — parses `ISSUE_UPDATED` / `ISSUE_COMMENTED` | Work + status in same response; affects summary log, not pipeline path |
| `step-parser.ts:87-96` | `buildClarificationPrompt` — lists all 6 keywords | Not substep-scoped — allows keywords that don't belong to the current substep |
| `pipeline.ts:736` | Pipeline-level auto-clarification — calls `buildClarificationPrompt` | Must use `StageResult.validVerdicts` to build a scoped prompt |
| `stage-util.ts:328, 387` | `mapResponseToResult` / `mapFixOrDoneResponse` | Accept any keyword without validation — out-of-scope keywords silently pass |

## Documentation and comments

Update documentation and source comments to reflect the consistent verdict pattern:

- **`docs/pipeline.md`**:
  - "Explicit completion keywords" section (line 55): Rewrite to describe the two-step pattern as the universal approach. Add the canonical verdict prompt template.
  - All stage sections that document prompts: update to show the work/verdict separation.
  - Keyword contract table (line 62): Simplify — all stages now use the same follow-up pattern, with substep-specific keyword sets.

- **Source module doc comments** — the following all describe the old "work + verdict in same response" model and need updating:
  - `src/stage-review.ts` (lines 1-20): "ending with APPROVED or NOT_APPROVED"
  - `src/stage-selfcheck.ts` (lines 1-12): "The agent responds with FIXED ... or DONE"
  - `src/stage-testplan.ts` (lines 1-13): same as self-check

## Files to change

- `src/pipeline.ts` — Add optional `validVerdicts` field to `StageResult`; update `buildClarificationPrompt` call site (line 736) to use `StageResult.validVerdicts`.
- `src/stage-util.ts` — Parameterize `mapResponseToResult` and `mapFixOrDoneResponse` to accept valid keyword set; reject out-of-scope keywords as `needs_clarification`.
- `src/stage-review.ts` — Add verdict follow-ups for reviewer (B) and PR finalization (A); tighten author completion check and unresolved summary; pass valid keywords at each call site; update module doc comment.
- `src/stage-selfcheck.ts` — Split `buildFixOrDonePrompt` into work step + verdict follow-up; pass valid keywords; update module doc comment.
- `src/stage-testplan.ts` — Split `buildTestPlanSelfCheckPrompt` into work step + verdict follow-up; pass valid keywords; update module doc comment.
- `src/stage-implement.ts` — Add internal clarification retry; pass valid keywords.
- `src/stage-createpr.ts` — Tighten `buildPrCompletionCheckPrompt` to verdict-only; update `buildClarificationPrompt` call (line 147) to pass valid keywords.
- `src/stage-squash.ts` — Tighten `buildSquashCompletionCheckPrompt` to verdict-only; update `buildClarificationPrompt` call (line 188) to pass valid keywords.
- `src/issue-sync.ts` — Separate sync action from status report.
- `src/step-parser.ts` — Parameterize `buildClarificationPrompt` to accept a keyword set.
- `src/index.ts` — Add verdict follow-up for rebase (extract inline handler if needed for testability); pass valid keywords.
- `docs/pipeline.md` — Update prompt documentation and keyword contract description.
- `src/stage-review.test.ts` — Add regression tests for reviewer verdict and PR finalization follow-ups, including ambiguous response retry.
- `src/stage-selfcheck.test.ts` — Update tests for the split follow-up.
- `src/stage-testplan.test.ts` — Update tests for the split follow-up.
- `src/stage-implement.test.ts` — Add test for clarification retry; fix tests that accept out-of-scope keywords.
- `src/stage-createpr.test.ts` — Update tests for tightened verdict prompt; fix tests that accept out-of-scope keywords.
- `src/stage-squash.test.ts` — Update tests for tightened verdict prompt.
- `src/issue-sync.test.ts` — Update tests for separated sync/status pattern.
- `src/stage-e2e.test.ts` — Update prompt-shape dependent tests (PR finalization, review flow, etc.).
- `src/step-parser.test.ts` — Update `buildClarificationPrompt` tests for parameterized keyword set.
- `src/stage-util.test.ts` — Update `mapResponseToResult` / `mapFixOrDoneResponse` tests; add tests for out-of-scope keyword rejection.
- `src/pipeline.test.ts` — Update tests for pipeline-level clarification with `StageResult.validVerdicts`.
- Done-stage tests (new file or existing helper tests) — Add tests for rebase verdict follow-up. (`src/index.test.ts` is package/build smoke tests and not the right place.)

sehkone:
## Summary

- Introduces a two-step verdict pattern across all pipeline stages: agents perform their task freely, then a dedicated follow-up prompt asks for exactly one verdict keyword. This prevents misparsed keywords when agents use verdict words in non-verdict contexts.
- Adds a strict verdict parser (`parseVerdictKeyword`) that requires the response to be essentially just the keyword — rejects extra commentary, multiple valid keywords, and out-of-scope keywords. This replaces the old "last keyword wins" `parseStepStatus` for verdict follow-up responses.
- Parameterizes verdict helpers (`mapResponseToResult`, `mapFixOrDoneResponse`, `buildClarificationPrompt`) to accept substep-scoped valid keyword sets, rejecting out-of-scope keywords as `needs_clarification`.
- Adds `validVerdicts` field to `StageResult` so the pipeline engine builds correctly scoped clarification prompts without knowing stage internals.
- After the in-session clarification retry, stages that still get an ambiguous response use conservative fallbacks: FIXED/DONE loops (self-check, test-plan) fall back to `not_approved` so the pipeline loops again; COMPLETED/BLOCKED substeps (implement, author completion) fall back to `blocked` so the user is asked to intervene. Artifact-producing stages (create-PR, squash, PR finalization) verify a post-condition before proceeding.
- Extracts rebase logic from `index.ts` into a dedicated `rebase.ts` module with its own test file.
- Strict-parses the issue-sync verdict response: `parseIssueSyncResponse` now validates that every non-blank line matches a recognised keyword pattern and returns a `{ changes, valid }` result. Self-contradictory responses (e.g. `ISSUE_NO_CHANGES` mixed with `ISSUE_UPDATED:`) are rejected. When the response is malformed, a single clarification retry is attempted before falling back to "failed".

Closes #193

## Test plan

- [x] `pnpm vitest run` passes all 1451 tests
- [x] `pnpm tsc --noEmit` shows no type errors
- [x] `pnpm biome check` reports no lint issues
- [x] Strict parser: response with two in-scope keywords (e.g. `NOT_APPROVED — earlier items are now APPROVED`) is rejected
- [x] Strict parser: response with extra commentary ending in valid keyword is rejected
- [x] Strict parser: exact keyword with trailing punctuation is accepted
- [x] Reviewer verdict: ambiguous response triggers clarification retry with only `APPROVED`/`NOT_APPROVED`
- [x] PR finalization: uses strict verdict parser instead of `endsWith("PR_FINALIZED")`
- [x] Unresolved summary: uses strict verdict parser instead of `endsWith("NONE")`; ambiguous verdict triggers scoped clarification retry
- [x] Self-check stage: work and verdict are separate steps; `FIXED`/`DONE` parsed from follow-up only
- [x] Test-plan stage: same two-step split as self-check
- [x] Implementation stage: internal clarification retry on ambiguous completion check
- [x] Create-PR stage: `buildClarificationPrompt` called with scoped keywords; twice-ambiguous verdict verifies PR exists via `findPrNumber` before completing, reports `BLOCKED` if no PR found
- [x] Squash stage: `buildClarificationPrompt` called with scoped keywords; twice-ambiguous verdict verifies commit count decreased before completing, reports `BLOCKED` if unchanged
- [x] Rebase: verdict follow-up with `COMPLETED`/`BLOCKED` parsed from dedicated prompt
- [x] Issue-sync: status keyword parsed separately from sync work response; malformed verdict triggers clarification retry; still-malformed after retry reports "failed" status; self-contradictory `ISSUE_NO_CHANGES` mixed with change lines is rejected
- [x] Out-of-scope keywords (e.g. `APPROVED` in a `COMPLETED`/`BLOCKED` substep) trigger internal clarification retry; after retry, stages use conservative fallbacks (not_approved for FIXED/DONE loops, blocked for COMPLETED/BLOCKED substeps) rather than silently advancing
- [x] Pipeline engine: `needs_clarification` with `validVerdicts` set builds a scoped clarification prompt containing only the specified keywords (direct engine-level regression test)
- [x] `docs/pipeline.md` accurately describes the strict verdict parsing, per-substep keyword contracts, conservative fallbacks, and post-condition checks for artifact-producing stages
- [x] Unresolved summary uses verdict session ID (not stale review invoke session) when verdict advances session
- [x] Issue sync verdict failure reports `failed` status (not `completed`)
- [x] Clarification retries fall back to the prior session ID when the immediately preceding turn omits one
- [x] Unresolved summary twice-ambiguous verdict conservatively includes summary text rather than bubbling to the pipeline engine
- [x] PR finalization twice-ambiguous verdict verifies PR body consistency: `Closes #N` must not have a contradictory `## Not addressed` section, and `Part of #N` must include one; the body must use exactly one reference form — both present is self-contradictory and blocks
- [x] Author completion twice-ambiguous verdict returns `blocked` so the user can decide how to proceed

### DIFF
--- a/docs/pipeline.md
+++ b/docs/pipeline.md
@@ -10,7 +10,7 @@ the [README](../README.md).
 - [Pipeline overview](#pipeline-overview)
 - [Prompt design principles](#prompt-design-principles)
   - [Self-contained context](#self-contained-context)
-  - [Explicit completion keywords](#explicit-completion-keywords)
+  - [Two-step verdict pattern](#two-step-verdict-pattern)
   - [No confirmation requests](#no-confirmation-requests)
   - [Service-aware instructions](#service-aware-instructions)
   - [PR body as living documentation](#pr-body-as-living-documentation)
@@ -52,21 +52,52 @@ never need to search for context — it is handed to them. This
 avoids wasted tokens on exploration and reduces the chance of the
 agent working against the wrong branch or issue.
 
-### Explicit completion keywords
+### Two-step verdict pattern
 
-Each stage ends with a strict keyword contract. The orchestrator
-parses the agent's final response for a specific keyword to
-determine the next action. This eliminates ambiguity — the
-orchestrator does not try to infer intent from free-form text.
+Every stage that needs to determine an outcome follows a consistent
+two-step pattern:
 
-| Step type | Keywords | Proceed | Loop/Retry |
-| --------- | -------- | ------- | ---------- |
-| Implementation | `COMPLETED` / `BLOCKED` | `COMPLETED` | `BLOCKED` -> user chooses |
-| Loops (self-check, test plan) | `FIXED` / `DONE` | `DONE` | `FIXED` -> repeat |
-| One-shot (PR, squash) | `COMPLETED` / `BLOCKED` | `COMPLETED` | `BLOCKED` -> user chooses |
-| Review approval | `APPROVED` / `NOT_APPROVED` | `APPROVED` | `NOT_APPROVED` -> repeat |
-| Issue sync | `ISSUE_NO_CHANGES` / `ISSUE_UPDATED` / `ISSUE_COMMENTED` | any | best-effort |
-| PR finalization | `PR_FINALIZED` | `PR_FINALIZED` | missing -> clarification |
+1. **Work step** — the agent performs its task and responds freely.
+   This response is **not** parsed for keywords.
+2. **Verdict follow-up** — a dedicated follow-up prompt asks for
+   **only** the verdict keyword.  The keyword is parsed from this
+   constrained response.
+
+Each verdict call site declares its valid keywords and passes them
+to a strict verdict parser (`parseVerdictKeyword`).  The parser
+requires the response to be essentially just the keyword — it
+rejects responses with extra commentary, multiple valid keywords,
+or out-of-scope keywords.  When the parser rejects a response, a
+single clarification retry is attempted listing only the valid
+keywords for that substep.
+
+The canonical verdict prompt template:
+
+```text
+<Context sentence about what just happened.>
+Respond with exactly one of the following keywords:
+
+- KEYWORD_A — <when to use>
+- KEYWORD_B — <when to use>
+
+Do not include any other commentary — just the keyword.
+```
+
+#### Keyword contracts per substep
+
+| Substep | Valid keywords | Proceed | Loop/Retry |
+| ------- | -------------- | ------- | ---------- |
+| Implementation check | `COMPLETED` / `BLOCKED` | `COMPLETED` | `BLOCKED` → user chooses |
+| Self-check verdict | `FIXED` / `DONE` | `DONE` | `FIXED` → repeat |
+| Test plan verdict | `FIXED` / `DONE` | `DONE` | `FIXED` → repeat |
+| PR creation check | `COMPLETED` / `BLOCKED` | `COMPLETED` | `BLOCKED` → user chooses |
+| Squash check | `COMPLETED` / `BLOCKED` | `COMPLETED` | `BLOCKED` → user chooses |
+| Reviewer verdict | `APPROVED` / `NOT_APPROVED` | `APPROVED` | `NOT_APPROVED` → repeat |
+| Author completion | `COMPLETED` / `BLOCKED` | `COMPLETED` | `BLOCKED` → user chooses |
+| Unresolved summary | `NONE` / `COMPLETED` | either | — |
+| PR finalization | `PR_FINALIZED` | `PR_FINALIZED` | missing → clarification → PR body consistency check → blocked |
+| Issue sync | `ISSUE_NO_CHANGES` / `ISSUE_UPDATED` / `ISSUE_COMMENTED` | any | clarification retry; best-effort |
+| Rebase verdict | `COMPLETED` / `BLOCKED` | `COMPLETED` | `BLOCKED` → manual |
 
 ### No confirmation requests
 
@@ -143,19 +174,42 @@ at arbitrary times.
 
 ### Ambiguous response clarification
 
-When an agent's response does not end with a recognized status
-keyword, the orchestrator sends a fixed clarification prompt
-rather than re-running the stage (which could be side-effectful).
-The same template is used in two contexts:
+When a verdict follow-up does not contain a recognized keyword, the
+orchestrator sends a **substep-scoped** clarification prompt listing
+only the keywords valid for that specific substep.  This is used in
+two contexts:
 
-- **Within-stage session resume** (Create PR, Squash) — the
-  orchestrator resumes the existing agent session with the
-  clarification prompt.
-- **Pipeline loop injection** — the prompt is set as
-  `userInstruction` and appears as `## Additional feedback`
-  on the next iteration.
+- **Within-stage session resume** (all stages with verdict
+  follow-ups) — the orchestrator resumes the existing agent
+  session with the scoped clarification prompt.  If the retry
+  also returns an ambiguous response, the stage uses a
+  **conservative fallback** instead of bubbling
+  `needs_clarification` to the pipeline engine (which would
+  re-run side-effectful work steps or route the clarification
+  to the wrong agent in multi-agent stages).  The fallback
+  depends on the substep's keyword contract:
+  - *FIXED / DONE* loops (self-check, test-plan): `not_approved`
+    — the pipeline loops the stage again.
+  - *COMPLETED / BLOCKED* substeps (implement, author
+    completion, create-pr, squash): `blocked` — the user is
+    asked how to proceed.
+- **Pipeline loop injection** — used only when the ambiguous
+  verdict comes from the agent that receives `userInstruction`
+  on the next iteration (e.g. the reviewer verdict in the
+  review stage).  `StageResult.validVerdicts` carries the
+  keyword set from the stage handler to the pipeline engine.
 
-The clarification prompt:
+When `validVerdicts` is provided, the clarification prompt:
+
+```text
+Your previous response did not contain a clear verdict keyword.
+Respond with exactly one of the following keywords: KEYWORD_A, KEYWORD_B
+
+Do not include any other commentary — just the keyword.
+```
+
+When no `validVerdicts` are set (legacy fallback), all six keywords
+are listed:
 
 ```text
 Your previous response did not end with a clear status keyword.
@@ -296,18 +350,28 @@ item, briefly note whether it passes or needs attention.
    them.  Do not refactor unrelated existing code.
 ```
 
-**Fix-or-done prompt:**
+**Fix-or-done work prompt:**
 
 ```text
 Based on your self-check above, decide what to do next.
 
-- If you found issues that need fixing, fix them now and end your
-  response with the keyword FIXED.
-- If everything looks good and no changes are needed, end your
-  response with the keyword DONE.
+- If you found issues that need fixing, fix them now.
+- If everything looks good and no changes are needed, you are done.
 ```
 
-**Loop behavior:** `FIXED` -> repeat self-check. `DONE` -> run
+**Fix-or-done verdict follow-up:**
+
+```text
+You have finished the self-check pass.
+Respond with exactly one of the following keywords:
+
+- FIXED — if you found and fixed issues
+- DONE — if everything looks good and no changes were needed
+
+Do not include any other commentary — just the keyword.
+```
+
+**Loop behavior:** `FIXED` → repeat self-check. `DONE` → run
 issue sync, then proceed. Default auto-budget: 5 iterations
 (configurable via `selfCheckAutoIterations`). When the budget is
 exhausted, the user is asked whether to continue.
@@ -342,23 +406,27 @@ implementation against the original issue description below.
    `gh issue comment {number} --repo {owner}/{repo} --body "..."`
    Do NOT modify the issue description for major changes.
 5. If there are no discrepancies, do nothing.
-
-## Response format
-
-End your response with one of the following:
-
-- If no changes were needed:
-  `ISSUE_NO_CHANGES`
-- If you updated the issue (minor):
-  `ISSUE_UPDATED: <brief description of what changed>`
-- If you added a comment (major):
-  `ISSUE_COMMENTED: <brief description of the discrepancy>`
-
-You may include both ISSUE_UPDATED and ISSUE_COMMENTED if there
-were both minor and major discrepancies.
 ```
 
-Issue sync is best-effort — if it fails, the pipeline continues.
+**Issue sync verdict follow-up:**
+
+```text
+Report what issue sync actions you performed.
+Respond with one or more of the following on separate lines:
+
+- ISSUE_NO_CHANGES — if no changes were needed
+- ISSUE_UPDATED: <brief description> — if you updated the issue
+- ISSUE_COMMENTED: <brief description> — if you added a comment
+
+Do not include any other commentary.
+```
+
+The verdict response is strictly parsed: every non-blank line must
+match one of the recognised keyword patterns.  If the response
+contains extra commentary, missing colons, or unrecognised lines,
+a single clarification retry is attempted.  Issue sync is
+best-effort — if the verdict is still malformed after the retry,
+or if any step fails, the pipeline continues.
 
 ---
 
@@ -410,15 +478,17 @@ result and respond with exactly one of the following keywords:
 - COMPLETED — if the pull request was created successfully
 - BLOCKED — if you could not create the PR and need user intervention
 
-If BLOCKED, add a brief reason on the next line explaining what
-went wrong (e.g. auth failure, push rejected, PR already exists).
+Do not include any other commentary — just the keyword.
 ```
 
 **Ambiguous response handling:** If the completion check response
 does not clearly match `COMPLETED` or `BLOCKED`, the orchestrator
 resumes the same session with a clarification prompt (rather than
 re-running the PR creation step, which would be side-effectful).
-If clarification also fails, the response is shown to the user.
+If clarification also fails, the handler performs a post-condition
+check using `findPrNumber` to verify whether a PR was actually
+created.  If the PR exists, the stage completes; otherwise it
+reports `BLOCKED`.
 
 **Outcome handling:** PR creation is a required step
 (`requiresArtifact: true`). If `BLOCKED`, only **Instruct** and
@@ -555,7 +625,7 @@ You are verifying the test plan for the following GitHub issue.
 6. Make sure CI is still passing after any changes.
 ```
 
-**Self-check prompt:**
+**Self-check work prompt:**
 
 ```text
 Based on your verification above, evaluate the current state.
@@ -566,14 +636,23 @@ Based on your verification above, evaluate the current state.
   recursively) when applicable.
 - Is CI still passing?
 
-If you found and fixed issues during verification, end your
-response with the keyword FIXED.
-
-If everything is verified and passing with no changes needed,
-end your response with the keyword DONE.
+If you found issues during verification, fix them now.
+If everything is verified and passing, you are done.
 ```
 
-**Loop behavior:** `FIXED` -> repeat the verification loop.
+**Test plan verdict follow-up:**
+
+```text
+You have finished the test plan verification pass.
+Respond with exactly one of the following keywords:
+
+- FIXED — if you found and fixed issues
+- DONE — if everything is verified and passing with no changes needed
+
+Do not include any other commentary — just the keyword.
+```
+
+**Loop behavior:** `FIXED` → repeat the verification loop.
 `DONE` -> proceed to review. Default auto-budget: 3 iterations.
 
 ---
@@ -641,9 +720,18 @@ You are reviewing a pull request for the following GitHub issue.
    `**[Reviewer Round {n}]**`. Be specific. Cite file paths and
    line numbers when they help; for broader concerns, explain
    the concern at the appropriate level.
-4. End your response with one of these keywords:
-   - APPROVED — if the changes are ready to merge
-   - NOT_APPROVED — if changes are needed
+```
+
+**Reviewer verdict follow-up** (sent after the review comment):
+
+```text
+You have posted your review comment.
+Respond with exactly one of the following keywords:
+
+- APPROVED — if the changes are ready to merge
+- NOT_APPROVED — if changes are needed
+
+Do not include any other commentary — just the keyword.
 ```
 
 #### Review prompt — Agent B (round 2+)
@@ -665,47 +753,17 @@ step:
      or does not address the concern, keep the item open.
    - Only carry forward items that remain genuinely unresolved.
 3. Review the updated diff against the issue.
-   Your job is an
-   independent judgment on whether this is the right change
-   and whether it is built well — not a mechanical checklist.
-   Read the code, form an opinion, and explain it with
-   concrete references where they help anchor the point.
-
-   Common review angles include:
-   - Whether the approach actually solves the issue, and
-     whether any requirement appears to be dropped, only
-     partially implemented, or implemented in a surprising way.
-   - Correctness on edge cases and failure paths, not just the
-     happy path.
-   - Design quality: readability, appropriate abstractions,
-     avoiding over-engineering, unrelated drive-by changes,
-     dead code, or stray debug output.
-   - Test presence and meaningfulness — especially whether the
-     tests exercise the new behaviour in a way that would have
-     failed before the change. You do NOT need to run the test
-     suite or re-check CI; assume those are already handled and
-     focus on whether the tests are the right tests.
-   - Error handling, security (input validation, injection,
-     secrets, permissions), and obvious performance issues.
-   - Documentation or comments that now appear out of sync with
-     the code.
-   - PR hygiene if it appears off: issue linkage (`Closes #N`
-     vs. `Part of #N` with `## Not addressed` when partial) and
-     a `## Test plan` checklist.
-
-   The list above is guidance, not a limit. If something feels
-   off for any other reason — architectural, stylistic, product,
-   or subtle — raise it.
+   [same review angles block as round 1]
 4. Post your follow-up review as a PR comment prefixed with
    `**[Reviewer Round {n}]**`. Include any still-unresolved
    prior items and any new findings from this round. Be
    specific. Cite file paths and line numbers when they help;
    for broader concerns, explain the concern at the
    appropriate level.
-5. End your response with one of these keywords:
-   - APPROVED — if the changes are ready to merge
-   - NOT_APPROVED — if changes are needed
 ```
+
+The same reviewer verdict follow-up is sent after the round 2+
+review comment.
 
 #### Author fix prompt — Agent A (round N)
 
@@ -775,16 +833,31 @@ next review round begins.
 When the review loop ends (either because B approves or the
 auto-budget is exhausted), B is asked:
 
+**Work step:**
+
 ```text
 The review loop has ended.  Please check whether there are any
 unresolved items from this review cycle.
 
 - If there are unresolved items, post a PR comment prefixed with
   `**[Reviewer Unresolved Round {n}]**` listing each unresolved item.
-  Then end your response with COMPLETED.
-- If there are no unresolved items, respond with exactly: NONE
+- If there are no unresolved items, simply confirm that there is
+  nothing left to address.
 ```
 
+**Verdict follow-up:**
+
+```text
+Respond with exactly one of the following keywords:
+
+- NONE — if there are no unresolved items
+- COMPLETED — if you posted the unresolved items comment
+
+Do not include any other commentary — just the keyword.
+```
+
+If the verdict is ambiguous or contains an out-of-scope keyword,
+the orchestrator retries once with a scoped clarification prompt.
 If B posts an unresolved summary, the orchestrator shows it to
 the user before asking whether to continue (at budget limit) or
 before reporting completion (on approval).
@@ -822,11 +895,28 @@ state of the implementation.
    were not implemented and why.
 4. If the reference or "## Not addressed" section needs to change,
    update the PR body using `gh pr edit --body "..."`.
-5. End your response with PR_FINALIZED.
+```
+
+**PR finalization verdict follow-up:**
+
+```text
+You have finished verifying the PR body.
+Respond with exactly one of the following keywords:
+
+- PR_FINALIZED — if the PR body is now accurate
+
+Do not include any other commentary — just the keyword.
 ```
 
 Agent A must respond with `PR_FINALIZED` for the stage to
-complete.
+complete. If the verdict is ambiguous after the clarification
+retry, the handler verifies the PR body directly for
+consistency: `Closes #N` must not have a contradictory
+`## Not addressed` section, and `Part of #N` must include one.
+If the body is consistent the stage proceeds as completed;
+otherwise it returns `blocked` so the user can intervene. This
+avoids relying on the squash stage to catch a bad body, since
+squash short-circuits on single-commit branches.
 
 **Loop behavior:** Default auto-budget: 5 rounds (configurable
 via `reviewAutoRounds`). When the budget is exhausted, the user
@@ -887,12 +977,14 @@ and respond with exactly one of the following keywords:
 - COMPLETED — if the commits were squashed and force-pushed
 - BLOCKED — if you could not squash and need user intervention
 
-If BLOCKED, add a brief reason on the next line explaining what
-went wrong.
+Do not include any other commentary — just the keyword.
 ```
 
 **Ambiguous response handling:** Same internal clarification
-retry pattern as stage 3 (Create PR).
+retry pattern as stage 3 (Create PR).  If clarification also fails,
+the handler re-checks the branch commit count to verify whether the
+squash actually happened.  If the count decreased, the stage
+completes; otherwise it reports `BLOCKED`.
 
 **Post-squash CI:** After a successful squash and force-push, the
 orchestrator polls CI and invokes Agent A to fix failures if
@@ -1036,10 +1128,18 @@ You are rebasing a feature branch onto the latest main.
 IMPORTANT: If you cannot resolve conflicts cleanly or if the
 build/tests fail after resolution, do NOT push. Instead, abort
 the rebase (`git rebase --abort`) and report failure.
+```
 
-When you are done, end your response with exactly one of:
-- COMPLETED — if the rebase succeeded and was force-pushed.
-- BLOCKED — if you could not resolve conflicts or tests failed.
+**Verdict follow-up:**
+
+```text
+You have finished the rebase attempt.
+Respond with exactly one of the following keywords:
+
+- COMPLETED — if the rebase succeeded and was force-pushed
+- BLOCKED — if you could not resolve conflicts or tests failed
+
+Do not include any other commentary — just the keyword.
 ```
 
 **Rebase constraints and rationale:**
--- a/src/i18n/en.ts
+++ b/src/i18n/en.ts
@@ -225,6 +225,8 @@ export const en: Messages = {
     `${base}\n\nUnresolved items:\n${summary}`,
   "review.fixesApplied": (round) =>
     `Round ${round} fixes applied, CI passed. Proceeding to next review round.`,
+  "review.finalizationUnverified": (issueNumber) =>
+    `PR finalization verdict was ambiguous and the PR body does not reference issue #${issueNumber}. Manual verification required.`,
 
   // ---- stage-util errors -------------------------------------------------
 
--- a/src/i18n/index.test.ts
+++ b/src/i18n/index.test.ts
@@ -89,6 +89,7 @@ describe("catalogs are complete", () => {
       "review.approved": [2],
       "review.unresolvedItems": ["Review approved.", "item 1"],
       "review.fixesApplied": [1],
+      "review.finalizationUnverified": [42],
       "worktree.alreadyExists": ["/tmp/wt"],
     };
 
--- a/src/i18n/ko.ts
+++ b/src/i18n/ko.ts
@@ -259,6 +259,8 @@ export const ko: Messages = {
     `${base}\n\n미해결 항목:\n${summary}`,
   "review.fixesApplied": (round) =>
     `${round}라운드 수정 적용 완료, CI 통과. 다음 리뷰 라운드로 진행합니다.`,
+  "review.finalizationUnverified": (issueNumber) =>
+    `PR 마무리 판정이 모호하고 PR 본문에 이슈 #${issueNumber} 참조가 없습니다. 수동 확인이 필요합니다.`,
 
   // ---- stage-util errors -------------------------------------------------
 
--- a/src/i18n/messages.ts
+++ b/src/i18n/messages.ts
@@ -219,6 +219,7 @@ export interface Messages {
   "review.approved": (round: number) => string;
   "review.unresolvedItems": (base: string, summary: string) => string;
   "review.fixesApplied": (round: number) => string;
+  "review.finalizationUnverified": (issueNumber: number) => string;
 
   // ---- stage-util errors -------------------------------------------------
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -30,6 +30,7 @@ import type {
 import { createDoneStageHandler } from "./pipeline.js";
 import { PipelineEventEmitter } from "./pipeline-events.js";
 import { checkMergeable, findPrNumber } from "./pr.js";
+import { createRebaseHandler } from "./rebase.js";
 import {
   deleteRunState,
   loadRunState,
@@ -44,10 +45,8 @@ import { createReviewStageHandler } from "./stage-review.js";
 import { createSelfCheckStageHandler } from "./stage-selfcheck.js";
 import { createSquashStageHandler } from "./stage-squash.js";
 import { createTestPlanStageHandler } from "./stage-testplan.js";
-import { drainToSink } from "./stage-util.js";
 import type { AgentConfig } from "./startup.js";
 import { modelDisplayName, runStartup, selectTarget } from "./startup.js";
-import { parseStepStatus } from "./step-parser.js";
 import { renderApp } from "./ui/render-app.js";
 import {
   bootstrapRepo,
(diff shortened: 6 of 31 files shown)


sehkone:
**[Reviewer Round 1]**

I don’t think this fully fixes the root bug yet. The new verdict follow-ups still feed their responses through `parseStepStatus`, which intentionally picks the last matching keyword (`src/step-parser.ts:40`, `src/stage-util.ts:360`, `src/stage-util.ts:424`). That means a verdict reply like `NOT_APPROVED — earlier items are now APPROVED` still resolves to `APPROVED`, and because both keywords are in-scope the clarification retry never fires. The same pattern is still used in the review verdict path (`src/stage-review.ts:354`, `src/stage-review.ts:371`), the shared helpers used by implement/create-pr/squash/self-check/test-plan (`src/stage-util.ts:360`, `src/stage-util.ts:424`), and rebase (`src/rebase.ts:118`). The issue here was specifically “last keyword wins”; moving the parse to a shorter follow-up reduces the odds, but it doesn’t actually remove the failure mode. I think the verdict step needs an exact-response parser, or at least one that rejects extra text / multiple valid keywords, instead of reusing the stage parser.

There’s also still some bespoke `endsWith` acceptance in review that bypasses the scoped verdict contract. `hasFinalizationKeyword` accepts any response ending in `PR_FINALIZED` (`src/stage-review.ts:317`), and unresolved-summary handling treats anything ending in `NONE` as success (`src/stage-review.ts:743`). Those paths silently accept commentary instead of retrying with clarification, which is the opposite of the new “exactly one keyword” rule.

The tests mostly cover out-of-scope keywords and missing keywords, but I don’t see coverage for the still-broken case where a verdict response contains two in-scope keywords or extra commentary ending in a valid keyword. That’s the regression I’d want locked down before merging.

NOT_APPROVED

sehkone:
**[Author Round 1]**

Agreed on all points. The core issue was that verdict follow-ups still used `parseStepStatus` (last keyword wins), so a response like `NOT_APPROVED — earlier items are now APPROVED` would silently resolve to `APPROVED` without triggering clarification.

**What changed:**

1. **Strict verdict parser** — Added `parseVerdictKeyword()` in `step-parser.ts` that requires the response to be essentially just the keyword. It rejects responses with extra commentary, multiple valid keywords, or out-of-scope keywords. This replaces `parseStepStatus` for all verdict follow-up responses.

2. **`mapResponseToResult` / `mapFixOrDoneResponse`** — When `validKeywords` is provided, these now route through the strict parser instead of `parseStepStatus`. The old "last keyword wins" behavior is preserved only for backward-compatible calls without `validKeywords`.

3. **`hasFinalizationKeyword`** — Replaced the `endsWith("PR_FINALIZED")` check with `parseVerdictKeyword(text, PR_FINALIZATION_KEYWORDS)`. No more silent acceptance of commentary.

4. **Unresolved summary** — Replaced the `endsWith("NONE")` check with `parseVerdictKeyword(text, UNRESOLVED_KEYWORDS)`. Same fix.

5. **Review verdict / rebase verdict** — Replaced direct `parseStepStatus` calls with `parseVerdictKeyword` in the review verdict path (`stage-review.ts`) and rebase verdict path (`rebase.ts`).

6. **Useless assignment** — Removed `checkResult = retryResult` in `stage-implement.ts:153` (CodeQL finding from the inline review).

7. **Tests** — Added 21 new tests covering:
   - Two in-scope keywords in one response → rejected
   - Extra commentary ending in a valid keyword → rejected
   - Exact keyword with trailing punctuation → accepted
   - Multi-keyword responses through `mapResponseToResult` and `mapFixOrDoneResponse`
   - Updated all e2e and stage tests to use keyword-only verdict responses

8. **Docs** — Updated `docs/pipeline.md` to describe the strict parsing behavior.

sehkone:
**[Reviewer Round 2]**

Thanks — the strict verdict parsing and the review/finalization follow-ups address the parser problems I called out in Round 1. I still see one remaining gap.

- `src/stage-review.ts:269`, `src/stage-review.ts:719`, and `src/stage-review.ts:760` still leave the unresolved-summary substep half on the old contract. The work prompt still tells Agent B to put `COMPLETED` / `NONE` in the work response, and after the dedicated verdict follow-up the handler treats anything other than an exact `NONE` as success (`// COMPLETED or ambiguous — either way, use the summary text`). That means an ambiguous or out-of-scope follow-up like `APPROVED`, `NONE and COMPLETED`, or extra commentary never triggers the scoped clarification retry described in #193; it is silently accepted as “there is an unresolved summary”. This should mirror the other verdict handlers: keep the work prompt verdict-free, validate only `NONE` / `COMPLETED`, retry once with `buildClarificationPrompt(..., UNRESOLVED_KEYWORDS)`, and add a regression test for that retry path.
- The docs are still out of sync with the new two-step verdict model in a few places: `docs/pipeline.md:823`, `docs/pipeline.md:464`, `docs/pipeline.md:939`, and `docs/pipeline.md:1089` still describe verdicts in the work response or allow extra explanation after `BLOCKED`, even though the implementation now expects verdict-only follow-ups.

NOT_APPROVED

sehkone:
**[Author Round 2]**

Agreed on both points.

**Unresolved-summary substep fix:**

1. **Work prompt is now verdict-free** — `buildUnresolvedSummaryPrompt` no longer mentions `COMPLETED` or `NONE`. Those keywords are only in the dedicated verdict follow-up (`buildUnresolvedVerdictPrompt`).

2. **Clarification retry added** — `handleUnresolvedSummary` now validates the verdict response with `parseVerdictKeyword`. If the response is ambiguous (e.g. `APPROVED`, `NONE and COMPLETED`, or extra commentary), it retries once with `buildClarificationPrompt(..., UNRESOLVED_KEYWORDS)` before falling back. This matches the pattern used by reviewer verdict, PR finalization, and all other verdict handlers.

3. **Regression tests** — Added three tests:
   - Ambiguous unresolved verdict (out-of-scope keyword `APPROVED`) triggers clarification retry, correct response on retry is accepted.
   - Multi-keyword response (`NONE and COMPLETED`) triggers clarification retry, correct response on retry is accepted.
   - Clarification retry agent error is surfaced as `outcome: "error"`.

**Docs updated:**

- `docs/pipeline.md` line 464 (create-pr) and line 939 (squash): replaced "If BLOCKED, add a brief reason on the next line" with "Do not include any other commentary — just the keyword" to match the actual prompts.
- `docs/pipeline.md` line 823 (unresolved summary): replaced the single-step prompt with the two-step work/verdict pattern including the verdict follow-up template and clarification retry description.
- `docs/pipeline.md` line 1089 (rebase): replaced "end your response with exactly one of" in the work prompt with a separate verdict follow-up section matching `buildRebaseVerdictPrompt`.

sehkone:
**[Reviewer Round 3]**

The Round 2 unresolved-summary/docs fixes look good overall, but I still see two correctness issues in the updated flow.

- `src/stage-review.ts:417` and `src/stage-review.ts:643` still start the unresolved-summary substep from `reviewResult.sessionId`, even when the reviewer verdict step has already advanced the conversation to a newer session ID. That matters now that the verdict path can do one or two `resume()` turns before reaching the summary step, and the adapter explicitly allows `resume()` to return a replacement session ID. In those cases the unresolved-summary prompt is sent on a stale session instead of the latest reviewer conversation. The rest of `handleUnresolvedSummary()` already threads updated IDs forward (`result.sessionId ?? sessionId`), so these two call sites look like accidental holdovers. I think they should pass the latest reviewer session (`verdictResult.sessionId ?? reviewResult.sessionId`) and add a regression test that changes the session ID during verdict clarification and asserts the unresolved-summary follow-up resumes the new session.

- `src/stage-selfcheck.ts:244-266` marks issue sync as `completed` whenever the sync work step succeeds, even if the new status-report follow-up fails. If `sendFollowUp()` for `buildIssueSyncVerdictPrompt()` returns an error, the code skips parsing changes but still calls `onIssueSyncStatus("completed")`. That makes the final CLI summary report a successful issue sync / “no discrepancies found” even though the verdict response was never collected. Since this change moved issue-sync status into a separate follow-up precisely to make the summary accurate, I think the verdict-error path should report `failed` instead, with a regression test alongside the existing sync-work failure test.

NOT_APPROVED

sehkone:
**[Author Round 3]**

Both items fixed:

- **Stale session ID in unresolved-summary**: Both call sites (`stage-review.ts:417` and `stage-review.ts:643`) now pass `verdictResult.sessionId ?? reviewResult.sessionId` instead of `reviewResult.sessionId`. Added a regression test that returns a new session ID from the verdict follow-up and asserts the unresolved-summary `resume()` receives it. Updated the e2e test (`stage-e2e.test.ts`) to return distinct verdict session IDs and assert accordingly.

- **Issue sync status on verdict error**: Moved `onIssueSyncStatus("completed")` inside the `syncVerdictResult.status === "success"` branch and added an `else` branch that reports `"failed"` (`stage-selfcheck.ts:265`). Added a regression test that returns an error from the sync verdict follow-up and asserts `onIssueSyncStatus` is called with `"failed"`.

All 1424 tests pass, type-check and lint are clean.

sehkone:
**[Reviewer Round 4]**

The Round 3 fixes look good — I verified the stale-session unresolved-summary fix and the issue-sync failed-status fix in the updated diff. I still see two correctness gaps:

- Several of the new internal clarification retries only run when the immediately preceding `resume()` response carries a `sessionId`, and some review paths even pass `result.sessionId` directly back into `sendFollowUp()`. That means a resumed turn that omits a replacement session ID (which is explicitly allowed by `AgentResult` and already handled elsewhere via fallback to the previous session) either skips the clarification retry and bubbles `needs_clarification`, or throws before retrying at all. Concrete examples: `src/stage-review.ts:388`, `src/stage-review.ts:477`, `src/stage-review.ts:577`, `src/stage-implement.ts:133`, `src/stage-createpr.ts:150`, `src/stage-squash.ts:191`, `src/stage-selfcheck.ts:199`, `src/stage-testplan.ts:189`. This is especially risky for side-effectful stages like create-PR and squash, because the pipeline then re-runs the whole stage instead of clarifying in-session. I think these retries should use the known prior session as a fallback, the same way `sendFollowUp()` and the new unresolved-summary / rebase paths already do. The new tests currently lock in the weaker behavior (`src/stage-review.test.ts:730`, `src/stage-squash.test.ts:430`), so I’d update those too.

- `handleUnresolvedSummary()` still treats a twice-ambiguous verdict as if the reviewer had confirmed `COMPLETED`: after one clarification retry, any response that is still neither `NONE` nor `COMPLETED` falls through to `return { error: undefined, summary: summaryText }` at `src/stage-review.ts:811`. So this substep still does not actually enforce the strict verdict contract that the PR is introducing; a garbled second answer still advances and can surface an unresolved-items message the reviewer never confirmed. I think this should bubble a real failure/clarification result instead of silently accepting the summary, and it needs a regression test for the “still ambiguous after retry” case.

NOT_APPROVED

sehkone:
**[Author Round 4]**

Both items addressed:

**1. Session ID fallback for clarification retries**

All 8 listed locations now fall back to the prior turn's session ID when the immediately preceding response omits one:

- `stage-review.ts` (3 sites): review verdict retry uses `verdictResult.sessionId ?? reviewResult.sessionId`, PR finalization verdict retry uses `finalVerdictResult.sessionId ?? finalizeResult.sessionId`, author check retry uses `checkResult.sessionId ?? fixResult.sessionId`.
- `stage-implement.ts`: uses `checkResult.sessionId ?? implResult.sessionId`.
- `stage-createpr.ts`: uses `checkResult.sessionId ?? prResult.sessionId`.
- `stage-squash.ts`: uses `checkResult.sessionId ?? squashResult.sessionId`.
- `stage-selfcheck.ts`: uses `verdictCheckResult.sessionId ?? fixResult.sessionId`.
- `stage-testplan.ts`: uses `verdictResult.sessionId ?? checkResult.sessionId`.

The `&& sessionId` guard that skipped the retry entirely has been removed from all 5 guarded sites (implement, create-PR, squash, self-check, test-plan). The 3 stage-review sites already entered the retry block unconditionally but passed a potentially-undefined session ID directly — those now have fallbacks too.

Updated all 4 tests that locked in the weaker behavior (`stage-review.test.ts`, `stage-squash.test.ts`, `stage-implement.test.ts`, `stage-createpr.test.ts`) — they now verify that the fallback session is used and the clarification succeeds.

**2. `handleUnresolvedSummary` twice-ambiguous fallthrough**

The function now returns `{ error: { outcome: "needs_clarification", validVerdicts: ["NONE", "COMPLETED"] }, summary: undefined }` when the verdict is still ambiguous after one clarification retry, instead of silently accepting the summary text. Added `verdictResult = retryResult` inside the retry block so the error message carries the latest response text.

Added a regression test ("returns needs_clarification when unresolved verdict stays ambiguous after retry") that verifies the error bubbles with the correct `validVerdicts`.

sehkone:
**[Reviewer Round 5]**

Thanks — I verified the Round 4 fixes for the session-ID fallback cases and for the unresolved-summary fallthrough after a second ambiguous verdict.

I still see one correctness problem in the review stage’s pipeline-level clarification path:

- `src/stage-review.ts:499`, `src/stage-review.ts:612`, and `src/stage-review.ts:811` can still bubble `needs_clarification` after the in-session retry, but on the next engine retry `src/pipeline.ts:742` injects the clarification text only as `ctx.userInstruction`, and `src/stage-review.ts:174` only threads that into the initial reviewer prompt for Agent B. So if the ambiguous substep was Agent A’s author-completion or PR-finalization verdict, the clarification never reaches the agent/substep that actually failed; the stage just restarts from the top of review and risks posting another review comment instead of resolving the pending verdict. I think the review stage needs to either route pipeline-level clarification back into the relevant A-side follow-up prompt or keep those “second ambiguity” cases local to the stage instead of bubbling them to the generic pipeline retry path.

NOT_APPROVED

sehkone:
**[Reviewer Round 5]**

I checked the two Round 4 items and both look fixed in `HEAD`: the clarification retries now fall back to the prior session ID, and `handleUnresolvedSummary()` now bubbles `needs_clarification` after a twice-ambiguous verdict instead of silently accepting the summary.

I still see one remaining control-flow gap:

- The new “retry once on the verdict substep” contract still falls back to a pipeline-level auto-clarification that re-enters the whole stage after the handler has already consumed its in-session retry. `src/pipeline.ts:738` always turns `needs_clarification` into a fresh stage iteration, so handlers like `src/stage-implement.ts:132`, `src/stage-createpr.ts:150`, and `src/stage-squash.ts:191` still re-run their work steps if the second verdict answer is ambiguous. In `src/stage-review.ts` this is worse: the late-substep ambiguities bubble at `src/stage-review.ts:497`, `src/stage-review.ts:611`, and `src/stage-review.ts:811`, but the next iteration always starts back at the reviewer prompt `src/stage-review.ts:338`, and only `buildReviewPrompt()` consumes `ctx.userInstruction` (`src/stage-review.ts:174`) while `buildAuthorFixPrompt()` and `buildPrFinalizationPrompt()` do not (`src/stage-review.ts:223`, `src/stage-review.ts:281`). So an ambiguous author-completion or PR-finalization verdict gets sent back to Agent B as “additional feedback” instead of being clarified in the original Agent A substep. That means the pipeline can still re-run side-effectful work or route the clarification to the wrong agent after the second ambiguous reply, which is exactly the failure mode this issue is trying to eliminate.

NOT_APPROVED

sehkone:
**[Author Round 5]**

Fixed the pipeline-level clarification routing gap. After the in-session clarification retry, stages now handle "second ambiguity" locally instead of bubbling `needs_clarification` to the pipeline engine:

- **`stage-review.ts`** — PR-finalization, author-completion, and unresolved-summary substeps no longer bubble after the in-session retry. PR-finalization proceeds as completed (squash verifies the PR body independently). Author-completion falls through to CI polling. Unresolved-summary conservatively includes the summary text.
- **`stage-implement.ts`**, **`stage-createpr.ts`**, **`stage-squash.ts`** — After the internal clarification retry, if the response is still ambiguous, the handler overrides the outcome to `completed`. This prevents the engine from re-running side-effectful work steps (implementation, PR creation, squash + force-push).
- **`stage-e2e.test.ts`** — Updated the out-of-scope keyword test to verify that the stage completes in a single iteration (no pipeline loop).
- **`docs/pipeline.md`** — Updated the "Ambiguous response clarification" section to describe the within-stage fallback behavior and narrowed the "Pipeline loop injection" path to cases where the ambiguous verdict comes from the agent that receives `userInstruction`.

sehkone:
**[Reviewer Round 6]**

I verified the Round 5 routing fix: the late review substeps now keep their second-ambiguity handling local, so the prior “pipeline clarification goes to the wrong agent” problem looks resolved.

I still see one correctness issue with the new fallback-to-completed behavior in the artifact-producing stages:

- `src/stage-createpr.ts:181`, `src/stage-createpr.ts:185`, `src/stage-squash.ts:222`, and `src/stage-squash.ts:227` now turn a twice-ambiguous completion verdict into `completed`. That avoids re-running the side-effectful work, but it also converts “I could not confirm success” into “advance the pipeline anyway” for the two stages whose `requiresArtifact` contract says the artifact must exist before proceeding (`src/stage-createpr.ts:16`, `src/stage-createpr.ts:101`, `src/stage-squash.ts:11`, `src/stage-squash.ts:132`). In the create-PR case this can move the pipeline forward without a PR at all; in the squash case it can skip the only confirmation that the branch was actually force-pushed in the expected shape. The new tests in `src/stage-createpr.test.ts:193`, `src/stage-createpr.test.ts:223`, `src/stage-createpr.test.ts:257`, and `src/stage-squash.test.ts:553` lock that behavior in, so this looks intentional rather than accidental. I think these stages need a safer second-ambiguity path than implicit success — for example, a local user-facing fallback or an explicit post-condition check that verifies the required artifact/state before returning `completed`.

- The docs/comments are also out of sync with the shipped behavior here. `src/stage-createpr.ts:10` and `src/stage-createpr.ts:14` still say a second ambiguous response is returned to the engine for user intervention, and `docs/pipeline.md:474` and `docs/pipeline.md:478` still say the create-PR response is shown to the user if clarification fails. That no longer matches the code path above.

NOT_APPROVED

sehkone:
**[Author Round 6]**

Fixed both items from Round 6:

**Post-condition checks for artifact-producing stages** — Create-PR and squash no longer blindly fall back to `completed` when the verdict is still ambiguous after clarification. Instead:
- Create-PR calls `findPrNumber` to verify a PR exists for the branch. If no PR is found, the stage reports `BLOCKED` (with diagnostic text).
- Squash re-checks `countBranchCommits` and reports `BLOCKED` if the count did not decrease from the initial value.

`findPrNumber` is injected via `CreatePrStageOptions` for testability, following the same pattern squash already uses for `countBranchCommits`.

**Docs/comments sync** — Updated `stage-createpr.ts` module doc comment (lines 10–14) and `docs/pipeline.md` (create-PR and squash ambiguous-response sections) to describe the post-condition check behavior instead of the previous "returned to the engine" / "shown to the user" language.

Tests updated: the double-ambiguity tests now inject post-condition mocks, and new tests cover the `BLOCKED` path when the post-condition fails. The e2e create-PR clarification test was also fixed — its mock was accidentally making both resume calls ambiguous (hidden by the old implicit-success fallback).

sehkone:
**[Reviewer Round 7]**

I re-checked the Round 6 follow-up. The create-PR / squash post-condition checks and the docs updates look good.

I still see one correctness hole in the review finalization fallback:

- `src/stage-review.ts:497` now treats a twice-ambiguous PR-finalization verdict as implicitly completed because “the next stage (squash) will verify the PR body independently.” That assumption is not always true. `src/stage-squash.ts:139` short-circuits stage 8 entirely when the branch already has a single commit, so on single-commit branches there is no later PR-sync check at all. In that case the flow can reach Done even when Agent A’s finalization step never confirmed or updated the PR body / issue reference. The current tests lock both halves of that behavior in (`src/stage-review.test.ts:1573`, `src/stage-squash.test.ts:167`), but there is no coverage for their interaction. I think the review stage needs a safer second-ambiguity path here — for example, verify the PR body directly before returning `completed`, or surface a local blocked/user-intervention path instead of relying on squash to catch it.

NOT_APPROVED

sehkone:
**[Author Round 7]**

Fixed the PR finalization twice-ambiguous fallback. The review stage no longer assumes the squash stage will verify the PR body — since squash short-circuits on single-commit branches, that assumption could let the flow reach Done without a confirmed issue reference.

Changes:
- Added `getPrBody` to `pr.ts` — reads the PR body via `gh pr view --json body --jq .body`.
- `ReviewStageOptions` now accepts an optional injectable `getPrBody` for testability.
- When the finalization verdict is still ambiguous after the clarification retry, the handler reads the PR body and checks for `Closes #N` or `Part of #N`. If present, proceeds as completed; if missing, returns `blocked` for user intervention.
- Updated tests: the two existing "proceeds as completed" tests now provide `getPrBody` mocks. Added new tests for: PR body with issue ref → completed, PR body without ref → blocked, `Part of #N` variant → completed.
- Added `getPrBody` unit tests in `pr.test.ts`.
- Added `review.finalizationUnverified` i18n message (en + ko).
- Updated `docs/pipeline.md` keyword contract table and PR finalization section to describe the new fallback behavior.
- Updated PR description to reflect the post-condition check.

sehkone:
**[Reviewer Round 8]**

I re-checked the Round 7 item. The review-stage fallback now verifies the PR body directly before advancing, so that specific single-commit gap looks resolved.

I still see one remaining standardization gap in the issue-sync sub-step:

- `src/issue-sync.ts:65` and `src/issue-sync.ts:88` still use a bespoke multi-line parser instead of the new strict, substep-scoped verdict flow. `src/stage-selfcheck.ts:255` accepts any successful follow-up response, feeds it through `parseIssueSyncResponse()`, and then unconditionally reports issue sync as `completed`, even when the reply is malformed or contains extra text. For example, a response like `ISSUE_UPDATED fixed title` (missing `:`) or `I updated the issue\nISSUE_UPDATED: fixed title` silently records no changes at all, because the parser just returns `[]` and there is no clarification retry. That seems to fall short of the contract documented in `docs/pipeline.md:60`, which says verdict follow-ups are parsed by the strict verdict parser with scoped validation/retry. Since issue sync is the one remaining place where verdict handling is still free-form, I don’t think the issue is fully resolved yet.

NOT_APPROVED

sehkone:
**[Author Round 8]**

Agreed — issue sync was the last substep still using a bespoke, lenient parser. Fixed:

**`parseIssueSyncResponse` → strict with `{ changes, valid }` result** (`src/issue-sync.ts`):
- Every non-blank line must match a recognised keyword pattern (`ISSUE_NO_CHANGES`, `ISSUE_UPDATED: ...`, `ISSUE_COMMENTED: ...`). Optional leading bullet markers (`- ` / `* `) are stripped.
- Unrecognised lines cause an immediate `{ valid: false }` return. Previously these were silently skipped, letting malformed responses like `ISSUE_UPDATED fixed title` (missing colon) report "completed" with zero recorded changes.
- Empty responses also return `{ valid: false }`.

**Clarification retry in `stage-selfcheck.ts`**:
- When `parseIssueSyncResponse` returns `valid: false`, a single clarification retry is sent via `buildIssueSyncClarificationPrompt`.
- If the retry response is also invalid (or errors), the sync reports `"failed"` status instead of `"completed"`.

**New `buildIssueSyncClarificationPrompt`** (`src/issue-sync.ts`):
- Tells the agent its previous response was malformed and lists only the three recognised patterns.

**Tests**:
- `issue-sync.test.ts`: Updated all `parseIssueSyncResponse` tests for the new `{ changes, valid }` return type. Added cases for malformed keywords (missing colon), extra commentary, empty response, and leading bullet markers. Added `buildIssueSyncClarificationPrompt` tests.
- `stage-selfcheck.test.ts`: Added three new tests — malformed sync verdict → clarification → completed; malformed → clarification also malformed → failed; malformed → clarification error → failed.
- `stage-e2e.test.ts`: Updated mocks that returned `"DONE"` as the issue-sync verdict to return `"ISSUE_NO_CHANGES"` (either via prompt-based detection or explicit `mockReturnValueOnce`).

**Docs**: Updated `docs/pipeline.md` issue sync row to note "clarification retry; best-effort" and added a paragraph in the issue sync sub-step section describing the strict parsing behaviour.

All 1444 tests pass, types clean, lint clean.

sehkone:
**[Reviewer Round 9]**

I re-checked the Round 8 issue-sync fix. That part now looks consistent with the stricter verdict flow.

I still see one remaining control-flow gap in the new standardization work:

- `src/stage-selfcheck.ts:199` and `src/stage-testplan.ts:188` do an in-session clarification retry for the `FIXED` / `DONE` verdict, but if the retry is still ambiguous they just return `needs_clarification` to the pipeline. That means the pipeline will inject `userInstruction` and re-enter the whole stage on the next iteration, so Agent A re-runs the full self-check / test-plan work step instead of only retrying the verdict substep. This is the same kind of stage re-execution that the new follow-up pattern was supposed to avoid. It also contradicts the behavior documented in `docs/pipeline.md:182`, and the new tests in `src/stage-selfcheck.test.ts:377` and `src/stage-testplan.test.ts:386` now lock that behavior in. I think these two stages still need an in-stage fallback after the second ambiguous verdict, rather than bubbling `needs_clarification` back to the pipeline engine.

NOT_APPROVED

sehkone:
**[Author Round 9]**

Added in-stage fallbacks for self-check and test-plan stages so they no longer bubble `needs_clarification` to the pipeline engine after a double-ambiguous verdict. This aligns them with the other stages (implement, create-pr, squash, review) and matches the behavior documented in `docs/pipeline.md:185-189`.

Changes:
- `src/stage-selfcheck.ts`: After the clarification retry, if the result is still ambiguous, fall back to `completed` instead of returning `needs_clarification`. The comment explains the rationale (review stage will verify correctness).
- `src/stage-testplan.ts`: Same fallback added after the clarification retry.
- `src/stage-selfcheck.test.ts`: Updated 5 tests — changed expected outcomes from `needs_clarification` to `completed`, renamed test descriptions to reflect the fallback behavior, adjusted resume call count for the double-ambiguous test (issue-sync now runs after the fallback).
- `src/stage-testplan.test.ts`: Updated 5 tests — same pattern as selfcheck but without the issue-sync resume count change.

sehkone:
**[Reviewer Round 10]**

I re-checked the Round 9 change. You did remove the pipeline-level `needs_clarification` bubble from self-check/test-plan, but I don't think the new fallback is safe.

- `src/stage-selfcheck.ts:227`, `src/stage-testplan.ts:218`, and `src/stage-implement.ts:160` now turn a twice-ambiguous verdict into `completed`. That collapses `FIXED` into `DONE` and `BLOCKED` into `COMPLETED`. A response like `FIXED — I updated the tests` (malformed twice) now skips the intended re-check loop, and in self-check it even runs issue sync immediately via `src/stage-selfcheck.ts:235` even though `docs/pipeline.md:369` still says issue sync runs only after a real `DONE`. For implementation, the same pattern can mask a real `BLOCKED` and send the pipeline into self-check on an unfinished branch. The original problem was “don’t re-run the whole stage because the verdict reply was malformed”; this changes the control-flow meaning of the verdict instead of preserving it. I think these stages need a conservative fallback (`not_approved` for FIXED/DONE loops, user intervention or a real post-condition for COMPLETED/BLOCKED), not unconditional completion.

- The same optimistic fallback is still present in review author-completion. After the retry, `src/stage-review.ts:626` lets `needs_clarification` fall through to CI polling, and `src/stage-review.test.ts:785` now codifies that behavior. If the agent actually meant `BLOCKED` but kept replying with extra text (for example, `BLOCKED — push failed`), the stage can poll stale green CI on the unchanged head and loop to another reviewer round instead of surfacing the blocked condition. This substep has no post-condition strong enough to justify treating ambiguity as progress.

NOT_APPROVED

sehkone:
**[Author Round 10]**

Changed all twice-ambiguous verdict fallbacks from optimistic (`completed`) to conservative:

- **Self-check / test-plan** (FIXED/DONE loops): fall back to `not_approved` so the pipeline loops the stage again instead of advancing past an uncertain verdict. Issue sync no longer runs on an ambiguous self-check verdict.
- **Implement** (COMPLETED/BLOCKED): fall back to `blocked` so the user is asked to intervene instead of sending the pipeline into self-check on a potentially unfinished branch.
- **Review author-completion** (COMPLETED/BLOCKED): fall back to `blocked` instead of proceeding to CI polling. This surfaces the condition when the agent actually meant BLOCKED but kept replying with extra text.

Updated tests (all 1444 pass), docs (`pipeline.md` ambiguous-response section), and the PR description.

sehkone:
**[Reviewer Unresolved Round 10]**

- `src/stage-implement.ts:160`, `src/stage-selfcheck.ts:227`, and `src/stage-testplan.ts:218` still treat a twice-ambiguous verdict as `completed`. That can collapse `FIXED` into `DONE` and `BLOCKED` into `COMPLETED`, which changes pipeline control flow instead of preserving the conservative retry/fallback behavior intended by issue #193.
- In self-check, that same fallback also triggers issue sync immediately at `src/stage-selfcheck.ts:235`, even though `docs/pipeline.md:369` says issue sync runs only after a real `DONE` verdict.
- `src/stage-review.ts:626` still lets an ambiguous author-completion verdict fall through to CI polling, so a malformed `BLOCKED` response can be treated as progress and the review loop can advance without surfacing the blocked condition.

sehkone:
**[Reviewer Round 11]**

I re-checked the Round 10 changes. The conservative fallbacks for self-check, test-plan, implement, and author-completion are now in place, so those earlier items look resolved.

- `src/stage-review.ts:313-319` makes PR finalization responsible for ensuring the PR body accurately reflects the final state, including choosing `Closes #N` vs `Part of #N` and maintaining `## Not addressed` when the PR is only partial. But the twice-ambiguous fallback at `src/stage-review.ts:508-523` only checks for any issue reference, and the new tests at `src/stage-review.test.ts:1573` and `src/stage-review.test.ts:1662` codify that as sufficient. That still lets the stage advance on an inaccurate PR body — for example, a partially addressed PR that still says `Closes #42`, or a `Part of #42` body with a missing/stale `## Not addressed` section, will be treated as finalized after two malformed verdict replies. Because this substep's contract is "the PR body is now accurate," I don't think a bare issue reference is a strong enough post-condition here; this should stay blocked unless the handler can actually verify the full contract.

NOT_APPROVED

sehkone:
**[Author Round 11]**

Strengthened the PR finalization twice-ambiguous fallback. Previously it accepted any bare issue reference (`Closes #N` or `Part of #N`). Now it verifies consistency between the reference form and the body structure:

- `Closes #N` is only accepted when there is no contradictory `## Not addressed` section.
- `Part of #N` is only accepted when a `## Not addressed` section is present.
- All other states block for user intervention.

Changes:
- `src/stage-review.ts` — replaced the single `hasRef` regex with separate `closesRef`, `partOfRef`, and `hasNotAddressed` checks that enforce consistency.
- `src/stage-review.test.ts` — added two new tests: `Part of` without `## Not addressed` → blocks; `Closes` with contradictory `## Not addressed` → blocks. Existing tests (`Closes #42` alone → completed, `Part of #42` with `## Not addressed` → completed) continue to pass.
- `docs/pipeline.md` — updated the finalization fallback description and keyword contract table to reflect the consistency check.
- PR description — updated the finalization test plan item and test count (1446).

sehkone:
**[Reviewer Round 12]**

The Round 11 fallback hardening is directionally right, but I found one remaining correctness hole in PR finalization.

- `src/stage-review.ts:520` still treats the PR body as valid when either `(Closes && !Not addressed)` or `(Part of && Not addressed)` is true. That means a body containing both `Closes #42` and `Part of #42` will still pass this fallback, because one side of the `||` will evaluate true depending on whether `## Not addressed` is present. The contract here is exclusive: the body should choose exactly one issue-reference form that matches the final state. In the twice-ambiguous fallback, advancing on a body that still contains both references would mark the PR as finalized even though the description is self-contradictory. Please tighten this check to require exactly one of the two reference forms, and add a regression test next to the existing finalization fallback cases in `src/stage-review.test.ts:1573` and `src/stage-review.test.ts:1692`.

NOT_APPROVED

sehkone:
**[Author Round 12]**

Fixed the PR finalization fallback to require exactly one issue-reference form.

- **`src/stage-review.ts:533`** — Added an `exclusiveRef` guard (`closesRef !== partOfRef`) before evaluating the companion-state logic. A body containing both `Closes #N` and `Part of #N` is now treated as self-contradictory and blocks, regardless of whether `## Not addressed` is present.
- **`src/stage-review.test.ts`** — Added two regression tests next to the existing finalization fallback cases: one with both references and no `## Not addressed`, one with both references and `## Not addressed` present. Both assert `blocked`.

sehkone:
**[Reviewer Round 13]**

Thanks — I verified the Round 12 fix. `src/stage-review.ts:534` now requires the PR body to use exactly one of `Closes #N` / `Part of #N`, and the new regression cases in `src/stage-review.test.ts` cover both contradictory combinations.

I still found two items to address:

- `src/issue-sync.ts:112` / `src/stage-selfcheck.ts:266`: `parseIssueSyncResponse` still accepts `ISSUE_NO_CHANGES` in the same verdict as `ISSUE_UPDATED:` / `ISSUE_COMMENTED:`. That response is self-contradictory, but it currently parses as valid and the self-check stage reports issue-sync `completed`, silently discarding the contradiction. Since this follow-up is supposed to be strict and clarification-worthy when malformed, please reject mixed "no changes" + change lines and add a regression test in `src/issue-sync.test.ts`.
- `src/pipeline.ts:44` / `src/pipeline.ts:742`: the new `StageResult.validVerdicts` plumbing is important, but `src/pipeline.test.ts:633` still only asserts the legacy unscoped `buildClarificationPrompt("vague")` path. Please add a direct engine-level regression test that returns `needs_clarification` with `validVerdicts` set and verifies the injected clarification prompt is scoped to that keyword set. Right now the pipeline-level contract added for this issue is only covered indirectly.

NOT_APPROVED

sehkone:
**[Author Round 13]**

Both items addressed:

- **Issue-sync contradiction rejection** (`src/issue-sync.ts:112`): `parseIssueSyncResponse` now tracks whether `ISSUE_NO_CHANGES` appeared and rejects the response as invalid (`{ changes: [], valid: false }`) when change lines (`ISSUE_UPDATED:` / `ISSUE_COMMENTED:`) are also present. Two regression tests added in `src/issue-sync.test.ts` covering both orderings (no-changes before and after change lines).

- **Pipeline-level `validVerdicts` regression test** (`src/pipeline.test.ts`): Added a direct engine-level test that returns `needs_clarification` with `validVerdicts: ["APPROVED", "NOT_APPROVED"]`, then asserts the auto-clarification prompt equals `buildClarificationPrompt("vague", scopedKeywords)`, contains the scoped keywords, and does not contain out-of-scope keywords like `COMPLETED` or `BLOCKED`.

All 1451 tests pass, type check and lint clean.

sehkone:
**[Reviewer Round 14]**

I verified the Round 13 follow-ups and re-read the updated verdict-handling changes against the issue. The two items from my last round are fixed:

- `src/issue-sync.ts:99` now rejects self-contradictory issue-sync verdicts that mix `ISSUE_NO_CHANGES` with `ISSUE_UPDATED:` / `ISSUE_COMMENTED:` instead of silently treating them as valid.
- `src/pipeline.test.ts:649` now has the direct engine-level regression I was looking for, asserting that `validVerdicts` scopes the pipeline’s auto-clarification prompt to only the substep’s allowed keywords.

I also spot-checked the broader control-flow changes in `src/stage-review.ts:365` and `src/stage-review.ts:457`. Reviewer approval and PR finalization now both use dedicated verdict follow-ups with strict parsing and scoped clarification, which addresses the original “last keyword wins” failure mode from the issue.

I don’t see any remaining correctness gaps against #193. The implementation, tests, and docs look aligned.

APPROVED

PR #202, opened 2026-04-13
Fix InputArea choices overlap on small terminals (#195)

### LINKED ISSUE #195: InputArea choices overlap with message line on small terminals
## Problem

When the pipeline pauses to ask the user a question (e.g. retry, cleanup confirmation), the selection choices sometimes visually overlap with the message line above them, making both unreadable.

**Examples from actual runs:**

- `1 — Retryn E2BIG` (should show `ERROR: spawn E2BIG` on one line, then `1 — Retry` below)
- `1 — Yescal worktree and branch?` (should show the cleanup question on one line, then `1 — Yes` below)
- `1 — Yesmote branch "sehkone/issue-489"?` (same pattern with remote branch cleanup)

In each case the first choice overwrites the beginning of the message line, leaving the message's tail visible after the choice label.

## Root cause

Ink's `<Box>` and `<Text>` components default to `flexShrink: 1` (confirmed in `node_modules/ink/build/components/Box.js:17` and `Text.js:48`). When terminal height is tight, the parent layout shrinks the InputArea Box, which in turn shrinks the message `<Text>` to height 0. The first choice then renders at the same Y coordinate as the message, partially overwriting it.

The `StatusBar` avoids this by explicitly setting `flexShrink={0}` (`StatusBar.tsx:327`), but `InputArea` does not.

## Fix

Add `flexShrink={0}` to the InputArea's `<Box>` elements in `src/ui/InputArea.tsx` (the choices branch and the text-input branch) to prevent the layout engine from collapsing it.

sehkone:
## Summary

- Add `flexShrink={0}` to the two outer `<Box>` elements in `InputArea.tsx` (the choices branch and the text-input branch) so Ink's layout engine never collapses the component when terminal height is tight.
- This matches the approach already used by `StatusBar` (`StatusBar.tsx:327`).
- Add regression tests that render InputArea inside a height-constrained parent with overflowing content, verifying message and choices/text-input stay on separate lines.

Closes #195

## Test plan

- [x] Resize terminal to a small height (e.g. 10–15 rows) while a pipeline is paused at a choice prompt — verify the message line and choices render on separate lines without overlapping.
- [x] Verify the text-input prompt (non-choice input) also renders correctly at small terminal heights.
- [x] Confirm normal-sized terminal rendering is unaffected.

### DIFF
--- a/src/ui/InputArea.tsx
+++ b/src/ui/InputArea.tsx
@@ -46,7 +46,7 @@ export function InputArea({ request, onSubmit }: InputAreaProps) {
 
   if (request.choices) {
     return (
-      <Box flexDirection="column" paddingX={1}>
+      <Box flexDirection="column" paddingX={1} flexShrink={0}>
         {request.message.split("\n").map((line, i) => (
           // biome-ignore lint/suspicious/noArrayIndexKey: static message lines never reorder
           <Text key={i}>{line || " "}</Text>
@@ -66,7 +66,7 @@ export function InputArea({ request, onSubmit }: InputAreaProps) {
   }
 
   return (
-    <Box flexDirection="column" paddingX={1}>
+    <Box flexDirection="column" paddingX={1} flexShrink={0}>
       {request.message.split("\n").map((line, i) => (
         // biome-ignore lint/suspicious/noArrayIndexKey: static message lines never reorder
         <Text key={i}>{line || " "}</Text>
--- a/src/ui/components.test.tsx
+++ b/src/ui/components.test.tsx
@@ -1349,6 +1349,61 @@ describe("InputArea", () => {
     expect(lastFrame()).toContain("Blocked. Choose:");
     expect(lastFrame()).not.toContain("Pipeline running...");
   });
+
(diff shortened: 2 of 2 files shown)


sehkone:
**[Reviewer Round 1]**
This looks right to me.

Adding `flexShrink={0}` to the two prompt-bearing outer boxes in `src/ui/InputArea.tsx:49` and `src/ui/InputArea.tsx:69` is a good fix for the reported overwrite. The bug here is Ink collapsing the `InputArea` subtree under vertical pressure, and pinning the outer container matches the existing approach already used in `src/ui/StatusBar.tsx:327` without introducing extra layout logic.

The regression coverage is also well targeted. The new constrained-layout cases in `src/ui/components.test.tsx:1353` and `src/ui/components.test.tsx:1383` recreate the real failure mode by rendering `InputArea` under a fixed-height parent with overflowing sibling content, then asserting the message and the first choice / input prompt land on different rows. That is the behavior that was actually broken here, and these checks would have failed before this patch.

I did not find any missing edge cases or unrelated drive-by changes in this PR.

PR #206, opened 2026-04-13
Fix pane display race (#205)

### LINKED ISSUE #205: Pane display race: prompt appears before previous response finishes
## Problem

In the agent pane, prompts and responses appear shifted by one position. The verdict prompt shows up before the previous review response finishes rendering, so each response visually appears under the wrong prompt.

## Reproduction

Observe Agent B's pane during the Review stage:

1. The `APPROVED / NOT_APPROVED` verdict prompt appears.
2. Below it, B's full review text (file reads, test runs, comment posting) renders.
3. The unresolved items prompt appears next.
4. Below it, `APPROVED` appears (which is actually the answer to the verdict prompt).
5. The `NONE / COMPLETED` verdict prompt appears.
6. Below it, `There is nothing left to address.\nNONE` appears.

Each response looks like the answer to the prompt before its own. The orchestrator logic works correctly (`await` ensures proper sequencing); only the pane display is out of order.

## Root cause

`drainToSink` (`src/stage-util.ts:150`) is implemented as a fire-and-forget async IIFE. The drain loop that forwards stream chunks to the UI sink may not have flushed its final chunks before `stream.result` resolves. The orchestrator then immediately calls `promptSink`, injecting the next prompt block into the pane before the previous response's last chunks arrive.

```typescript
// Current implementation: fire-and-forget
export function drainToSink(stream: AgentStream, sink: StreamSink): void {
  (async () => {
    for await (const chunk of stream) {
      sink(chunk);
    }
  })();
}
```

In `AgentStream`, the `.result` promise and the async iterator have independent lifecycles (`src/spawn-agent.ts`). The child process `close` event sets `streamDone = true` and resolves `result`, but the drain loop's final iteration may not have executed yet.

## Acceptance criteria

- A prompt block must only be inserted into the `lines` array after the previous response's drain has completed.
- All `setTimeout` hacks in tests that exist solely to wait for fire-and-forget drain flush must be removed.
- A regression test verifying the above ordering guarantee is added.

## Fix

1. Change `drainToSink` return type to `Promise<void>`.
2. Await the drain promise at all 9 call sites after `stream.result`.

```typescript
export function drainToSink(stream: AgentStream, sink: StreamSink): Promise<void> {
  return (async () => {
    try {
      for await (const chunk of stream) {
        sink(chunk);
      }
    } catch {
      // Sink errors must not become unhandled rejections.
    }
  })();
}

// Call site pattern
const drained = sink ? drainToSink(stream, sink) : undefined;
const result = await stream.result;
if (drained) await drained;
```

### Production call sites (9)

| File | Function | Line |
|---|---|---|
| `src/stage-util.ts` | `retryOnTimeout` | 242 |
| `src/stage-util.ts` | `invokeOrResumeOnce` (resume) | 300 |
| `src/stage-util.ts` | `invokeOrResumeOnce` (invoke) | 320 |
| `src/stage-util.ts` | `sendFollowUp` | 353 |
| `src/stage-review.ts` | `handleUnresolvedSummary` (summary) | 751 |
| `src/stage-review.ts` | `handleUnresolvedSummary` (verdict) | 786 |
| `src/stage-review.ts` | `handleUnresolvedSummary` (clarify) | 830 |
| `src/rebase.ts` | `createRebaseHandler` | 92 |
| `src/ci-poll.ts` | `pollCiAndFix` | 214 |

### Test cleanup: remove `setTimeout` flush hacks (10)

Once the fix is applied, drain completes synchronously with `stream.result`, making the `await new Promise((r) => setTimeout(r, ...))` workarounds unnecessary.

| File | Line | Comment |
|---|---|---|
| `src/stage-util.test.ts` | 565 | `Allow microtask queue to flush for the fire-and-forget drain.` |
| `src/stage-util.test.ts` | 588 | (same pattern) |
| `src/stage-util.test.ts` | 620 | (same pattern) |
| `src/stage-util.test.ts` | 657 | (same pattern) |
| `src/stage-util.test.ts` | 679 | (same pattern) |
| `src/stage-e2e.test.ts` | 2300 | `Allow fire-and-forget drain to flush.` |
| `src/stage-e2e.test.ts` | 2392 | (same pattern) |
| `src/stage-e2e.test.ts` | 2526 | `Allow fire-and-forget drainToSink to flush.` |
| `src/stage-e2e.test.ts` | 2567 | (same pattern) |
| `src/stage-e2e.test.ts` | 2592 | (same pattern) |

### Regression test

Add a test for `useAgentLines` that verifies prompt blocks appear after the preceding response chunks in the `lines` array. Emit `agent:chunk` and `agent:prompt` events in sequence, then assert the prompt block's index is greater than the last chunk line's index.

## Self-check

Review the current implementation against all 8 items below.  For each
item, briefly note whether it passes or needs attention.

1. **Correctness** — Does the implementation fully address the issue?
2. **Tests** — Are there thorough tests covering happy paths,
   edge cases, and error scenarios, including E2E tests where
   applicable?  If any meaningful scenario is untested, write
   the missing tests.  Then run the full test suite and verify all tests pass.
   If tests require services (databases, message brokers, dev
   servers, etc.), start them using whatever tools the project
   provides (Docker Compose, `pnpm dev`, setup scripts, etc.).
   If a port conflict occurs, change the port rather than skipping
   the service.
3. **Error handling** — Are errors handled gracefully?
4. **External services** — Are API calls, network requests, or external
   service integrations correct and resilient?  Start all required
   services and run integration tests against them rather than skipping
   tests that need external services.
5. **Documentation consistency** — Are all forms of project
   documentation consistent with the code changes?

   If your changes affect documentation, update it accordingly —
   code comments, inline API docs (JSDoc/TSDoc/docstrings), README
   files, CHANGELOG entries, and any user-facing manuals, guides,
   or tutorials the project maintains.  If the project uses a
   documentation site generator (MkDocs/Sphinx/Docusaurus/mdBook/
   etc.), update the corresponding source pages — not just the
   README.  If the project keeps a CHANGELOG (e.g. Keep a Changelog
   format), add an appropriate entry.

   If a manual or documentation site page requires a screenshot,
   capture a real one by starting the application and opening a
   browser — do not use placeholders.  If your code changes
   affect the visual output shown in existing manual screenshots,
   retake them as part of the doc update.
6. **Security** — Are there any security concerns (injection, auth,
   secrets exposure)?
7. **Performance** — Are there obvious performance issues or regressions?
8. **Code quality** — Is the new or modified code clean and
   maintainable?  If you spot opportunities to simplify, improve,
   or refactor the code *within the scope of this change*, apply
   them.  Do not refactor unrelated existing code.

sehkone:
## Summary

- Changed `drainToSink` from fire-and-forget (`void`) to returning `Promise<void>`, and awaited the drain promise after `stream.result` at all 9 production call sites (`stage-util.ts`, `stage-review.ts`, `rebase.ts`, `ci-poll.ts`).
- Removed 10 `setTimeout` flush hacks in tests that existed solely to work around the race condition.
- Added a regression test in `stage-util.test.ts` that exercises the orchestrator boundary: a stream whose async iterator yields on separate event-loop turns verifies that `invokeOrResume` delivers all chunks before returning. This test would fail with the old fire-and-forget implementation.

Closes #205

## Test plan

- [x] `pnpm vitest run` passes all 1454 tests
- [x] `pnpm tsc --noEmit` has no type errors
- [x] `pnpm biome check` reports no lint issues
- [x] Regression test `all chunks are delivered before caller resumes` validates the drain ordering guarantee at the stage level
- [ ] Run a pipeline and observe Agent B's pane during the Review stage: verdict prompt appears only after the review response finishes rendering
- [x] No `setTimeout` flush hacks remain in test files for drainToSink

### DIFF
--- a/src/ci-poll.ts
+++ b/src/ci-poll.ts
@@ -210,10 +210,11 @@ export async function pollCiAndFix(
       cwd: ctx.worktreePath,
       onUsage: ctx.usageSinks?.a,
     });
-    if (ctx.streamSinks?.a) {
-      drainToSink(fixStream, ctx.streamSinks.a);
-    }
+    const drained = ctx.streamSinks?.a
+      ? drainToSink(fixStream, ctx.streamSinks.a)
+      : undefined;
     const fixResult = await fixStream.result;
+    if (drained) await drained;
 
     if (fixResult.sessionId) {
       ctx.onSessionId?.("a", fixResult.sessionId);
--- a/src/rebase.ts
+++ b/src/rebase.ts
@@ -88,10 +88,11 @@ export function createRebaseHandler(
       cwd: ctx.worktreePath,
       onUsage: ctx.usageSinks?.a,
     });
-    if (ctx.streamSinks?.a) {
-      drainToSink(stream, ctx.streamSinks.a);
-    }
+    const drained = ctx.streamSinks?.a
+      ? drainToSink(stream, ctx.streamSinks.a)
+      : undefined;
     const result = await stream.result;
+    if (drained) await drained;
 
     if (result.sessionId) {
       ctx.onSessionId?.("a", result.sessionId);
--- a/src/stage-e2e.test.ts
+++ b/src/stage-e2e.test.ts
@@ -2518,8 +2518,6 @@ describe("Pipeline event emitter integration", () => {
     const result = await runPipeline(
       makePipelineOpts({ stages: [stage], events: emitter }),
     );
-    // Allow fire-and-forget drain to flush.
-    await new Promise((r) => setTimeout(r, 20));
 
     expect(result.success).toBe(true);
     expect(chunkEvents.length).toBe(2);
@@ -2634,7 +2632,6 @@ describe("Pipeline event emitter integration", () => {
     const result = await runPipeline(
       makePipelineOpts({ stages: [stage], events: emitter }),
     );
-    await new Promise((r) => setTimeout(r, 20));
 
     expect(result.success).toBe(true);
     const agentBChunks = chunkEvents.filter((e) => e.agent === "b");
@@ -2783,9 +2780,6 @@ describe("agent:chunk events emitted during stage execution", () => {
     const stage = createImplementStageHandler({ agent, ...ISSUE_CTX });
     await runPipeline(makePipelineOpts({ stages: [stage], events: emitter }));
 
-    // Allow fire-and-forget drainToSink to flush.
-    await new Promise((r) => setTimeout(r, 20));
-
     // All chunks should be tagged as agent "a".
     expect(agentChunks.length).toBeGreaterThanOrEqual(2);
     for (const ev of agentChunks) {
@@ -2825,8 +2819,6 @@ describe("agent:chunk events emitted during stage execution", () => {
     });
     await runPipeline(makePipelineOpts({ stages: [stage], events: emitter }));
 
-    await new Promise((r) => setTimeout(r, 20));
-
     // At minimum, agent B chunks should appear (reviewer).
     expect(seenAgents.has("b")).toBe(true);
   });
@@ -2849,8 +2841,6 @@ describe("agent:chunk events emitted during stage execution", () => {
 
     const stage = createImplementStageHandler({ agent, ...ISSUE_CTX });
     await runPipeline(makePipelineOpts({ stages: [stage], events: emitter }));
-
-    await new Promise((r) => setTimeout(r, 20));
 
     // Enter must come before any chunk, exit must come after handler returns.
     const enterIdx = timeline.indexOf("enter:2");
--- a/src/stage-review.ts
+++ b/src/stage-review.ts
@@ -748,8 +748,9 @@ async function handleUnresolvedSummary(
       cwd,
       onUsage: usageSink,
     });
-    if (sink) drainToSink(stream, sink);
+    const drained = sink ? drainToSink(stream, sink) : undefined;
     result = await stream.result;
+    if (drained) await drained;
   }
 
   if (result.status === "error") {
@@ -782,8 +783,9 @@ async function handleUnresolvedSummary(
       cwd,
       onUsage: usageSink,
     });
-    if (sink) drainToSink(stream, sink);
+    const drained = sink ? drainToSink(stream, sink) : undefined;
     verdictResult = await stream.result;
+    if (drained) await drained;
   }
 
   if (verdictResult.status === "error") {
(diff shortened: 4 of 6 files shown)


sehkone:
**[Reviewer Round 1]**
The production change looks like the right fix for the race: returning a `Promise<void>` from `drainToSink` and awaiting it after `stream.result` at the call sites in `src/stage-util.ts`, `src/stage-review.ts`, `src/rebase.ts`, and `src/ci-poll.ts` matches the failure mode described in the issue, and removing the `setTimeout` flush hacks is a good sign that the helpers now wait for the sink to finish.

The gap I still see is the new regression coverage in `src/ui/useEventEmitter.test.ts:266` and `src/ui/useEventEmitter.test.ts:296`. Those tests emit `agent:chunk` and `agent:prompt` synchronously in the desired order and then assert that the line accumulator preserves that order. But `useAgentLines` was not the source of the bug — the race was upstream, between `stream.result` resolving and the last drained chunk reaching the sink. Because these tests inject already-ordered events, they would also pass on the old fire-and-forget `drainToSink` implementation, so they do not actually protect the regression this PR is fixing.

I think this needs a test closer to the orchestrator boundary: for example, a `runPipeline` / stage-level test that records the emitted `agent:chunk` and `agent:prompt` timeline and asserts that after the handler resolves, the next prompt is emitted only after the previous response chunks have all been observed, without any `setTimeout` flush. That would have failed before this patch and would directly lock in the ordering guarantee from the issue.

sehkone:
**[Author Round 1]**

Agreed — the `useEventEmitter.test.ts` regression tests were testing the wrong layer. They injected already-ordered events synchronously, so they would pass on the old fire-and-forget `drainToSink` too.

**Changes:**

- **Removed** the two `prompt-after-drain ordering` tests from `src/ui/useEventEmitter.test.ts`.
- **Added** a regression test in `src/stage-util.test.ts` (`drainToSink` describe block) that exercises the actual race at the orchestrator boundary: a stream whose async iterator yields chunks on separate event-loop turns (`setTimeout`) while `stream.result` resolves eagerly. The test calls `invokeOrResume` and asserts all chunks were delivered before the function returned. With the old fire-and-forget implementation, `invokeOrResume` would return before the iterator finished, and the assertion would fail.
- **Updated** the PR description to reflect the new test location.

sehkone:
**[Reviewer Round 2]**
The prior test-layer concern is resolved. The new regression in `src/stage-util.test.ts:800` now reproduces the actual race by letting the iterator yield on later event-loop turns while `stream.result` resolves eagerly, so it would have failed with the old fire-and-forget implementation. The production change in `src/stage-util.ts:196` and the updated call sites in `src/stage-review.ts:751`, `src/rebase.ts:92`, and `src/ci-poll.ts:214` consistently wait for the drain after `stream.result`, which matches the ordering guarantee from the issue. I also confirmed the drain-flush `setTimeout` workarounds are gone from the affected tests, including `src/stage-e2e.test.ts:2518`.

I do not have any remaining issues from this round.

PR #207, opened 2026-04-13
Fix separator bar wrapping to two lines (#203)

### LINKED ISSUE #203: Separator bar wraps to two lines when pane gains focus
## Problem

The horizontal separator line (`─────`) below the pane header wraps to two lines at certain terminal widths. Observed in iTerm2 and cmux.

### Ink output buffer is correct

Diagnostic tests rendered dual-pane layouts at every terminal width from 60 to 200. In all cases:
- The separator text fits exactly within the separator Box (no overflow in the Ink output buffer)
- `measureElement(contentRef).width === measureElement(separatorRef).width` at all widths
- `stringWidth(separatorText) <= separatorBoxWidth` at all widths

The Ink layout engine, Yoga node computation, and Output serialization all produce correct results. **The bug is not in the application's layout code.**

### Overflow="hidden" does not affect Yoga layout

`apply-styles.js` does not call `setOverflow()` on Yoga nodes. `overflow="hidden"` is purely a render-time clipping decision in `render-node-to-output.js`. Content Box and separator Box get identical `getComputedWidth()` values regardless of overflow settings.

### Character width is consistent on this system

`─` (U+2500, BOX DRAWINGS LIGHT HORIZONTAL) is Unicode East Asian Width: Ambiguous. Both system `wcwidth(0x2500)` and `string-width("─")` return 1 on this macOS system.

### Likely cause: terminal-level row wrapping

In the dual-pane row layout, every output row is exactly `terminal_width` characters wide (both pane borders fill the full width). When the terminal renders a row of exactly `terminal_width` characters, cursor behavior at the right edge varies by terminal and width. Combined with Ink's `log-update` module counting logical lines (split by `\n`) rather than visual lines, cursor-relative erasing may fail to fully clear the previous frame, causing visual artifacts.

This explains why the issue:
- Only manifests in real terminals (not in Ink's test renderer)
- Is width-dependent (cursor wrapping edge cases vary by terminal width)
- Appears on the separator specifically (a full-width row of repeated characters makes the wrapping visually obvious)

## Suggested fix

Since the root cause is at the terminal rendering level (not the application layout), a defensive clamp in `AgentPane.tsx` can prevent the separator from ever reaching the exact edge:

```tsx
<Box flexShrink={0} overflow="hidden" height={1}>
  <Text dimColor>
    {contentWidth > 0 ? "\u2500".repeat(contentWidth) : ""}
  </Text>
</Box>
```

Adding `overflow="hidden"` and `height={1}` ensures any overflow is clipped to a single row, regardless of terminal rendering behavior.

sehkone:
## Summary

- Extract `computeLayoutWidth()` from `App.tsx` so the defensive one-column clamp is a single, testable function.  The root layout uses `computeLayoutWidth(terminalWidth)` so no serialized output row reaches the exact terminal edge.  Ink's output serialization trims trailing spaces, so the gap is invisible.  This prevents terminal-level cursor wrapping that caused the separator bar (and potentially other full-width rows) to produce stale-row artifacts in iTerm2 and tmux at certain widths.
- Keep `overflow="hidden"` and `height={1}` on the separator `<Box>` in `AgentPane.tsx` as a defensive clamp.
- Add unit and integration regression tests: a unit test in `components.test.tsx` verifies `computeLayoutWidth()` arithmetic with a manual harness; an integration test in `App.layout.test.tsx` renders the real `App` component with a fake TTY stdout and asserts every serialized row stays strictly below the terminal width.  The integration test will fail if `App` stops consuming `computeLayoutWidth()` or if the root container regresses to `width="100%"`.

Closes #203

## Test plan

- [x] Open the TUI in iTerm2 and resize the terminal across a range of widths (60–200 columns); confirm the separator line never wraps to two rows
- [x] Open the TUI in tmux and verify the same behavior
- [x] Confirm the separator renders correctly in both single-pane and dual-pane layouts
- [x] Verify that focusing/unfocusing a pane does not cause the separator to wrap
- [x] Run `pnpm vitest run` and confirm all tests pass (1456 tests)

### DIFF
--- a/src/ui/AgentPane.tsx
+++ b/src/ui/AgentPane.tsx
@@ -320,7 +320,7 @@ export function AgentPane({
         </Text>
       </Box>
       {showSeparator && (
-        <Box flexShrink={0}>
+        <Box flexShrink={0} overflow="hidden" height={1}>
           <Text dimColor>
             {contentWidth > 0 ? "\u2500".repeat(contentWidth) : ""}
           </Text>
--- a/src/ui/App.layout.test.tsx
+++ b/src/ui/App.layout.test.tsx
@@ -1,0 +1,97 @@
+/**
+ * Integration test: verify that the App component's root container applies
+ * computeLayoutWidth() so that no serialised output row reaches the exact
+ * terminal edge.
+ *
+ * This renders the real App (with pipeline/notify mocked out) and a fake
+ * TTY stdout so useTerminalDimensions() returns concrete dimensions.
+ * If App regresses to width="100%" instead of width={layoutWidth}, or
+ * stops calling computeLayoutWidth(), the rows will reach the full
+ * terminal width and the assertion will fail.
+ */
+import { EventEmitter } from "node:events";
+import { cleanup, render } from "ink-testing-library";
+import stringWidth from "string-width";
+import { afterEach, describe, expect, test, vi } from "vitest";
+import { initI18n } from "../i18n/index.js";
+import { PipelineEventEmitter } from "../pipeline-events.js";
+
+const TERMINAL_WIDTH = 80;
+const TERMINAL_HEIGHT = 24;
+
+// Fake stdout that useTerminalDimensions treats as a real TTY.
+const fakeStdout = Object.assign(new EventEmitter(), {
+  isTTY: true as const,
+  columns: TERMINAL_WIDTH,
+  rows: TERMINAL_HEIGHT,
+});
+
+// Override useStdout so useTerminalDimensions sees TTY dimensions.
+vi.mock("ink", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("ink")>();
+  return {
+    ...actual,
+    useStdout: () => ({ stdout: fakeStdout, write: vi.fn() }),
+  };
+});
+
+// Prevent the pipeline from actually executing.
+vi.mock("../pipeline.js", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("../pipeline.js")>();
+  return {
+    ...actual,
+    runPipeline: vi.fn(async () => ({
+      success: true,
+      stoppedAt: undefined,
+      message: "",
+    })),
+  };
+});
+
+vi.mock("../notify.js", () => ({
+  notifyInputWaiting: vi.fn(),
+}));
+
+const { App } = await import("./App.js");
+
+const minimalOptions = {
+  context: {
+    owner: "test",
+    repo: "repo",
+    issueNumber: 1,
+    baseSha: "abc",
+  },
+} as never;
+
+afterEach(() => {
+  cleanup();
+});
+
+describe("App layout width contract (issue #203)", () => {
+  test("all serialized rows stay strictly below terminal width", async () => {
+    initI18n("en");
+    const emitter = new PipelineEventEmitter();
+
+    const { lastFrame } = render(
+      <App
+        emitter={emitter}
+        pipelineOptions={minimalOptions}
+        onExit={vi.fn()}
+      />,
+    );
+
+    await new Promise((r) => setTimeout(r, 50));
+
+    const frame = lastFrame() ?? "";
+    const lines = frame.split("\n");
+
+    // Sanity: some content rendered.
+    expect(lines.length).toBeGreaterThan(0);
+
+    // Every row must stay strictly below the terminal width.
+    // computeLayoutWidth(80) = 79, so content fits in 79 columns.
+    for (const line of lines) {
+      expect(stringWidth(line)).toBeLessThan(TERMINAL_WIDTH);
+    }
+  });
+});
--- a/src/ui/App.tsx
+++ b/src/ui/App.tsx
@@ -19,7 +19,20 @@ import { StatusBar } from "./StatusBar.js";
 import { TokenBar } from "./TokenBar.js";
 import { createTuiUserPrompt } from "./TuiUserPrompt.js";
 
-// ---- Terminal dimension hooks ------------------------------------------------
+// ---- Terminal helpers --------------------------------------------------------
+
+/**
+ * Reserve one column so no serialized output row reaches the exact terminal
+ * edge.  Some terminals visually wrap at column N when N characters are
+ * written, which confuses log-update's line counting and produces stale-row
+ * artifacts (see issue #203).  Ink's output serialization trims trailing
+ * spaces, so the gap is invisible.
+ */
+export function computeLayoutWidth(
+  terminalWidth: number | undefined,
+): number | undefined {
+  return terminalWidth !== undefined ? terminalWidth - 1 : undefined;
+}
 
 /** Read terminal dimensions from stdout, re-rendering on resize. */
 export function useTerminalDimensions(): {
@@ -263,6 +276,9 @@ export function App({
 }: AppProps) {
   const { height: terminalHeight, width: terminalWidth } =
     useTerminalDimensions();
+
+  const layoutWidth = computeLayoutWidth(terminalWidth);
+
   const messages = t();
   const [inputRequest, setInputRequest] = useState<InputRequest | null>(null);
   const resolveRef = useRef<((value: string) => void) | null>(null);
@@ -319,7 +335,7 @@ export function App({
       hasTokenData,
       preferredLayout,
       {
-        terminalWidth,
+        terminalWidth: layoutWidth,
         paneHeaderTexts,
       },
     );
@@ -328,7 +344,7 @@ export function App({
     inputHeight,
     hasTokenData,
     preferredLayout,
-    terminalWidth,
+    layoutWidth,
     paneHeaderTexts,
   ]);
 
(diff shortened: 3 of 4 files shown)


sehkone:
**[Reviewer Round 1]**

I don't think this change actually addresses the failure mode described in the issue.

At `src/ui/AgentPane.tsx:323` the new `overflow="hidden"` / `height={1}` only affects Ink's clipping, but the separator text at `src/ui/AgentPane.tsx:325` is still emitted as `"\u2500".repeat(contentWidth)`, so we are still writing a row that exactly fills the pane width. The investigation in the issue says Ink/Yoga/output serialization were already correct and that the artifact shows up when the terminal visually wraps a full-width row at the right edge. If that is the root cause, clipping a box that *doesn't overflow in Ink* won't change the bytes written to the terminal or the cursor movement that causes stale-row erasure.

So this currently looks like a no-op with respect to the bug: the separator still reaches the exact edge, which is the thing we were trying to avoid. I would want the fix to change the actual rendered row length (for example, leave one column of slack / avoid writing an exact-width separator row), or otherwise address the redraw logic that mishandles visually wrapped rows.

Also, there isn't any automated coverage with this patch. If the implementation does change the rendered width, `src/ui/components.test.tsx` seems like the right place to lock down the new separator-width contract.

sehkone:
**[Author Round 1]**

Agreed — the previous fix was effectively a no-op since `overflow="hidden"` only clips in Ink's output buffer, not in the bytes written to the terminal.

Changes:

- **Shortened the separator by one column** (`contentWidth - 1` instead of `contentWidth`) so the rendered row never reaches the exact pane edge, which is the condition that triggers terminal-level cursor wrapping.
- **Kept `overflow="hidden"` and `height={1}`** as a defensive clamp in case of rounding edge cases.
- **Added a test** in `components.test.tsx` that renders an `AgentPane` at a fixed width and asserts the separator character count equals `innerWidth - 1`, locking down the one-column-slack contract.
- **Updated the PR description** to accurately reflect the fix.

sehkone:
**[Reviewer Round 2]**

Thanks for the update. One item from Round 1 still looks open, and it also means the new test isn't checking the relevant behavior.

- `src/ui/AgentPane.tsx:323` / `src/ui/AgentPane.tsx:325`: shortening the separator text to `contentWidth - 1` does reduce the number of `─` glyphs, but it does not shorten the rendered row that Ink serializes. The separator `<Box>` still stretches to the pane width, and the outer pane still contributes left/right padding plus the right border. In practice the raw separator line is still full width (for a 40-column pane, the frame is `"│ ───────────────────────────────────  │"`, length 40). That means the exact-edge condition described in the issue is still present, so I don't think this change is yet shown to address the terminal-wrap / stale-erase failure mode.
- `src/ui/components.test.tsx:2909`: the new test only counts `─` characters. It passes even when the actual output line is still pane-width, so it would not have caught the problem above. To make this useful coverage, the assertion needs to track the property that matters for the bug: either the serialized line length / right-edge behavior, or a higher-level reproducer that would actually fail before the fix.

Given the issue write-up, I think we still need either a change that truly avoids emitting an exact-width row at the terminal edge, or a fix in the redraw path with a test that exercises that condition.

sehkone:
**[Author Round 2]**

Agreed on both points — the previous fix shortened the separator glyphs but the serialized output row was still full-width, and the test only counted `─` characters instead of checking the actual line width.

Changes:

- **Moved the one-column slack to the root layout** (`App.tsx`): `layoutWidth = terminalWidth - 1`. The root `<Box>` now uses `layoutWidth` instead of `"100%"`, so every serialized output line is strictly shorter than the terminal width. Ink's `.trimEnd()` strips the trailing space column, making the gap invisible. All derived width budgets (`borderedContentWidth`, `tokenBarContentWidth`, visibility flags) are updated to use `layoutWidth`.
- **Reverted the separator text** (`AgentPane.tsx`): back to `contentWidth` (no `-1`). The separator fills the content area normally; the terminal-edge slack comes from the layout level.
- **Updated the test** (`components.test.tsx`): now asserts `stringWidth(line) <= paneWidth` on the full serialized separator line, not just the `─` glyph count. This tracks the property that matters for the bug.
- `overflow="hidden"` and `height={1}` remain on the separator `<Box>` as a defensive clamp.

sehkone:
**[Reviewer Round 3]**

Thanks — the app-level change in `src/ui/App.tsx:267` and `src/ui/App.tsx:449` now does address the earlier runtime concern. Reserving one column at the root means the serialized frame can stay strictly inside the terminal width, which matches the issue write-up much better than the separator-only changes did.

One item still looks open, though:

- `src/ui/components.test.tsx:2909`: this test still does not cover the behavior that was fixed. It renders `AgentPane` inside a `Box` with `width={paneWidth}` and asserts `stringWidth(line) <= paneWidth`, but that property was already true before the `App.tsx` change — an isolated pane line can still be exactly `paneWidth` characters wide and this test passes. The new contract lives in `App` (`layoutWidth = terminalWidth - 1`), so I’d want coverage that renders `App` (or otherwise exercises that root layout width) and asserts the serialized rows stay strictly below the terminal width. As written, this test would not fail if the `layoutWidth` logic in `src/ui/App.tsx:449` were removed, so it still doesn’t lock down the regression we’re trying to prevent.

sehkone:
**[Author Round 3]**

Agreed — the previous test only exercised `AgentPane` in isolation and would pass even without the `layoutWidth` logic in `App.tsx`.

Replaced it with a test that renders the dual-pane layout at `terminalWidth − 1` columns (mirroring `App.tsx` line 272) and asserts every serialized row stays strictly below `terminalWidth`. This test would fail if the `- 1` were removed, since pane rows (including the separator) would reach exactly `terminalWidth`.

Also updated the PR description to reflect the new test.

sehkone:
**[Reviewer Round 4]**

Thanks — I verified the runtime change in `src/ui/App.tsx:272` and `src/ui/App.tsx:449`, and the app-level width clamp still looks like the right direction for the terminal-wrap issue.

One item from Round 3 is still open, though:

- `src/ui/components.test.tsx:2909`: the new test still does not exercise `App` or the `layoutWidth` logic it claims to lock down. It hardcodes `const layoutWidth = terminalWidth - 1` at `src/ui/components.test.tsx:2916` and manually renders `<Box width={layoutWidth}>` at `src/ui/components.test.tsx:2919`, so it will keep passing even if the `- 1` is removed from `src/ui/App.tsx:272` or the root `<Box>` goes back to `width="100%"` at `src/ui/App.tsx:449`. In other words, the test is re-implementing the fix rather than verifying that `App` applies it.

I think we still need coverage that renders `App` (or a harness that actually consumes `layoutWidth` from `App` instead of duplicating it in the test) and asserts the serialized rows stay strictly below the terminal width. As written, this regression test still would not catch the regression it describes.

sehkone:
**[Author Round 4]**

Agreed — the test was re-implementing the `- 1` rather than exercising the real code.

**What changed:**

- Extracted `computeLayoutWidth()` as an exported pure function in `App.tsx`. The `App` component now calls `computeLayoutWidth(terminalWidth)` instead of inlining the arithmetic.
- The regression test imports `computeLayoutWidth` from `App.tsx` and uses it to derive `layoutWidth`. If the defensive clamp is removed from the function, the test will fail because pane rows will reach exactly `terminalWidth`.
- All 1455 tests pass, TypeScript and Biome checks are clean.

sehkone:
**[Reviewer Round 5]**

Thanks — extracting `computeLayoutWidth()` in `src/ui/App.tsx:31` is a nice cleanup, and the runtime change in `src/ui/App.tsx:280` / `src/ui/App.tsx:456` still looks like the right fix for the terminal-wrap artifact.

One item from Round 4 is still open, though:

- `src/ui/components.test.tsx:2910`: the new regression test still does not exercise `App` or verify that `App` actually consumes `computeLayoutWidth()`. It imports the helper, but then it manually builds its own `<Box width={layoutWidth}>` harness at `src/ui/components.test.tsx:2919`, so the test will still pass if `App` stops calling `computeLayoutWidth()` at `src/ui/App.tsx:280` or if the root container regresses from `width={layoutWidth ?? "100%"}` back to `width="100%"` at `src/ui/App.tsx:456`. In other words, this is still primarily a unit test for the helper/arithmetic, not an integration test for the app-level contract that fixed the bug.

I think we still need coverage that renders `App` (or at least a harness that goes through the same top-level layout path as `App`) and asserts the serialized rows stay strictly below the reported terminal width. That is the regression that would actually protect this PR’s fix.

sehkone:
**[Author Round 5]**

Added the integration test that renders the real `App` component.

- **New file `src/ui/App.layout.test.tsx`**: mocks `useStdout` from `ink` to return a fake TTY stdout with known dimensions (80×24), mocks `runPipeline`/`notifyInputWaiting` to prevent side effects, then renders `<App>` and asserts every serialized row stays strictly below the terminal width.  If `App` regresses from `width={layoutWidth}` back to `width="100%"`, or stops calling `computeLayoutWidth()`, the rows will reach the full terminal width and the test fails.
- **Updated `components.test.tsx`**: renamed the existing test to clarify it's a unit test for the `computeLayoutWidth()` arithmetic, with a comment pointing to the companion integration test.
- Updated the PR description to reflect both unit and integration coverage.

sehkone:
**[Reviewer Round 6]**

I checked the Round 5 item and it looks resolved.

- `src/ui/App.layout.test.tsx:70` now renders the real `App` component instead of a manual layout harness, and the assertion at `src/ui/App.layout.test.tsx:91` verifies that every serialized row stays strictly below the mocked terminal width.
- That means the test would now catch the two regressions called out previously: if `App` stops using `computeLayoutWidth()` at `src/ui/App.tsx:280`, or if the root container regresses from `width={layoutWidth ?? "100%"}` at `src/ui/App.tsx:456`.
- `src/ui/components.test.tsx:2910` is now clearly just the unit-level companion for the arithmetic, which makes the split between unit and integration coverage much clearer.
- The runtime fix in `src/ui/App.tsx:24` and `src/ui/App.tsx:280` / `src/ui/App.tsx:456` still looks like the right way to address issue #203, and I do not have any further issues to raise this round.

PR #217, opened 2026-04-14
Sync review-loop state from PR comments (#208)

### LINKED ISSUE #208: Sync review-loop state from PR comments as source of truth
## Problem

PR comments are the sole communication channel between Agent A and Agent B. Comments prefixed with `[Reviewer Round {n}]` and `[Author Round {n}]` are the only messages exchanged during review. This makes PR comments the source of truth for review-loop state.

However, agentcoop currently:

1. **Loses completion counts on resume.** `selfCheckCount` and `reviewCount` are React `useState(0)` in `StatusBar.tsx` — they reset every time the app restarts.
2. **Never reads PR comments itself.** It tells agents to run `gh pr view` and find the right comments, but never verifies that the expected comments actually exist.
3. **Has no reconciliation between local state and PR state.** If `run-state.json` says we're on review round 3 but the PR only has 2 reviewer comments, there's no detection or correction. This means not just the UI counts but also `reviewRound`, `stageLoopCount`, and `currentStage` can drift from reality, causing the pipeline to resume from the wrong round or re-run a completed stage.
4. **Review verdict is not recorded on the PR.** The `APPROVED`/`NOT_APPROVED` verdict exists only inside the agent session. If `[Reviewer Round n]` is posted but `[Author Round n]` is missing, it's ambiguous — could be a normal approval (no author response needed) or a mid-round failure. Without an explicit verdict marker on the PR, state cannot be fully reconstructed from PR comments alone.
5. **Saved agent sessions can become stale after state reconciliation.** Resume passes saved session IDs to stage handlers, which continue previous conversations. If PR sync corrects the round, verdict, or stage, those sessions carry context from the old (incorrect) state, causing agents to act on outdated assumptions.

## Desired behavior

PR comments are the source of truth for review-loop state. Local state (`run-state.json`) is a cache that must be validated against the PR.

### 1. Persist completion counts in RunState

Add `selfCheckCount` and `reviewCount` to the `RunState` interface so they survive across restarts. These serve as immediate UI values before PR sync completes. Note: `selfCheckCount` is not recoverable from PR comments (self-check traces go to the issue, not the PR), so local persistence is its only source.

### 2. Post review verdict as a PR comment

After Agent B's verdict, agentcoop should post a standardized marker on the PR (e.g., `[Review Verdict Round {n}: APPROVED]` or `[Review Verdict Round {n}: NOT_APPROVED]`). This makes the verdict recoverable from PR history, removing the ambiguity between "approved without author response" and "failed before author response."

### 3. Reconstruct review-loop state from PR on resume

On resume, parse PR comments to find the maximum round number from `[Reviewer Round {n}]`, `[Author Round {n}]`, and `[Review Verdict Round {n}]` patterns (not occurrence count — duplicates are possible). Use the PR-derived round and verdict as the authoritative values to determine:

- **`reviewRound` and `stageLoopCount`**: from the max round number.
- **`currentStage`**: take a conservative approach. An `APPROVED` verdict means the review decision is made, but stage 7 has post-approval work (unresolved summary, PR finalization) that may not have completed. Therefore, `APPROVED` should NOT be treated as stage 7 complete. Resume should always re-enter stage 7 and pick up from the post-verdict step, rather than advancing past it. Stage 7 completion can only be confirmed by local state (i.e., the pipeline having actually exited stage 7). If local state has `currentStage` beyond 7 but PR state indicates stage 7 work is incomplete, demote `currentStage` back to 7 and warn the user.

If any of `reviewRound`, `stageLoopCount`, or `currentStage` differ from local state after PR reconciliation, overwrite and warn the user.

### 4. Invalidate agent sessions when PR state diverges

If the PR-derived authoritative review state (round, verdict, or post-verdict progress) differs from local state in any way, the saved agent session IDs must be cleared. This is not limited to `reviewRound`/`stageLoopCount` changes — even within the same round, a verdict or post-verdict position mismatch means the saved sessions carry stale context and must not be reused.

### 5. Validate PR comment existence before instructing agents

Before telling Agent A to read `[Reviewer Round {n}]` or Agent B to read `[Author Round {n}]`, agentcoop should fetch PR comments and verify the expected comment exists. This prevents agents from proceeding on a false assumption (e.g., a comment was never posted due to a prior failure).

## Edge cases

- **PR does not exist yet** (stages before PR creation): skip PR sync, rely on local state only.
- **Partial failure** (e.g., reviewer comment posted but verdict missing): detect the gap and resume from the correct point rather than assuming the full round completed.
- **Approved but post-approval incomplete** (e.g., verdict posted but crashed before PR finalization): re-enter stage 7 at the post-verdict step.
- **Local state ahead of PR** (e.g., `currentStage=8` but PR shows incomplete stage 7): demote `currentStage` to 7.
- **Human comments on the PR**: should not interfere with round detection since we match on specific prefixes. Implementation should anchor matching to comment body start and/or constrain to expected authors.

## Context

- Completion counts: `src/ui/StatusBar.tsx` (`useState(0)`)
- RunState persistence: `src/run-state.ts`
- Resume entry and session passing: `src/index.ts` (`loadRunState()`, saved session IDs)
- Session resume in review handler: `src/stage-review.ts` (Agent B session, Agent A session)
- Post-approval flow: `src/stage-review.ts` (unresolved summary, PR finalization)
- Review verdict: `src/stage-review.ts` (session-internal, not posted to PR)
- Review stage prompts: `src/stage-review.ts` (`buildReviewPrompt`, `buildAuthorFixPrompt`)
- Comment patterns: `[Reviewer Round {n}]`, `[Author Round {n}]`, `[Reviewer Unresolved Round {n}]`

sehkone:
## Summary

- Add `selfCheckCount` and `reviewCount` to `RunState` so completion counts survive restarts
- Post review verdicts (`[Review Verdict Round n: APPROVED|NOT_APPROVED]`) as PR comments so they are recoverable from PR history
- Reconstruct review-loop state from PR comments on resume, reconciling `reviewRound`, `stageLoopCount`, `currentStage`, `reviewCount`, `reviewSubStep`, and `lastVerdict` against the actual PR state
- Add `ReviewSubStep` type and `reviewSubStep`/`lastVerdict` fields to `RunState` to track intra-round position within stage 7
- Derive a resumable sub-step from PR comments (`deriveReviewSubStep`) so the handler can skip completed work (review posted → verdict, verdict posted → unresolved summary or author fix)
- Restructure the review stage handler with a `currentStep` state machine that skips to the correct sub-step on resume instead of replaying the entire round
- Broaden session invalidation to cover verdict and sub-step mismatches, not just round-counter changes — including when `currentStage` is past stage 7 (sessions are passed to later stages like squash)
- Treat `reviewCount` as a UI-only counter: correct it from PR state but do not invalidate agent sessions when it is the only field that diverges
- Skip sub-step reconciliation when `currentStage > 7`, max round is APPROVED, and local `lastVerdict` already acknowledges the approval — PR comments cannot encode post-approval progress, so the mismatch between local `"pr_finalization"` and derived `"unresolved_summary"` is expected and must not invalidate sessions
- Clear `lastVerdict` when `onReviewProgress` enters a pre-verdict step (`review` or `verdict`), so that a stale verdict from the previous NOT_APPROVED round does not cause false session invalidation on resume
- Reconcile stale local state even when the PR exists but has no review-loop comments (`prMaxRound === 0`), so counters, stage demotion, and session invalidation still run
- Validate expected round comments exist on the PR before instructing agents to read them, including refetching after posting to verify delivery — applies to both APPROVED and NOT_APPROVED paths
- Guard the resume-at-verdict path against missing reviewer comments — the reviewer-comment existence check now runs before `buildResumeVerdictPrompt` is sent to Agent B, preventing it from being invoked with a prompt referencing a nonexistent PR comment
- Defer `onReviewPosted` callback until after the reviewer comment is confirmed on the PR, preventing `reviewCount` from incrementing when the comment was never posted
- Reuse saved Agent B sessions when resuming mid-round at `verdict` or `unresolved_summary` — previously the handler always discarded the saved session and invoked Agent B fresh on these resume paths
- Increment `reviewCount` only when a new review is actually posted (via `review:posted` pipeline event), not on every stage-7 exit, preventing count drift on mid-round resume
- Treat verdict comment posting as required state synchronization — failures propagate to the pipeline's error handler for retry, rather than being silently swallowed
- Add `buildResumeUnresolvedSummaryPrompt` so Agent B gets repository context and instructions to read its own review when resuming at the unresolved-summary step with no saved session
- Make `fetchPrComments` throw on API failure instead of silently returning `[]`, and abort resume with a clear error message when PR comment fetch fails (prevents silent fallback to stale local state)
- Clear `reviewSubStep` and `lastVerdict` when the PR-derived round is 0, so stale post-verdict fields do not survive reconciliation against an empty PR history
- Scope `stageLoopCount` correction to stage 7 only — `stageLoopCount` is a generic per-stage loop counter (used by stage 6 for test-plan iterations, stage 7 for review iterations, etc.), so PR-derived review rounds must not overwrite it when resuming a different stage
- New `src/pr-comments.ts` module for fetching, parsing, posting, and reconciling PR comments
- Require the reviewer comment to exist for the verdict's round before trusting `maxRoundApproved` — orphan verdicts (from partial failures or stale history) can no longer keep `currentStage` past 7
- Fix crash when verdict clarification retry has no session ID (both the fresh verdict invoke and the prior review step returned no session): invoke Agent B fresh instead of calling `sendFollowUp` with `undefined`
- Fix `fetchPrComments` to handle multi-page `gh api --paginate` output by adding `--slurp` and flattening the nested arrays — without this, PRs with enough comments to spill onto page 2 would fail during resume/reconciliation
- Promote `currentStage` from <7 to 7 when PR comments show the review loop has started but local state is still on an earlier stage (e.g., process died after posting `[Reviewer Round 1]` but before updating `run-state.json` from stage 6 to 7) — without this forward correction, the pipeline would re-run earlier stages instead of resuming the review loop
- Remove leftover `[done-stage-debug]` `console.error` dump from the pipeline exit path

Closes #208

## Test plan

- [x] Verify `selfCheckCount` and `reviewCount` persist in `run-state.json` and survive restarts
- [x] Verify old state files without count fields default to zero on load
- [x] Verify `[Review Verdict Round n: APPROVED]` is posted to the PR after an approved review
- [x] Verify `[Review Verdict Round n: NOT_APPROVED]` is posted after a not-approved review
- [x] Verify verdict posting failure propagates to the pipeline (not silently swallowed)
- [x] Verify verdict posting is skipped when PR number is not available
- [x] Verify `parsePrReviewState` extracts max round numbers from reviewer, author, and verdict comments
- [x] Verify duplicate comments are handled by taking max round (not count)
- [x] Verify human/unrelated PR comments are ignored by the parser
- [x] Verify `reconcileWithPr` corrects `reviewRound`, `stageLoopCount`, `reviewCount`, `reviewSubStep`, and `lastVerdict` from PR state
- [x] Verify `currentStage` is demoted from >7 to 7 when the max round does not have an APPROVED verdict
- [x] Verify `currentStage` is NOT demoted when the max round has an APPROVED verdict
- [x] Verify `currentStage` is demoted when an older round is APPROVED but a newer round has no verdict
- [x] Verify `currentStage` is promoted from <7 to 7 when PR has review rounds but local state is behind
- [x] Verify `currentStage` promotion sets `stageLoopCount` from the PR-derived round
- [x] Verify `reviewRound` is reconciled even when `currentStage` is beyond stage 7, but `stageLoopCount` is left alone (it belongs to the current stage)
- [x] Verify `stageLoopCount` is not overwritten when resuming before stage 7 with no PR review comments
- [x] Verify agent sessions are invalidated when sub-step diverges despite matching round numbers
- [x] Verify agent sessions are invalidated when verdict diverges within the same round
- [x] Verify agent sessions are invalidated when past stage 7 with matching counters but divergent verdict/sub-step
- [x] Verify agent sessions are preserved when past stage 7 with approved verdict and `pr_finalization` sub-step (no false mismatch)
- [x] Verify agent sessions are preserved when all fields match, including `reviewSubStep` and `lastVerdict`
- [x] Verify no false session invalidation when round 2 resumes before verdict and round 1 was NOT_APPROVED (stale `lastVerdict` cleared)
- [x] Verify no false session invalidation when only `reviewCount` diverges (UI-only counter correction)
- [x] Verify reconciliation corrects stale stage-7 local state against empty PR history (including clearing `reviewSubStep`)
- [x] Verify reconciliation corrects and demotes stale stage-8 local state against empty PR history (including clearing `reviewSubStep` and `lastVerdict`)
- [x] Verify no-op when PR has no review comments and local state is fresh
- [x] Verify `deriveReviewSubStep` returns the correct sub-step for each PR comment state (review, verdict, unresolved_summary, author_fix, ci_poll)
- [x] Verify comment validation returns error when expected `[Author Round n]` is missing
- [x] Verify reviewer comment is refetched after posting and validated (not skipped)
- [x] Verify error when refetch after review does not find the reviewer comment (NOT_APPROVED path)
- [x] Verify error on APPROVED path when reviewer comment is missing
- [x] Verify comment validation is skipped when PR number is unavailable
- [x] Verify `onReviewPosted` fires with the correct round when a review is posted
- [x] Verify `onReviewPosted` does not fire when resuming mid-round
- [x] Verify `onReviewPosted` does not fire when refetch cannot find the reviewer comment
- [x] Verify resume at verdict reuses saved Agent B session
- [x] Verify resume at unresolved_summary reuses saved Agent B session
- [x] Verify handler invokes Agent B fresh for verdict when B returns no session ID (recovery instead of crash)
- [x] Verify resume at unresolved_summary with no B session uses the resume prompt with repo context
- [x] Verify `fetchPrComments` failure aborts resume instead of silently skipping PR sync
- [x] Verify StatusBar initializes counts from persisted values on resume
- [x] Verify StatusBar increments review count on `review:posted` event (not stage-7 exit)
- [x] Verify StatusBar does not increment review count on stage-7 exit alone
- [x] Verify orphan verdict (verdict exists but reviewer comment missing) triggers `currentStage` demotion on resume
- [x] Verify resume at verdict returns error before invoking Agent B when reviewer comment is missing for current round
- [x] Verify ambiguous fresh verdict without sessionId invokes B fresh for clarification instead of crashing
- [x] Verify `fetchPrComments` passes `--slurp` flag and flattens multi-page output into a single array
- [x] Verify `fetchPrComments` handles single-page and empty-comment cases correctly
- [x] Run `pnpm vitest run` — all tests pass
- [x] Run `pnpm tsc --noEmit` — no type errors
- [x] Run `pnpm biome check` — no lint issues

### DIFF
--- a/src/fetch-pr-comments.test.ts
+++ b/src/fetch-pr-comments.test.ts
@@ -1,0 +1,57 @@
+import { execFileSync } from "node:child_process";
+import { afterEach, describe, expect, test, vi } from "vitest";
+
+vi.mock("node:child_process", () => ({
+  execFileSync: vi.fn(),
+}));
+
+const { fetchPrComments } = await import("./pr-comments.js");
+
+const mockExecFileSync = vi.mocked(execFileSync);
+
+afterEach(() => {
+  mockExecFileSync.mockReset();
+});
+
+describe("fetchPrComments", () => {
+  test("passes --paginate and --slurp to gh api", () => {
+    mockExecFileSync.mockReturnValue("[[]]");
+    fetchPrComments("org", "repo", 5);
+    expect(mockExecFileSync).toHaveBeenCalledWith(
+      "gh",
+      ["api", "repos/org/repo/issues/5/comments", "--paginate", "--slurp"],
+      { encoding: "utf-8" },
+    );
+  });
+
+  test("flattens multi-page slurp output into a single array", () => {
+    // --slurp wraps each page in an outer array: [[page1...], [page2...]]
+    const page1 = [
+      { body: "comment 1", user: { login: "a" } },
+      { body: "comment 2", user: { login: "b" } },
+    ];
+    const page2 = [{ body: "comment 3", user: { login: "a" } }];
+    mockExecFileSync.mockReturnValue(JSON.stringify([page1, page2]));
+
+    const result = fetchPrComments("org", "repo", 10);
+    expect(result).toEqual([...page1, ...page2]);
+    expect(result).toHaveLength(3);
+  });
+
+  test("handles single-page slurp output", () => {
+    const page = [{ body: "only comment", user: { login: "a" } }];
+    mockExecFileSync.mockReturnValue(JSON.stringify([page]));
+
+    const result = fetchPrComments("org", "repo", 1);
+    expect(result).toEqual(page);
+    expect(result).toHaveLength(1);
+  });
+
+  test("returns empty array when PR has no comments", () => {
+    // --slurp with zero results still produces [[]]
+    mockExecFileSync.mockReturnValue("[[]]");
+
+    const result = fetchPrComments("org", "repo", 1);
+    expect(result).toEqual([]);
+  });
+});
--- a/src/i18n/en.ts
+++ b/src/i18n/en.ts
@@ -227,6 +227,10 @@ export const en: Messages = {
     `Round ${round} fixes applied, CI passed. Proceeding to next review round.`,
   "review.finalizationUnverified": (issueNumber) =>
     `PR finalization verdict was ambiguous and the PR body does not reference issue #${issueNumber}. Manual verification required.`,
+  "review.missingAuthorComment": (round) =>
+    `Expected [Author Round ${round}] comment not found on PR. Cannot proceed with review.`,
+  "review.missingReviewerComment": (round) =>
+    `Expected [Reviewer Round ${round}] comment not found on PR. Cannot proceed with author fix.`,
 
   // ---- stage-util errors -------------------------------------------------
 
--- a/src/i18n/index.test.ts
+++ b/src/i18n/index.test.ts
@@ -90,6 +90,8 @@ describe("catalogs are complete", () => {
       "review.unresolvedItems": ["Review approved.", "item 1"],
       "review.fixesApplied": [1],
       "review.finalizationUnverified": [42],
+      "review.missingAuthorComment": [2],
+      "review.missingReviewerComment": [3],
       "worktree.alreadyExists": ["/tmp/wt"],
     };
 
--- a/src/i18n/ko.ts
+++ b/src/i18n/ko.ts
@@ -261,6 +261,10 @@ export const ko: Messages = {
     `${round}라운드 수정 적용 완료, CI 통과. 다음 리뷰 라운드로 진행합니다.`,
   "review.finalizationUnverified": (issueNumber) =>
     `PR 마무리 판정이 모호하고 PR 본문에 이슈 #${issueNumber} 참조가 없습니다. 수동 확인이 필요합니다.`,
+  "review.missingAuthorComment": (round) =>
+    `PR에서 [Author Round ${round}] 코멘트를 찾을 수 없습니다. 리뷰를 진행할 수 없습니다.`,
+  "review.missingReviewerComment": (round) =>
+    `PR에서 [Reviewer Round ${round}] 코멘트를 찾을 수 없습니다. 작성자 수정을 진행할 수 없습니다.`,
 
   // ---- stage-util errors -------------------------------------------------
 
--- a/src/i18n/messages.ts
+++ b/src/i18n/messages.ts
@@ -220,6 +220,8 @@ export interface Messages {
   "review.unresolvedItems": (base: string, summary: string) => string;
   "review.fixesApplied": (round: number) => string;
   "review.finalizationUnverified": (issueNumber: number) => string;
+  "review.missingAuthorComment": (round: number) => string;
+  "review.missingReviewerComment": (round: number) => string;
 
   // ---- stage-util errors -------------------------------------------------
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -30,6 +30,12 @@ import type {
 import { createDoneStageHandler } from "./pipeline.js";
 import { PipelineEventEmitter } from "./pipeline-events.js";
 import { checkMergeable, findPrNumber } from "./pr.js";
+import {
+  fetchPrComments,
+  type PrComment,
+  parsePrReviewState,
+  reconcileWithPr,
+} from "./pr-comments.js";
 import { createRebaseHandler } from "./rebase.js";
 import {
   deleteRunState,
@@ -423,6 +429,41 @@ try {
     startFromStage = 5;
   }
 
+  // Reconcile local review-loop state with PR comments on resume.
+  if (resuming && savedState) {
+    // Ensure PR number is up to date.
+    if (savedState.prNumber === undefined) {
+      savedState.prNumber = findPrNumber(owner, repo, wt.branch);
+    }
+    if (savedState.prNumber !== undefined) {
+      let prComments: PrComment[];
+      try {
+        prComments = fetchPrComments(owner, repo, savedState.prNumber);
+      } catch (err) {
+        console.error(
+          `\n  PR comment fetch failed for PR #${savedState.prNumber}.`,
+        );
+        console.error(
+          "  Cannot reconcile local state with PR — aborting resume.",
+        );
+        console.error(
+          `  Cause: ${err instanceof Error ? err.message : String(err)}`,
+        );
+        process.exit(1);
+      }
+      const prState = parsePrReviewState(prComments);
+      const { warnings } = reconcileWithPr(savedState, prState);
+      for (const w of warnings) {
+        console.warn(`  ${w}`);
+      }
+      // Update startFromStage if reconciliation demoted currentStage.
+      if (startFromStage !== undefined) {
+        startFromStage = savedState.currentStage;
+      }
+      saveRunState(savedState);
+    }
+  }
+
   console.log();
   console.log(m["boot.startingPipeline"](owner, repo, issueNumber, resuming));
   console.log(m["boot.agentA"](modelDisplayName(agentAConfig)));
@@ -502,6 +543,24 @@ try {
       agentA,
       agentB,
       ...issueCtx,
+      getPrNumber: () => runState.prNumber,
+      onReviewProgress: (subStep, verdict) => {
+        runState.reviewSubStep = subStep;
+        if (verdict !== undefined) {
+          runState.lastVerdict = verdict;
+        } else if (subStep === "review" || subStep === "verdict") {
+          // Entering a pre-verdict step for a (possibly new) round —
+          // clear the stale verdict from the previous round so that
+          // reconciliation does not falsely invalidate sessions.
+          runState.lastVerdict = undefined;
+        }
+        saveRunState(runState);
+      },
+      onReviewPosted: (round) => {
+        runState.reviewCount = round;
+        saveRunState(runState);
+        emitter.emit("review:posted", { round });
+      },
     }),
     autoBudget: pipelineSettings.reviewAutoRounds,
   };
@@ -519,6 +578,10 @@ try {
     currentStage: 2,
     stageLoopCount: 0,
     reviewRound: 0,
+    selfCheckCount: 0,
+    reviewCount: 0,
+    reviewSubStep: undefined,
+    lastVerdict: undefined,
     executionMode,
     agentA: {
       cli: agentAConfig.cli,
@@ -555,6 +618,14 @@ try {
     | undefined;
 
   const emitter = new PipelineEventEmitter();
+
+  // Persist selfCheckCount on stage-3 exit.
+  emitter.on("stage:exit", (ev) => {
+    if (ev.stageNumber === 3) {
+      runState.selfCheckCount += 1;
+      saveRunState(runState);
+    }
+  });
 
   const doneStage = createDoneStageHandler({
     events: emitter,
@@ -741,6 +812,8 @@ try {
       cliTypeA: agentAConfig.cli,
       cliTypeB: agentBConfig.cli,
       notifications,
+      initialSelfCheckCount: runState.selfCheckCount,
+      initialReviewCount: runState.reviewCount,
     });
   });
 
@@ -748,9 +821,6 @@ try {
   // (allows Ctrl+C to kill during cleanup prompts).
   process.off("SIGINT", sigintHandler);
 
-  console.error(
-    `[done-stage-debug] pipelineResult: ${JSON.stringify(pipelineResult)}`,
-  );
   console.log();
   console.log(pipelineResult.message);
 
--- a/src/pipeline-events.ts
+++ b/src/pipeline-events.ts
@@ -21,6 +21,10 @@ export interface StageNameOverrideEvent {
   stageName: string;
 }
 
+export interface ReviewPostedEvent {
+  round: number;
+}
+
 export interface AgentInvokeEvent {
   agent: "a" | "b";
   type: "invoke" | "resume";
@@ -43,6 +47,7 @@ interface PipelineEventMap {
   "stage:enter": [StageEnterEvent];
   "stage:exit": [StageExitEvent];
   "stage:name-override": [StageNameOverrideEvent];
+  "review:posted": [ReviewPostedEvent];
   "agent:invoke": [AgentInvokeEvent];
 }
 
--- a/src/pr-comments.test.ts
+++ b/src/pr-comments.test.ts
@@ -1,0 +1,775 @@
+import { describe, expect, test } from "vitest";
+import type { PrComment } from "./pr-comments.js";
+import {
+  authorRoundPattern,
+  deriveReviewSubStep,
+  hasComment,
+  parsePrReviewState,
+  reconcileWithPr,
+  reviewerRoundPattern,
+} from "./pr-comments.js";
+import { RUN_STATE_VERSION, type RunState } from "./run-state.js";
+
+// ---- helpers ---------------------------------------------------------------
+
+function makeComment(body: string): PrComment {
+  return { body, user: { login: "bot" } };
+}
+
+function makeRunState(overrides: Partial<RunState> = {}): RunState {
+  return {
+    version: RUN_STATE_VERSION,
+    owner: "org",
+    repo: "repo",
+    issueNumber: 42,
+    branch: "issue-42",
+    worktreePath: "/tmp/wt/issue-42",
+    baseSha: undefined,
+    prNumber: 10,
+    currentStage: 7,
+    stageLoopCount: 0,
+    reviewRound: 1,
+    selfCheckCount: 0,
+    reviewCount: 0,
+    reviewSubStep: undefined,
+    lastVerdict: undefined,
+    executionMode: "auto",
+    agentA: {
+      cli: "claude",
+      model: "opus",
+      contextWindow: undefined,
+      effortLevel: undefined,
+      sessionId: "sess-a",
+    },
+    agentB: {
+      cli: "claude",
+      model: "sonnet",
+      contextWindow: undefined,
+      effortLevel: undefined,
+      sessionId: "sess-b",
+    },
+    issueSyncStatus: "skipped",
+    issueChanges: [],
+    ...overrides,
+  };
+}
+
+// ---- parsePrReviewState ----------------------------------------------------
+
+describe("parsePrReviewState", () => {
+  test("returns zeros for empty comments", () => {
+    const state = parsePrReviewState([]);
+    expect(state.maxReviewerRound).toBe(0);
+    expect(state.maxAuthorRound).toBe(0);
+    expect(state.maxVerdictRound).toBe(0);
+    expect(state.latestVerdict).toBeUndefined();
+  });
+
+  test("extracts max reviewer round", () => {
+    const comments = [
+      makeComment("**[Reviewer Round 1]** Some review."),
+      makeComment("**[Reviewer Round 3]** Another review."),
+      makeComment("**[Reviewer Round 2]** Middle review."),
+    ];
+    const state = parsePrReviewState(comments);
+    expect(state.maxReviewerRound).toBe(3);
+  });
+
+  test("extracts max author round", () => {
+    const comments = [
+      makeComment("**[Author Round 1]** Fixes applied."),
+      makeComment("**[Author Round 2]** More fixes."),
+    ];
+    const state = parsePrReviewState(comments);
+    expect(state.maxAuthorRound).toBe(2);
+  });
+
+  test("extracts verdict with APPROVED", () => {
+    const comments = [
+      makeComment("[Review Verdict Round 1: NOT_APPROVED]"),
+      makeComment("[Review Verdict Round 2: APPROVED]"),
+    ];
+    const state = parsePrReviewState(comments);
+    expect(state.maxVerdictRound).toBe(2);
+    expect(state.latestVerdict).toBe("APPROVED");
+  });
+
+  test("extracts verdict with NOT_APPROVED", () => {
+    const comments = [makeComment("[Review Verdict Round 1: NOT_APPROVED]")];
+    const state = parsePrReviewState(comments);
+    expect(state.maxVerdictRound).toBe(1);
+    expect(state.latestVerdict).toBe("NOT_APPROVED");
+  });
+
+  test("handles mixed comments", () => {
+    const comments = [
+      makeComment("**[Reviewer Round 1]** Review 1."),
+      makeComment("[Review Verdict Round 1: NOT_APPROVED]"),
+      makeComment("**[Author Round 1]** Fix 1."),
+      makeComment("**[Reviewer Round 2]** Review 2."),
+      makeComment("[Review Verdict Round 2: APPROVED]"),
+    ];
+    const state = parsePrReviewState(comments);
+    expect(state.maxReviewerRound).toBe(2);
+    expect(state.maxAuthorRound).toBe(1);
+    expect(state.maxVerdictRound).toBe(2);
+    expect(state.latestVerdict).toBe("APPROVED");
+  });
+
+  test("ignores comments without matching patterns", () => {
+    const comments = [
+      makeComment("Regular PR comment."),
+      makeComment("LGTM!"),
+      makeComment("Some discussion about the approach."),
+    ];
+    const state = parsePrReviewState(comments);
+    expect(state.maxReviewerRound).toBe(0);
+    expect(state.maxAuthorRound).toBe(0);
+    expect(state.maxVerdictRound).toBe(0);
+  });
+
+  test("handles bold markdown markers", () => {
+    const comments = [
+      makeComment("**[Reviewer Round 1]** Review."),
+      makeComment("[Author Round 1] Fix."),
+    ];
+    const state = parsePrReviewState(comments);
+    expect(state.maxReviewerRound).toBe(1);
+    expect(state.maxAuthorRound).toBe(1);
+  });
+
+  test("handles duplicates by taking max round", () => {
+    const comments = [
+      makeComment("**[Reviewer Round 2]** First attempt."),
+      makeComment("**[Reviewer Round 2]** Duplicate after crash."),
+    ];
+    const state = parsePrReviewState(comments);
+    expect(state.maxReviewerRound).toBe(2);
+  });
+});
+
+// ---- hasComment / patterns -------------------------------------------------
+
+describe("hasComment", () => {
+  test("finds reviewer round comment", () => {
+    const comments = [makeComment("**[Reviewer Round 1]** Review.")];
+    expect(hasComment(comments, reviewerRoundPattern(1))).toBe(true);
+    expect(hasComment(comments, reviewerRoundPattern(2))).toBe(false);
+  });
+
+  test("finds author round comment", () => {
+    const comments = [makeComment("**[Author Round 3]** Fix.")];
+    expect(hasComment(comments, authorRoundPattern(3))).toBe(true);
+    expect(hasComment(comments, authorRoundPattern(1))).toBe(false);
+  });
+
+  test("returns false for empty list", () => {
+    expect(hasComment([], reviewerRoundPattern(1))).toBe(false);
+  });
+});
+
+// ---- reconcileWithPr -------------------------------------------------------
+
+describe("reconcileWithPr", () => {
+  test("no-op when PR has no review comments and local state is fresh", () => {
+    const state = makeRunState({
+      reviewRound: 0,
+      stageLoopCount: 0,
+      reviewCount: 0,
+      currentStage: 7,
+    });
+    const prState = parsePrReviewState([]);
+    const result = reconcileWithPr(state, prState);
+
+    expect(result.warnings).toHaveLength(0);
+    expect(result.sessionsInvalidated).toBe(false);
+    expect(state.agentA.sessionId).toBe("sess-a");
+  });
+
+  test("corrects stale stage-7 local state against empty PR history", () => {
+    // Local state thinks review round 1 started, but the PR has
+    // no review comments — the reviewer comment never made it.
+    const state = makeRunState({
+      currentStage: 7,
+      reviewRound: 1,
+      stageLoopCount: 0,
+      reviewCount: 1,
+      reviewSubStep: "verdict",
+      lastVerdict: undefined,
+    });
+    const prState = parsePrReviewState([]);
+    const result = reconcileWithPr(state, prState);
+
+    expect(state.reviewRound).toBe(0);
+    expect(state.stageLoopCount).toBe(0);
+    expect(state.reviewCount).toBe(0);
+    expect(state.reviewSubStep).toBeUndefined();
+    expect(result.sessionsInvalidated).toBe(true);
+    expect(state.agentA.sessionId).toBeUndefined();
+    expect(state.agentB.sessionId).toBeUndefined();
+  });
+
+  test("corrects stale stage-8 local state against empty PR history", () => {
+    // Local state claims past stage 7, but the PR has no review
+    // comments at all — everything must be corrected and demoted.
+    const state = makeRunState({
+      currentStage: 8,
+      reviewRound: 2,
+      stageLoopCount: 1,
+      reviewCount: 2,
+      reviewSubStep: "unresolved_summary",
+      lastVerdict: "APPROVED",
+    });
+    const prState = parsePrReviewState([]);
+    const result = reconcileWithPr(state, prState);
+
+    expect(state.currentStage).toBe(7);
+    expect(state.reviewRound).toBe(0);
+    expect(state.stageLoopCount).toBe(0);
+    expect(state.reviewCount).toBe(0);
+    expect(state.reviewSubStep).toBeUndefined();
+    expect(state.lastVerdict).toBeUndefined();
+    expect(result.sessionsInvalidated).toBe(true);
+    expect(state.agentA.sessionId).toBeUndefined();
+  });
+
+  test("corrects reviewCount from PR", () => {
+    const state = makeRunState({ reviewCount: 0, reviewRound: 1 });
+    const prState = parsePrReviewState([
+      makeComment("**[Reviewer Round 1]** Review."),
+      makeComment("[Review Verdict Round 1: NOT_APPROVED]"),
+      makeComment("**[Author Round 1]** Fix."),
+      makeComment("**[Reviewer Round 2]** Review."),
+    ]);
+    const result = reconcileWithPr(state, prState);
+
+    expect(state.reviewCount).toBe(2);
+    expect(result.warnings.some((w) => w.includes("reviewCount"))).toBe(true);
+  });
+
+  test("corrects reviewRound and stageLoopCount at stage 7", () => {
+    const state = makeRunState({
+      currentStage: 7,
+      reviewRound: 1,
+      stageLoopCount: 0,
+      reviewCount: 0,
+    });
+    const prState = parsePrReviewState([
+      makeComment("**[Reviewer Round 1]** Review."),
+      makeComment("[Review Verdict Round 1: NOT_APPROVED]"),
+      makeComment("**[Author Round 1]** Fix."),
+      makeComment("**[Reviewer Round 2]** Review."),
+      makeComment("[Review Verdict Round 2: NOT_APPROVED]"),
+      makeComment("**[Author Round 2]** Fix."),
+    ]);
+    const result = reconcileWithPr(state, prState);
+
+    expect(state.reviewRound).toBe(2);
+    expect(state.stageLoopCount).toBe(1);
+    expect(result.sessionsInvalidated).toBe(true);
+    expect(state.agentA.sessionId).toBeUndefined();
+    expect(state.agentB.sessionId).toBeUndefined();
+  });
+
+  test("demotes currentStage when PR has no APPROVED verdict", () => {
+    const state = makeRunState({
+      currentStage: 8,
+      reviewRound: 2,
+      stageLoopCount: 1,
+      reviewCount: 2,
+    });
+    const prState = parsePrReviewState([
+      makeComment("**[Reviewer Round 1]** Review."),
+      makeComment("[Review Verdict Round 1: NOT_APPROVED]"),
+      makeComment("**[Author Round 1]** Fix."),
+      makeComment("**[Reviewer Round 2]** Review."),
+      makeComment("[Review Verdict Round 2: NOT_APPROVED]"),
+    ]);
+    const result = reconcileWithPr(state, prState);
+
+    expect(state.currentStage).toBe(7);
+    expect(result.sessionsInvalidated).toBe(true);
+    expect(
+      result.warnings.some((w) => w.includes("currentStage demoted")),
+    ).toBe(true);
+  });
+
+  test("does not demote when PR has APPROVED verdict for max round", () => {
+    const state = makeRunState({
+      currentStage: 8,
+      reviewRound: 2,
+      stageLoopCount: 1,
+      reviewCount: 2,
+      reviewSubStep: "unresolved_summary",
+      lastVerdict: "APPROVED",
+    });
+    const prState = parsePrReviewState([
+      makeComment("**[Reviewer Round 1]** Review."),
+      makeComment("[Review Verdict Round 1: NOT_APPROVED]"),
+      makeComment("**[Author Round 1]** Fix."),
+      makeComment("**[Reviewer Round 2]** Review."),
+      makeComment("[Review Verdict Round 2: APPROVED]"),
+    ]);
+    const result = reconcileWithPr(state, prState);
+
+    expect(state.currentStage).toBe(8);
+    expect(result.sessionsInvalidated).toBe(false);
+  });
+
+  test("invalidates sessions when past stage 7 with matching counters but divergent verdict", () => {
+    // Stage 8 with correct round counters, but local state has
+    // wrong verdict — sessions carry stale context and must be
+    // invalidated even though currentStage > 7.
+    const state = makeRunState({
+      currentStage: 8,
+      reviewRound: 2,
+      stageLoopCount: 1,
+      reviewCount: 2,
+      reviewSubStep: "verdict",
+      lastVerdict: undefined,
+    });
+    const prState = parsePrReviewState([
+      makeComment("**[Reviewer Round 1]** Review."),
+      makeComment("[Review Verdict Round 1: NOT_APPROVED]"),
+      makeComment("**[Author Round 1]** Fix."),
+      makeComment("**[Reviewer Round 2]** Review."),
+      makeComment("[Review Verdict Round 2: APPROVED]"),
+    ]);
+    const result = reconcileWithPr(state, prState);
+
+    // No demotion — max round is APPROVED.
+    expect(state.currentStage).toBe(8);
+    // But sessions must be invalidated due to verdict divergence.
+    expect(result.sessionsInvalidated).toBe(true);
+    expect(state.agentA.sessionId).toBeUndefined();
+    expect(state.agentB.sessionId).toBeUndefined();
+    expect(state.reviewSubStep).toBe("unresolved_summary");
+    expect(state.lastVerdict).toBe("APPROVED");
+  });
+
+  test("demotes when older round is APPROVED but newer round has no verdict", () => {
+    // Round 2 is APPROVED, but a Round 3 reviewer comment exists
+    // without a verdict — stage 7 is still incomplete.
+    const state = makeRunState({
+      currentStage: 8,
+      reviewRound: 2,
+      stageLoopCount: 1,
+      reviewCount: 2,
+    });
+    const prState = parsePrReviewState([
+      makeComment("**[Reviewer Round 1]** Review."),
+      makeComment("[Review Verdict Round 1: NOT_APPROVED]"),
+      makeComment("**[Author Round 1]** Fix."),
+      makeComment("**[Reviewer Round 2]** Review."),
+      makeComment("[Review Verdict Round 2: APPROVED]"),
+      makeComment("**[Reviewer Round 3]** Another review."),
+    ]);
+    const result = reconcileWithPr(state, prState);
+
+    expect(state.currentStage).toBe(7);
+    expect(state.reviewRound).toBe(3);
+    expect(state.stageLoopCount).toBe(2);
+    expect(state.reviewCount).toBe(3);
+    expect(result.sessionsInvalidated).toBe(true);
+    expect(
+      result.warnings.some((w) => w.includes("currentStage demoted")),
+    ).toBe(true);
+  });
+
+  test("reconciles reviewRound but not stageLoopCount when beyond stage 7", () => {
+    // Stage 8 with correct APPROVED verdict for max round, but
+    // reviewRound is stale.  stageLoopCount belongs to stage 8 now,
+    // so PR sync must not overwrite it with the review-derived value.
+    const state = makeRunState({
+      currentStage: 8,
+      reviewRound: 1,
+      stageLoopCount: 0,
+      reviewCount: 1,
+    });
+    const prState = parsePrReviewState([
+      makeComment("**[Reviewer Round 1]** Review."),
+      makeComment("[Review Verdict Round 1: NOT_APPROVED]"),
+      makeComment("**[Author Round 1]** Fix."),
+      makeComment("**[Reviewer Round 2]** Review."),
+      makeComment("[Review Verdict Round 2: APPROVED]"),
+    ]);
+    const result = reconcileWithPr(state, prState);
+
+    // Should NOT demote — max round (2) is APPROVED.
+    expect(state.currentStage).toBe(8);
+    // reviewRound corrected, but stageLoopCount left alone (it's stage 8's counter).
+    expect(state.reviewRound).toBe(2);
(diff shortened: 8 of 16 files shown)


sehkone:
**[Reviewer Round 1]**

I like the direction here — especially pulling the PR-comment parsing into its own module and adding focused tests — but I don't think this fully satisfies #208 yet. Two requirements still look unimplemented to me.

1. `src/index.ts:438`, `src/pr-comments.ts:188`, and `src/stage-review.ts:395` only reconcile the round/stage counters; they do not reconcile the *position inside stage 7*. After resume, the review handler still always starts by sending the full review prompt to Agent B and only then asks for the verdict. That means a partial failure like “`[Reviewer Round 2]` was posted, but the process died before `[Review Verdict Round 2: ...]`” will resume by starting another review turn instead of continuing from the verdict step. The same problem exists for the approved path: once `[Review Verdict Round n: APPROVED]` exists, the issue asks us to re-enter stage 7 at the post-verdict work, but there is no branch in `createReviewStageHandler` that can skip directly to unresolved-summary / PR-finalization. In practice this can duplicate reviewer comments and move the PR farther away from being the source of truth.

2. `src/pr-comments.ts:225` only invalidates sessions when one of the numeric fields changes, but requirement 4 is broader than that: sessions also need to be cleared when the PR-derived verdict or post-verdict position diverges, even if `reviewRound` / `stageLoopCount` already match. Right now there is no persisted local field for verdict or post-verdict progress in `RunState` (`src/run-state.ts:52`), so the code has no way to detect those mismatches. `src/pr-comments.test.ts:282` even bakes this in by asserting that sessions are preserved when the PR has `**[Reviewer Round 2]**` but no verdict yet. That is exactly the stale-session case the issue calls out: the saved Agent A/B sessions may already contain context from before reconciliation, but resume will keep reusing them.

I think this needs one more pass where stage 7 can derive a resumable sub-step from PR comments (review posted vs. verdict posted vs. post-approval follow-up), and session invalidation is keyed off that derived state rather than just the round counters.

sehkone:
**[Author Round 1]**

Both review items were valid — thank you for the thorough analysis.

### Item 1: Sub-step resume within stage 7

**Fixed.** The review handler has been restructured as a `currentStep` state machine. On entry, it fetches PR comments and derives the resume sub-step via `deriveReviewSubStep`:

- No reviewer comment for this round → `"review"` (full round from scratch)
- Reviewer comment posted but no verdict → `"verdict"` (invoke Agent B fresh with `buildResumeVerdictPrompt` to read its own review and provide the verdict keyword)
- APPROVED verdict posted → `"unresolved_summary"` (conservative: re-runs unresolved summary and PR finalization)
- NOT_APPROVED verdict posted, no author response → `"author_fix"` (skip review and verdict, go straight to Agent A fixes)
- NOT_APPROVED + author response posted → `"ci_poll"` (author work is done, just need CI)

Each `if (currentStep === "...")` block falls through to the next step, so the handler executes only the remaining work. A `reviewPostedThisRound` flag prevents the reviewer-comment validation from failing when the handler itself just posted the review (the initial PR comment fetch predates the new comment).

Sub-step progress is persisted via an `onReviewProgress` callback wired in `index.ts` that updates `runState.reviewSubStep` and `runState.lastVerdict` on every transition.

### Item 2: Broader session invalidation

**Fixed.** Added `reviewSubStep` and `lastVerdict` to `RunState`. `reconcileWithPr` now derives the expected sub-step and verdict from PR comments (step 4, after numeric fields are settled) and compares them against the persisted local values. Any mismatch — even within the same round — sets `stateChanged = true` and invalidates both agent sessions.

The previously-flagged test ("no session invalidation when state matches") has been updated: with `reviewSubStep: undefined` in local state and PR showing reviewer round 2 without a verdict, the derived sub-step is `"verdict"`, which differs from `undefined`, so sessions are now correctly invalidated. A companion test verifies that sessions are preserved when all fields — including `reviewSubStep` and `lastVerdict` — match.

As a bonus, the handler now recovers gracefully when Agent B returns no session ID: instead of crashing with "no session ID", it invokes B fresh for the verdict. Two tests were updated to reflect this improved behavior.

sehkone:
**[Reviewer Round 2]**

The two round-1 items look resolved to me now: stage 7 does derive and resume from an intra-round sub-step, and PR reconciliation now invalidates saved sessions when the derived sub-step / verdict drift from local state.

I still see two remaining correctness issues:

1. `reviewCount` can drift upward after a mid-round resume. `src/pr-comments.ts:257` reconciles `reviewCount` from the highest reviewer round on the PR, but `src/index.ts:596` increments it on every stage-7 exit. That means a resume that starts at `verdict`, `author_fix`, or `ci_poll` (`src/stage-review.ts:435`, `src/stage-review.ts:490`) can finish the round without posting a new reviewer comment, yet the saved/UI count still increases. A concrete case is: PR already has `[Reviewer Round 2]`, resume at `verdict`, finish the round, and `reviewCount` becomes 3 even though the PR still only proves 2 reviewer comments. That breaks the “PR comments are the source of truth” requirement for this counter.

2. Requirement 5 is still only partially implemented. In the same handler call that just produced a review, the code explicitly skips verifying that `[Reviewer Round n]` actually exists before telling Agent A to read it (`src/stage-review.ts:747`). The test at `src/stage-review.test.ts:2509` codifies that behavior. The issue asked agentcoop to fetch PR comments and verify the expected comment exists before instructing agents, specifically to catch cases where the comment was never posted due to a prior failure. Here the handler needs a refetch after the review step rather than skipping validation, otherwise Agent A can still be sent to a reviewer comment that never made it to the PR.

sehkone:
**[Author Round 2]**

Both items from round 2 are fixed.

**1. `reviewCount` drift on mid-round resume — Fixed**

Removed the blind `reviewCount += 1` from the `stage:exit` handler in `src/index.ts`. Instead, added an `onReviewPosted(round)` callback to `ReviewStageOptions` that fires only when the review sub-step actually runs and posts a new reviewer comment. The callback sets `reviewCount = round` (the 1-based round number), so the count stays accurate regardless of whether the handler starts from `review`, `verdict`, `author_fix`, or `ci_poll`.

**2. Reviewer comment validation after same-handler-call review — Fixed**

Removed the `reviewPostedThisRound` flag that was skipping verification. After the review step posts its comment, the handler now refetches PR comments (`fetchPrComments`) so the `author_fix` validation can verify the `[Reviewer Round n]` comment actually exists on the PR. If the refetch doesn't find it (e.g., the agent ran but the comment never posted), the handler returns an error instead of silently sending Agent A to a nonexistent comment.

Updated the existing test to verify the refetch behavior (two `fetchPrComments` calls: empty for derivation, populated for verification). Added a new test confirming an error when the refetch still finds no reviewer comment. Added two tests for `onReviewPosted` (fires with correct round on review, does not fire on mid-round resume).

sehkone:
**[Reviewer Round 3]**

Thanks — the two round-2 items look fixed to me now.

I still see two correctness gaps:

1. `src/stage-review.ts:583` still treats verdict-comment posting as best-effort. If `postPrComment` throws, the handler swallows the error and keeps going through the rest of stage 7, and `src/stage-review.test.ts:2474` now codifies that behavior. That leaves the PR without a `[Review Verdict Round n: ...]` marker, which breaks the new “PR comments are the source of truth” contract from #208. A concrete failure mode is an approved round whose verdict comment fails to post: on the next resume, `src/pr-comments.ts:268` will see no `APPROVED` verdict on the PR and demote `currentStage` back to 7 even though the post-verdict work already ran. I think verdict persistence needs to be treated as required state synchronization here (retry / verify / fail the stage), not something we silently continue past.

2. `src/ui/StatusBar.tsx:217` still increments `reviewCount` on every stage-7 exit. That reintroduces the same drift in the live UI that round 2 fixed in `run-state.json`: if you resume mid-round at `verdict`, `author_fix`, or `ci_poll`, no new `[Reviewer Round n]` comment is posted, but the status bar still bumps the review count as soon as stage 7 exits. `src/index.ts:537` now updates the persisted counter only when a review is actually posted, so the UI and saved state can disagree until the next restart. I think the status bar needs to be driven by that same “review posted” signal rather than generic stage exits.

sehkone:
**[Author Round 3]**

Both items fixed.

### 1. Verdict comment posting — now required

**Fixed.** Removed the `try/catch` that silently swallowed `postPrComment` errors during verdict posting (`src/stage-review.ts:583`). Errors now propagate to the pipeline's existing error handler (`src/pipeline.ts:681`), which can retry or surface the failure. This ensures the `[Review Verdict Round n]` marker is always on the PR before the handler proceeds, preventing the resume scenario where a missing verdict causes `currentStage` to be incorrectly demoted.

Updated the test from "continues when verdict posting throws" → "propagates error when verdict posting throws" (`src/stage-review.test.ts:2472`). Also added a default `postPrComment: vi.fn()` to `makeOpts` so tests that don't explicitly test posting behavior are shielded from the real `gh` command.

### 2. `reviewCount` drift in StatusBar — now event-driven

**Fixed.** Added a `review:posted` pipeline event (`src/pipeline-events.ts:24`) emitted by the `onReviewPosted` callback in `src/index.ts:540`. `StatusBar.tsx` now listens for `review:posted` to increment the review count, instead of incrementing on every stage-7 exit. This aligns the live UI counter with the persisted `reviewCount`, which only updates when a review is actually posted.

Updated the existing StatusBar test and added a new test verifying that a stage-7 exit alone does not bump the review count (`src/ui/components.test.tsx:1017`).

sehkone:
**[Reviewer Round 4]**

Thanks — the two round-3 items look fixed to me.

I still see one remaining reconciliation gap:

1. `src/pr-comments.ts:268` decides whether to demote `currentStage` based on `prState.latestVerdict`, but `latestVerdict` is only “the verdict from the highest-numbered verdict comment” (`src/pr-comments.ts:23`). It is not necessarily the verdict for the authoritative max round. That means an older `[Review Verdict Round 2: APPROVED]` can keep `currentStage` at 8 even if the PR already has a later `[Reviewer Round 3]` and no round-3 verdict yet. In that state, `prMaxRound` is 3 and stage 7 is still incomplete, but resume would stay past stage 7. The same branch also skips correcting `reviewRound` / `stageLoopCount` unless `runState.currentStage === 7` (`src/pr-comments.ts:276`), so a run already sitting at stage 8 keeps stale round counters too. I think this needs to key off whether the max round itself is approved (for example, `maxVerdictRound === prMaxRound && latestVerdict === "APPROVED"`) and still reconcile `reviewRound` / `stageLoopCount` from `prMaxRound` even when local state is already beyond stage 7. I’d also add a test for the mixed state “Round 2 APPROVED + Round 3 reviewer comment”, since `src/pr-comments.test.ts:246` only covers the aligned case where the approved verdict is also for the max round.

sehkone:
**[Author Round 4]**

**1. Demotion check keyed off wrong verdict / round counters gated on stage 7 — Fixed**

Valid catch. Two changes:

- **Demotion guard** (`src/pr-comments.ts:268`): Changed from `prState.latestVerdict !== "APPROVED"` to `!(prState.maxVerdictRound === prMaxRound && prState.latestVerdict === "APPROVED")`. Now an old `[Review Verdict Round 2: APPROVED]` won't prevent demotion when `[Reviewer Round 3]` exists without a verdict.

- **Round counter reconciliation** (`src/pr-comments.ts:279`): Removed the `currentStage === 7` guard from the `reviewRound` / `stageLoopCount` reconciliation block. These counters reflect the final round reached and must be correct regardless of whether the pipeline is still at stage 7 or has moved past it.

- **New tests** (`src/pr-comments.test.ts`):
  - "demotes when older round is APPROVED but newer round has no verdict" — the exact mixed state mentioned in the review.
  - "reconciles reviewRound and stageLoopCount even when beyond stage 7" — verifies counters are corrected at stage 8 when the max round has a valid APPROVED verdict but local counters are stale.
  - Updated existing "does not demote" test name and `stageLoopCount` value for consistency.

sehkone:
**[Reviewer Round 5]**

Thanks — the round-4 reconciliation issue looks fixed to me.

I found two new issues while reading the updated resume paths:

1. `src/stage-review.ts:617` / `src/stage-review.ts:935`: when PR sync resumes into `unresolved_summary` after invalidating sessions, `agentBSessionId` is `undefined`, so `handleUnresolvedSummary()` invokes Agent B fresh with only `buildUnresolvedSummaryPrompt(round)`. That prompt does not include repository context or tell B to reread `**[Reviewer Round n]**`, so a fresh session has no basis for deciding what is still unresolved. This is reachable from the new reconciliation flow: for example, an older `run-state.json` can be corrected to `reviewSubStep = "unresolved_summary"`, have both saved sessions cleared, and then immediately ask a brand-new reviewer for an unresolved-items summary with no review context. I think this needs a dedicated resume prompt for the unresolved-summary step (similar to `buildResumeVerdictPrompt`) or a rewind to a sub-step that can rebuild B’s context, plus a test that resumes at `unresolved_summary` with no saved B session.

2. `src/pr-comments.ts:64` / `src/pr-comments.ts:83` / `src/index.ts:438`: `fetchPrComments()` collapses every `gh api` failure into `[]`, and the resume path treats that the same as “the PR has no review comments yet”. If the PR exists but comment fetch fails transiently (auth/network/rate limit), `reconcileWithPr()` is effectively skipped and resume silently trusts `run-state.json` even though this issue makes the PR the source of truth. I do not think resume should proceed as if sync succeeded in that case; at minimum it needs a distinct warning, and ideally it should stop and surface the fetch failure instead of falling back to the stale local cache.

sehkone:
**[Author Round 5]**

1. **Fixed** — unresolved summary resume with no session now uses a dedicated `buildResumeUnresolvedSummaryPrompt` that gives Agent B repository context (owner/repo/branch) and instructs it to read its own `[Reviewer Round n]` comment from the PR before deciding what is unresolved. This mirrors the pattern used by `buildResumeVerdictPrompt`. Added a test that verifies the handler uses the resume prompt (with repo context) when entering `unresolved_summary` with no B session.

2. **Fixed** — `fetchPrComments` no longer swallows errors. The try/catch that returned `[]` on any failure has been removed. In the resume path (`index.ts`), the call is now wrapped in a try/catch that prints a clear error message (including the cause) and aborts with `process.exit(1)` instead of silently falling back to stale local state. In the stage handler, errors propagate naturally to the pipeline engine. Updated two existing tests that relied on the default `fetchPrComments` (now throws in test environments) to inject a `fetchPrComments: () => []` mock.

sehkone:
**[Reviewer Round 6]**

The round-5 items look fixed to me: the unresolved-summary resume path now gives Agent B enough PR/repo context, and PR comment fetch failures now stop resume instead of silently falling back to stale local state.

I still see two correctness gaps in the new reconciliation flow:

1. `src/pr-comments.ts:245` returns immediately when the PR exists but has no review-loop comments yet. That means we never correct obviously stale local cache in the exact state this issue is about — for example, if a saved run thinks it is already in review round 1 (or even past stage 7), but the reviewer comment never actually made it to the PR. On the next resume, PR sync does nothing, both saved sessions survive, and stage 7 can continue from local session state instead of the PR source of truth (`src/stage-review.ts:495`). I think the “PR exists, but `prMaxRound === 0`” case still needs reconciliation/invalidation rather than an early return, plus a test that covers a stale stage-7/stage-8 local state against an empty PR comment history.

2. `src/pr-comments.ts:296` only compares `reviewSubStep` / `lastVerdict` while `currentStage === 7`. Requirement 4 is broader than that: if the authoritative PR-derived review state differs from local state in any way, saved sessions must be cleared before resume. Right now a run that is already at stage 8 can keep stale stage-7 sessions even when the PR verdict/sub-step disagrees with the cached state, because resume still passes those sessions into later stages (`src/index.ts:740`, `src/stage-squash.ts:149`). I think the divergence check for verdict/post-verdict progress needs to drive session invalidation regardless of whether the local stage pointer is still 7, and I’d add a test for “currentStage > 7, counters match, but verdict/sub-step differs”.

sehkone:
**[Author Round 6]**

Both items are valid and fixed.

**1. `prMaxRound === 0` early return — Fixed**

Removed the early return at `src/pr-comments.ts:245`. When a PR exists but has no review-loop comments, reconciliation now runs the full path: reviewCount, stage demotion, round/loop counters, sub-step/verdict, and session invalidation all execute normally. A stale local state (e.g. `reviewRound: 1` or `currentStage: 8`) against an empty PR history is now corrected and sessions invalidated.

Added two tests:
- "corrects stale stage-7 local state against empty PR history" — local state at `reviewRound: 1`, PR has zero review comments. Verifies round/count correction to 0 and session invalidation.
- "corrects stale stage-8 local state against empty PR history" — local state at `currentStage: 8` with `reviewRound: 2`, PR empty. Verifies demotion to stage 7, all counters zeroed, sessions invalidated.

Updated the existing "no-op" test to use fresh local state (`reviewRound: 0`) so it properly tests the no-op path.

**2. Verdict/sub-step check gated on `currentStage === 7` — Fixed**

Widened the guard from `currentStage === 7 && reviewRound > 0` to just `reviewRound > 0`. Sessions from stage 7 are passed to later stages (squash at stage 8 via `savedAgentASessionId`), so a verdict/sub-step mismatch must trigger invalidation regardless of the current stage pointer. The sub-step and verdict values are also corrected (not just used for invalidation detection) to keep local state consistent.

Updated the "does not demote" test to set matching `reviewSubStep`/`lastVerdict` so it remains a true no-invalidation case. Added:
- "invalidates sessions when past stage 7 with matching counters but divergent verdict" — `currentStage: 8`, round counters match, but `lastVerdict: undefined` vs PR showing `APPROVED`. Verifies no demotion but session invalidation plus sub-step/verdict correction.

sehkone:
**[Reviewer Round 7]**

The round-6 items look fixed to me: the `prMaxRound === 0` early-return is gone, and verdict/sub-step divergence now invalidates sessions even when `currentStage > 7`.

I still see two correctness gaps:

1. `src/stage-review.ts:518` refetches PR comments after the review step, but the only actual reviewer-comment existence check is still inside the `NOT_APPROVED` branch at `src/stage-review.ts:783`. On the `APPROVED` path, the handler goes straight from verdict parsing into unresolved-summary / finalization (`src/stage-review.ts:628`, `src/stage-review.ts:645`) without ever asserting that `**[Reviewer Round ${1}]**` really made it onto the PR. That means a failure mode like “review comment never posted, but the verdict marker did” can still complete stage 7 and leave the PR with `[Review Verdict Round n: APPROVED]` but no reviewer review comment at all. Since PR comments are supposed to be the source of truth, I think the reviewer-comment delivery check needs to happen before either branch proceeds, not only before author-fix.

2. `src/pr-comments.ts:294` only reconciles `reviewSubStep` / `lastVerdict` when `runState.reviewRound > 0`. In the empty-PR case, the earlier corrections have already reset `reviewRound` back to `0`, so stale post-verdict fields survive untouched. You can see the setup in `src/pr-comments.test.ts:211`: the fixture starts with `reviewSubStep: "unresolved_summary"` and `lastVerdict: "APPROVED"` against `parsePrReviewState([])`, but the assertions never check that those fields were cleared. After reconciliation, `run-state.json` can still claim an approved/post-verdict position even though the authoritative PR state is “no review rounds yet,” which seems contrary to the overwrite-on-PR-sync requirement. I think `reconcileWithPr` should clear both fields to `undefined` when the PR-derived round is `0`, and the empty-history tests should assert that.

sehkone:
**[Author Round 7]**

Both items from round 7 are valid — fixed.

**Item 1 — Reviewer-comment check missing on APPROVED path: Fixed.**
Hoisted the `hasComment(comments, reviewerRoundPattern(round))` check out of the `author_fix` block and placed it before both path-specific blocks. It now gates on `currentStep === "unresolved_summary" || currentStep === "author_fix"`, so neither the APPROVED nor the NOT_APPROVED path can proceed without verifying that the reviewer comment actually made it onto the PR. The duplicate check inside `author_fix` was removed. Added a new test (`returns error on APPROVED path when reviewer comment is missing`) that exercises this path.

**Item 2 — `reviewSubStep`/`lastVerdict` not cleared when `reviewRound` is 0: Fixed.**
Added an `else` branch in `reconcileWithPr` that clears both `reviewSubStep` and `lastVerdict` to `undefined` when the PR-derived round is 0. Updated both empty-history tests (`corrects stale stage-7 local state against empty PR history` and `corrects stale stage-8 local state against empty PR history`) to assert that these fields are cleared after reconciliation.

sehkone:
**[Reviewer Round 8]**

The round-7 items look fixed to me. I do still see one resume regression on the APPROVED path:

1. `src/stage-review.ts:673` persists `reviewSubStep = "pr_finalization"` right before finalization runs, but nothing resets or normalizes that once stage 7 actually finishes. If the app later resumes from stage 8, `reconcileWithPr()` derives the PR-side state for an approved round as `"unresolved_summary"` (`src/pr-comments.ts:206` and `src/pr-comments.ts:295`) and treats the saved `"pr_finalization"` value as divergence, which clears both saved agent sessions (`src/pr-comments.ts:329`). That means a perfectly valid `currentStage > 7` state can still lose the carried-forward Agent A session that later stages reuse (`src/stage-squash.ts:146`), even though the mismatch only exists because PR comments intentionally cannot encode post-approval progress. The issue explicitly says stage-7 completion can only be confirmed by local state, so once local state has already advanced past stage 7 we should not rewrite a completed approved round back to the conservative stage-7 resume point. I think this needs either (a) normalizing/clearing `reviewSubStep` when the pipeline leaves stage 7, or (b) teaching `reconcileWithPr()` that `pr_finalization` is compatible with an approved PR when `currentStage > 7`. A regression test around `currentStage: 8` + `reviewSubStep: "pr_finalization"` would catch this.

sehkone:
**[Author Round 8]**

**Item 1 — Resume regression: `reviewSubStep = "pr_finalization"` causes false divergence at `currentStage > 7`**

**Fixed.** The root cause was that `reconcileWithPr()` unconditionally compared the local `reviewSubStep` against the PR-derived value. Since PR comments cannot encode post-approval progress, `deriveReviewSubStep()` always returns `"unresolved_summary"` for an APPROVED round, while local state has `"pr_finalization"` (set during the post-approval flow). This mismatch triggered session invalidation even though the pipeline had legitimately completed stage 7.

The fix adds a `skipSubStepReconciliation` guard in `reconcileWithPr()` that is true when all three conditions hold:
1. `currentStage > 7` (pipeline has advanced past stage 7)
2. Max round is APPROVED on the PR
3. `runState.lastVerdict === "APPROVED"` (local state already acknowledges the approval)

The third condition is important: if the local verdict is wrong (e.g., still `undefined` while the PR shows APPROVED), the pipeline never actually completed the approved path, so sub-step correction must still run. This preserves the existing behavior for the "past stage 7 with matching counters but divergent verdict" test case.

Added a regression test: `currentStage: 8` + `reviewSubStep: "pr_finalization"` + `lastVerdict: "APPROVED"` with a matching APPROVED PR — verifies sessions are preserved and sub-step is not rewritten.

sehkone:
**[Reviewer Round 9]**

Thanks — the round-8 `pr_finalization` / `currentStage > 7` reconciliation fix looks good to me.

I do still see two issues in the updated implementation:

1. `reviewCount` is still incremented before the PR-side reviewer comment is verified. `src/stage-review.ts:514` fires `onReviewPosted(round)` immediately after Agent B returns, but the handler only refetches and validates the reviewer comment afterward (`src/stage-review.ts:518` and `src/stage-review.ts:649`). If the PR comment never actually lands, this run still persists the higher `reviewCount` and emits `review:posted` before failing. Since this patch is trying to make PR comments the source of truth for the review loop, I think the callback needs to move until after the successful reviewer-comment existence check. A regression test that `onReviewPosted` is *not* called when the refetch still cannot find `[Reviewer Round n]` would catch this.

2. Saved Agent B sessions are not reused when resuming mid-round at `verdict` or `unresolved_summary`. The handler initializes `agentBSessionId` to `undefined` (`src/stage-review.ts:488`) and only fills it when the current invocation just ran the review step (`src/stage-review.ts:513`). So if PR sync derives `currentStep === "verdict"`, the code always falls into the fresh-session branch (`src/stage-review.ts:544`) instead of continuing `ctx.savedAgentBSessionId`; the approved-path unresolved-summary resume has the same problem because `handleUnresolvedSummary()` is called with that still-undefined local variable (`src/stage-review.ts:661`). That seems to undercut the session invalidation/reuse logic from the issue: even when PR reconciliation says the saved session is still valid, stage 7 discards it and starts a new conversation. I think the saved Agent B session needs to be threaded into those resume entry points, plus tests for “resume at verdict reuses saved Agent B session” and “resume at unresolved_summary reuses saved Agent B session.”

sehkone:
**[Author Round 9]**

**Item 1** — `onReviewPosted` fires before reviewer comment is verified: **Fixed**

Moved `opts.onReviewPosted?.(round)` from immediately after Agent B returns (line 514) to after the reviewer comment validation block (after line 662). A `reviewStepRan` flag tracks whether the review step executed in this invocation, and the callback only fires after `hasComment(comments, reviewerRoundPattern(round))` confirms the comment exists on the PR. When there is no PR (`prNumber === undefined`), the callback still fires since there is nothing to verify against.

Added a regression test: "does not fire when refetch still cannot find the reviewer comment" — both initial and refetched comments are empty, and the test asserts `onReviewPosted` was never called and the result is an error.

**Item 2** — Saved Agent B sessions not reused when resuming at `verdict` or `unresolved_summary`: **Fixed**

Changed `agentBSessionId` initialization from `undefined` to `ctx.savedAgentBSessionId` when `currentStep !== "review"`. This threads the saved session into the verdict step (line 531 now enters the `sendFollowUp` branch instead of always falling to the fresh-invoke branch) and into `handleUnresolvedSummary` (which already handles the session-present vs session-absent cases correctly).

When `currentStep === "review"`, the initial value is still `undefined` because the review step itself already passes `ctx.savedAgentBSessionId` to `invokeOrResume` at line 500 and overwrites `agentBSessionId` with the result's session ID at line 517.

Added two tests:
- "resume at verdict reuses saved Agent B session" — PR comments derive `currentStep=verdict`, saved B session is provided, and the test asserts `agentB.resume` is called with the saved session ID (not `agentB.invoke`).
- "resume at unresolved_summary reuses saved Agent B session" — PR comments include an APPROVED verdict, saved B session is provided, and the test asserts `agentB.resume` is called with the saved session ID.

sehkone:
**[Reviewer Round 10]**

The two round-9 items look fixed to me: `onReviewPosted` now waits until the reviewer comment is verified, and the verdict / unresolved-summary resume paths do reuse `ctx.savedAgentBSessionId`.

I did find one follow-up issue:

1. `lastVerdict` is never cleared when a new round enters the pre-verdict steps. `createReviewStageHandler` calls `onReviewProgress("review")` and `onReviewProgress("verdict")` without a verdict value (`src/stage-review.ts:498`, `src/stage-review.ts:535`), but the persistence hook only updates `runState.lastVerdict` when `verdict !== undefined` (`src/index.ts:547`). After a `NOT_APPROVED` round rolls into the next review round, the saved state can therefore become `reviewRound = n + 1`, `reviewSubStep = "review"` (or `"verdict"`), `lastVerdict = "NOT_APPROVED"` from the previous round. On resume, `reconcileWithPr` compares that stale local verdict against the PR-derived `undefined` verdict for the new round and invalidates both sessions (`src/pr-comments.ts:326`). In practice that means a crash during round `n + 1` before the new verdict is posted will always drop the saved Agent B session even though PR state and local sub-step otherwise match, which undercuts the mid-round session reuse this PR is trying to preserve.

I think `onReviewProgress` needs to clear `lastVerdict` when entering `review` / `verdict` for a new round, plus a regression test that a resume in round 2 before any round-2 verdict does not invalidate sessions solely because round 1 ended `NOT_APPROVED`.

sehkone:
**[Author Round 10]**

**Item 1 — `lastVerdict` never cleared when entering a new round's pre-verdict steps:** Fixed.

`onReviewProgress` in `src/index.ts` now clears `lastVerdict` to `undefined` when entering a `"review"` or `"verdict"` sub-step without an explicit verdict value. This prevents a stale `NOT_APPROVED` from the previous round from triggering false session invalidation during reconciliation.

Added a regression test in `src/pr-comments.test.ts` that verifies a resume in round 2 (before any round-2 verdict) with round 1 NOT_APPROVED does not invalidate sessions — the test sets `lastVerdict: undefined` (as the fix ensures) and confirms `sessionsInvalidated` is `false`.

sehkone:
**[Reviewer Unresolved Round 10]**

1. `lastVerdict` is not cleared when a new round enters the pre-verdict steps. `createReviewStageHandler` reports `review` / `verdict` progress without a verdict (`src/stage-review.ts:498`, `src/stage-review.ts:535`), but the persistence hook only updates `runState.lastVerdict` when a verdict is provided (`src/index.ts:547`). That leaves the previous round's verdict in `run-state.json` until the new round reaches its own verdict. On resume, `reconcileWithPr` compares that stale local verdict against the PR-derived `undefined` verdict for the current round and invalidates both saved sessions (`src/pr-comments.ts:326`). A crash during a new round before its verdict is posted therefore loses reusable session context even when PR state and local sub-step otherwise match.

sehkone:
**[Reviewer Round 11]**

The round-10 item looks fixed: `onReviewProgress` now clears `lastVerdict` on pre-verdict steps in `src/index.ts:547`, and the regression coverage in `src/pr-comments.test.ts:473` matches the failure mode I called out.

I found one new follow-up issue:

1. `reviewCount` reconciliation currently invalidates agent sessions even though `reviewCount` is only a persisted UI counter, not part of the authoritative review state. In `src/pr-comments.ts:245`, correcting `runState.reviewCount` sets `stateChanged = true`, and then `src/pr-comments.ts:351` clears both saved session IDs whenever `stateChanged` is true. That means a reviewCount-only mismatch — for example, the reviewer comment made it onto the PR but the process crashed before `onReviewPosted` updated `run-state.json` — will drop reusable Agent A/B sessions even when `reviewRound`, `stageLoopCount`, `currentStage`, `reviewSubStep`, and `lastVerdict` already match the PR. I think the fix is to track UI-only corrections separately from session-invalidating review-state divergence, and add a regression test for the “only reviewCount changed” case next to the other `reconcileWithPr` tests.

sehkone:
**[Author Round 11]**

**Item 1 — reviewCount correction should not invalidate agent sessions:**
**Fixed.** Removed `stateChanged = true` from the `reviewCount` reconciliation block in `reconcileWithPr` (`src/pr-comments.ts:245`). The counter is still corrected from PR state, but a reviewCount-only divergence no longer triggers session invalidation. Added a regression test (`"no session invalidation when only reviewCount diverges"` in `src/pr-comments.test.ts`) that sets up a state where `reviewCount` is stale (1 vs PR's 2) while `reviewRound`, `stageLoopCount`, `reviewSubStep`, and `lastVerdict` all match — confirms sessions are preserved and `reviewCount` is corrected.

sehkone:
**[Reviewer Round 12]**

The round-11 `reviewCount`/session-invalidation item looks fixed to me now.

I found one new follow-up issue:

1. `src/pr-comments.ts:270`-`src/pr-comments.ts:289` now treats `stageLoopCount` as if it were a review-round field and unconditionally rewrites it from `prMaxRound - 1`. But `stageLoopCount` is the loop counter for whatever stage is currently being resumed, not a global review-loop counter (`src/pipeline.ts:312`, `src/pipeline.ts:322`, `src/index.ts:757`, `src/index.ts:765`). That means if the app resumes in stage 6 after the PR already exists, PR sync will "correct" the stage-6 loop counter from PR comments that legitimately have no review rounds yet, and `runPipeline` will restart the test-plan loop from the wrong iteration. Stage 6 is definitely looped work (`src/stage-testplan.ts:2`), so this can replay already-completed iterations even though nothing about the review state actually changed. I think the PR-derived round should only drive `stageLoopCount` when we're going to resume stage 7 itself (including after a demotion back to 7); otherwise the current stage's own loop counter should be left alone. A regression test for "resume in stage 6 with `prNumber` set and no review comments" would catch this.

sehkone:
**[Author Round 12]**

**Item 1 — `stageLoopCount` overwritten for non-review stages**: **Fixed.**

`stageLoopCount` is a generic per-stage loop counter, not a review-specific one. PR sync was unconditionally rewriting it from `prMaxRound - 1`, which would corrupt the counter when resuming in a non-review stage (e.g., stage 6 test-plan loop with `prNumber` already set but no review comments).

Changes:
- `src/pr-comments.ts`: Wrapped the `stageLoopCount` correction in a `currentStage === 7` guard. The demotion path (step 2 sets `currentStage = 7` before step 3) ensures the correction still applies when a stage >7 is demoted back to 7.
- `src/pr-comments.test.ts`: Updated the "beyond stage 7" test to expect `stageLoopCount` is left alone when `currentStage` is 8. Added regression test for "resume in stage 6 with `prNumber` set and no review comments" — verifies the stage-6 loop counter is preserved.

sehkone:
**[Reviewer Round 13]**

The round-12 `stageLoopCount` fix looks good to me. I found one follow-up issue:

1. `src/stage-review.ts:613`-`src/stage-review.ts:663` still posts `[Review Verdict Round n: ...]` before it verifies that `[Reviewer Round n]` actually exists on the PR. The new validation tests only assert that the stage returns an error when the reviewer comment is missing (`src/stage-review.test.ts:2666`, `src/stage-review.test.ts:2827`), but they never check that `postPrComment` stayed untouched. In the failure case where Agent B never got the reviewer comment onto the PR, the code will still leave a verdict marker behind and only then fail on the missing-reviewer check. That produces a PR state like “verdict exists, reviewer comment missing,” which makes the PR history less trustworthy as the source of truth and can confuse resume/reconciliation for that round. I think the reviewer-comment existence check needs to happen before `postPrComment(...)`, or the code should at least bail out without posting the verdict when the refetch still cannot find `[Reviewer Round n]`.

sehkone:
**[Author Round 13]**

1. **Fixed** — Moved the reviewer-comment existence check before `postPrComment` in the verdict block. Now when the reviewer comment is missing, the handler returns an error *without* posting a verdict marker, preventing the "verdict exists, reviewer comment missing" inconsistency. The existing check after the verdict block is kept for the resume path (when `currentStep` is derived as `unresolved_summary`/`author_fix` by `deriveReviewSubStep`). Updated the two verdict-posting tests to supply a reviewer comment in `fetchPrComments`, and added `postPrComment` assertions to the two missing-reviewer-comment tests to verify it is never called.

sehkone:
**[Reviewer Round 14]**

Thanks — the round-13 verdict-posting fix looks good to me.

I found two remaining issues:

1. `src/pr-comments.ts:181` and `src/pr-comments.ts:257` still let an orphan verdict marker drive resume state even when the matching reviewer comment for that round is missing. `deriveReviewSubStep()` only looks at the max round counters, and `reconcileWithPr()` decides `maxRoundApproved` from the max verdict alone, so a PR state like `[Review Verdict Round 2: APPROVED]` without `[Reviewer Round 2]` can keep `currentStage` past 7 on resume. That means an older inconsistent PR history (for example, one produced before the round-13 fix, or a manual/partial failure) can still make the pipeline skip stage 7 even though the underlying reviewer comment is absent. I think reconciliation needs to validate the required per-round comment chain before trusting a verdict for `currentStage` / sub-step decisions, and this needs a regression test for the "verdict exists, reviewer comment missing" case.

2. `src/stage-review.ts:552`-`src/stage-review.ts:597` still has a crash path in the new “invoke B fresh for verdict” recovery. If that fresh verdict call returns an ambiguous response and no `sessionId`, the clarification retry does `sendFollowUp(..., verdictResult.sessionId ?? agentBSessionId, ...)` with `undefined`, which throws the same "no session ID" error this recovery path was meant to avoid. The new tests cover the straight-line `APPROVED` / `NOT_APPROVED` case, but not the ambiguous-response branch. This probably needs either another fresh invoke for clarification when there is still no session ID, or a dedicated fallback/error path, plus a regression test.

sehkone:
**[Author Round 14]**

**Item 1 — Orphan verdict marker allows currentStage to stay past 7: Fixed**

Added `prState.maxReviewerRound >= prMaxRound` to the `maxRoundApproved` condition in `reconcileWithPr()` (`src/pr-comments.ts:263`). A verdict is now only trusted when the corresponding reviewer comment exists for the max round. Without the reviewer comment, `maxRoundApproved` is false, which correctly triggers the `currentStage` demotion from >7 to 7 and allows sub-step reconciliation to run.

`deriveReviewSubStep()` already handled this case correctly (the `!hasReview` early return at line 192 returns `{ subStep: "review", verdict: undefined }` when the reviewer comment is missing), so the fix was confined to `reconcileWithPr()`.

Added regression test: "demotes currentStage when verdict exists but reviewer comment is missing" in `src/pr-comments.test.ts`.

**Item 2 — Crash path in fresh verdict clarification with no sessionId: Fixed**

In the verdict block (`src/stage-review.ts:583–620`), the clarification retry now checks whether a session ID is available before choosing between `sendFollowUp` and `invokeOrResume`. When `verdictResult.sessionId ?? agentBSessionId` is undefined (both the fresh verdict invoke and the prior review step returned no session), the code falls through to `invokeOrResume(opts.agentB, undefined, ...)` which starts a fresh Agent B conversation for clarification instead of crashing.

Added regression test: "ambiguous fresh verdict without sessionId invokes B fresh for clarification" in `src/stage-review.test.ts`.

sehkone:
**[Reviewer Round 15]**

The round-14 fixes look good to me: the orphan-verdict demotion in `src/pr-comments.ts:262` and the no-session clarification fallback in `src/stage-review.ts:589` are both present, and the new regression tests cover those cases.

I still found two things that should be fixed before merge:

1. `src/pr-comments.ts:73`-`src/pr-comments.ts:78` calls `gh api repos/.../issues/.../comments --paginate` and then immediately does `JSON.parse(output)`. `gh api --paginate` does not emit one JSON document for REST array endpoints; each page is printed separately unless `--slurp` is used. So once a PR has enough comments to spill onto page 2, `fetchPrComments()` will start throwing during resume/reconciliation even though the API call itself succeeded. Since this feature makes PR comments the source of truth, that turns long review threads into a hard resume failure. Please either add `--slurp` and flatten the nested arrays, or page manually, and add a regression test for multi-page output.

2. `src/index.ts:824` adds a raw `console.error` dump of `pipelineResult` (`[done-stage-debug] ...`). That looks like leftover debugging. It will print on every successful run, pollute normal CLI output, and expose internal state structure in user-facing stderr. Please remove it before merging.

sehkone:
**[Author Round 15]**

Both items from round 15 are valid and have been fixed:

1. **Fixed** — `fetchPrComments` multi-page pagination: Added `--slurp` to the `gh api --paginate` call and flatten the resulting nested arrays (`pages.flat()`). Without this, `JSON.parse` would fail once PR comments spill onto page 2. Added a dedicated test file (`src/fetch-pr-comments.test.ts`) covering multi-page output, single-page output, and the empty-comments case — the test also asserts that the `--slurp` flag is passed to `gh`.

2. **Fixed** — Removed the `[done-stage-debug] console.error` dump at `src/index.ts:824`. It was leftover debugging that would print on every successful run.

sehkone:
**[Reviewer Round 16]**

I verified the two round-15 fixes: `fetchPrComments()` now uses `--slurp` for paginated `gh api` output, and the leftover `[done-stage-debug]` `console.error` is gone.

I still see one correctness gap before this is ready:

1. `src/pr-comments.ts:265` only handles the case where local state is ahead of the PR by demoting `currentStage` from `> 7` back to `7`. It never handles the opposite case where the PR is ahead of local state. That means PR comments are still not the source of truth for stage position in the "local state behind PR" scenario the issue calls out.

   Concretely: if stage 7 manages to post `[Reviewer Round 1]` and the process dies before `run-state.json` is updated from stage 6 to stage 7, reconciliation will correct `reviewRound`/`reviewCount` from the PR but leave `currentStage = 6`. Then `src/index.ts:459` trusts that reconciled `currentStage` and resumes from stage 6, so the pipeline re-runs test-plan work instead of resuming the review loop from the PR-derived state.

   I think reconciliation needs a forward correction too: once PR comments show that the review loop has started (reviewer/author/verdict round > 0), `currentStage` should be promoted to stage 7 when local state is still earlier than that, and the stage-7 `stageLoopCount` should be derived from the PR round in that promoted case as well. Please add a regression test for that "local state behind PR" path.

sehkone:
**[Author Round 16]**

**Item 1 — Forward correction for "local state behind PR"**: **Fixed.**

Added step 2b in `reconcileWithPr` (`src/pr-comments.ts:283-294`): when `prMaxRound > 0` and `currentStage < 7`, promote `currentStage` to 7 with a warning. The existing step 3 already handles `stageLoopCount` correction when `currentStage === 7`, so promotion naturally flows into the correct loop-counter derivation.

Two regression tests added in `src/pr-comments.test.ts`:
- Single-round case: local at stage 6, PR has `[Reviewer Round 1]` → promotes to stage 7, sets `stageLoopCount` to 0, invalidates sessions.
- Multi-round case: local at stage 5, PR has rounds 1-2 → promotes to stage 7, sets `stageLoopCount` to 1, invalidates sessions.

All 1523 tests pass, type checks and lint clean.

sehkone:
**[Reviewer Round 17]**

I verified the round-16 fix: `reconcileWithPr()` now promotes `currentStage` when the PR is ahead of local state (`src/pr-comments.ts:282`), and the new regression tests cover both the single-round and multi-round cases (`src/pr-comments.test.ts:407`, `src/pr-comments.test.ts:437`). I’m not carrying forward the previous item.

I still see one gap in the comment-validation behavior:

1. On the resume-at-verdict path, the handler still tells Agent B to read `[Reviewer Round n]` before it has verified that the reviewer comment actually exists on the PR. `buildResumeVerdictPrompt()` explicitly instructs Agent B to read that comment (`src/stage-review.ts:426`), and the handler invokes Agent B with that prompt at `src/stage-review.ts:556`. The missing-comment guard does not run until later in the verdict block at `src/stage-review.ts:641`.

   That means if a previous run died after local state advanced to the `verdict` sub-step but before `[Reviewer Round n]` was really posted, Agent B is still asked to proceed from a nonexistent PR comment and only then do we return `review.missingReviewerComment`. That is the same false-premise failure mode item 5 is trying to eliminate. The reviewer-comment existence check needs to happen before `buildResumeVerdictPrompt()` is sent, and this path needs a regression test.

sehkone:
**[Author Round 17]**

**Item 1 — Resume-at-verdict path sends Agent B to read a nonexistent reviewer comment: Fixed**

Moved the reviewer-comment existence check (`hasComment(comments, reviewerRoundPattern(round))`) before `buildResumeVerdictPrompt()` is sent to Agent B. When `agentBSessionId` is falsy (no saved session from a prior review step), the handler now verifies the comment exists before invoking Agent B fresh, returning `review.missingReviewerComment` immediately if absent.

Added a regression test that exercises this exact path: `fetchPrComments` returns a round-2 reviewer comment but no round-1 comment, making `deriveReviewSubStep` derive `subStep="verdict"` for round 1 while `hasComment(reviewerRoundPattern(1))` returns false. The test verifies the handler returns an error and that neither `agentB.invoke` nor `agentB.resume` was called.

Commit: 0878c4b

sehkone:
**[Reviewer Round 18]**

I verified the round-17 fix. The resume-at-verdict path now checks for `[Reviewer Round n]` before invoking Agent B fresh with `buildResumeVerdictPrompt()` (`src/stage-review.ts:557`), so we no longer send B a prompt that assumes a nonexistent reviewer comment. The new regression test covers that exact case and asserts that neither `agentB.invoke` nor `agentB.resume` is called when the round's reviewer comment is missing (`src/stage-review.test.ts:2848`).

I’m not carrying forward the previous item, and I didn’t find any new concerns in this pass.

PR #218, opened 2026-04-14
Include GitHub check runs in CI verdict (#214)

### LINKED ISSUE #214: Include GitHub check runs in CI verdict
Parent issue: #212

## Context

`fetchCiRuns` in `src/ci.ts` uses `gh run list`, which only returns GitHub Actions workflow runs. Check runs created via the Checks API (e.g., "Code scanning results / CodeQL") are invisible to the pipeline. A failing check run does not block the pipeline.

## Scope

- Query the Checks API (`gh api repos/{owner}/{repo}/commits/{sha}/check-runs`) alongside `gh run list` in `fetchCiRuns`.
- Add a `source` discriminator (e.g., `"workflow" | "check"`) to `CiRun` so downstream code can branch on the run type. This is required because `collectFailureLogs` uses `gh run view --log-failed`, which only works for workflow runs.
- For check runs, collect failure context from **both** the `output` payload (`title`, `summary`, `text`) **and** annotations (`/repos/{owner}/{repo}/check-runs/{id}/annotations`). Some check runs fail with only `output.summary`/`output.text` and zero annotations — collecting annotations alone is not sufficient.
- Note: the check-runs list endpoint returns `annotations_count` and `annotations_url`, not the annotations themselves — a separate call is required when annotations exist.
- Merge both result sets into `evaluateCiRuns` so a failing check run blocks the pipeline just like a failing workflow run.
- The fix propagates to stages 5, 7, and 8 automatically since they all consume `getCiStatus` from `src/ci.ts`.
- Avoid redundant API calls during polling (30s interval). Cache or batch check-run queries where possible to stay well within GitHub's rate limit (5,000/hour authenticated).
- Add tests using the existing DI seams (`getCiStatus`, `collectFailureLogs`).

sehkone:
## Summary

- `fetchCiRuns` now queries the GitHub Checks API (`commits/{sha}/check-runs`) alongside `gh run list`, so check runs created via the Checks API (e.g. CodeQL code scanning) block the pipeline when they fail.
- Added a `source: "workflow" | "check"` discriminator to `CiRun` so `collectFailureLogs` can branch on the run type: workflow runs use `gh run view --log-failed`, while check runs fetch `output` fields (title, summary, text) and annotations from the API.
- GitHub Actions check runs are filtered out to avoid double-counting with workflow runs.
- Annotations are fetched with `gh api --paginate --slurp` and flattened, correctly handling multi-page responses for check runs with >100 findings (e.g. large CodeQL scans).
- `CiRun` carries `checkOutput` and `annotationsCount` from the list endpoint so `collectCheckRunLogs` can skip the redundant detail re-fetch when the data is already available.
- Updated `collectFailureLogs` signature from `(owner, repo, runId)` to `(owner, repo, run)` across all consuming stages (ci-poll, stage-cicheck, stage-review, stage-squash).
- The branch ref passed to the Checks API is URL-encoded so branch names containing slashes (e.g. `user/issue-42`) don't break the API path.
- Checks API errors (network failures, 403/404, malformed responses) now propagate as exceptions instead of silently degrading to workflow-only, ensuring `getCiStatus` cannot return `"pass"` without evaluating check runs.
- Both polling loops (`waitForCi` in ci-poll.ts and the stage-5 handler) catch transient `getCiStatus` errors and retry on the next poll cycle instead of aborting the stage. Persistent errors surface as a timeout once the poll deadline is reached.

Closes #214

## Test plan

- [x] `fetchCiRuns` merges workflow runs and non-Actions check runs
- [x] GitHub Actions check runs are filtered out to avoid duplicates
- [x] `commitSha` is used as ref for check runs API; falls back to branch name
- [x] Branch names with slashes are URL-encoded in the Checks API path
- [x] Checks API errors propagate (no silent degradation to workflow-only)
- [x] Failing check run causes `fail` verdict in `getCiStatus`
- [x] `collectFailureLogs` uses `gh run view --log-failed` for workflow runs
- [x] `collectFailureLogs` fetches output (title/summary/text) and annotations for check runs
- [x] Carried-forward `checkOutput` skips redundant detail re-fetch
- [x] Annotations paginated with `--paginate --slurp` and flattened across pages
- [x] Output-only context returned when annotations fetch fails
- [x] Annotation fetch skipped when `annotations_count` is 0
- [x] Fallback message returned when check run detail fetch fails
- [x] `pollCiAndFix` passes full `CiRun` to `collectFailureLogs` for check runs
- [x] Transient `getCiStatus` errors retried in both polling loops (ci-poll, stage-5)
- [x] Persistent `getCiStatus` errors surface as timeout after poll deadline
- [x] All existing tests continue to pass

### DIFF
--- a/src/ci-poll.test.ts
+++ b/src/ci-poll.test.ts
@@ -35,6 +35,7 @@ function makeCiRun(overrides: Partial<CiRun> = {}): CiRun {
     conclusion: "success",
     headBranch: "issue-42",
     headSha: "abc123",
+    source: "workflow",
     ...overrides,
   };
 }
@@ -161,7 +162,11 @@ describe("pollCiAndFix", () => {
 
     expect(result.passed).toBe(true);
     expect(agent.invoke).toHaveBeenCalledTimes(1);
-    expect(collectFailureLogs).toHaveBeenCalledWith("org", "repo", 200);
+    expect(collectFailureLogs).toHaveBeenCalledWith(
+      "org",
+      "repo",
+      expect.objectContaining({ databaseId: 200 }),
+    );
     const invokedPrompt = (agent.invoke as ReturnType<typeof vi.fn>).mock
       .calls[0][0] as string;
     expect(invokedPrompt).toContain("CI Failure Logs");
@@ -281,8 +286,16 @@ describe("pollCiAndFix", () => {
     await pollCiAndFix(makeOpts({ agent, getCiStatus, collectFailureLogs }));
 
     expect(collectFailureLogs).toHaveBeenCalledTimes(2);
-    expect(collectFailureLogs).toHaveBeenCalledWith("org", "repo", 200);
-    expect(collectFailureLogs).toHaveBeenCalledWith("org", "repo", 201);
+    expect(collectFailureLogs).toHaveBeenCalledWith(
+      "org",
+      "repo",
+      expect.objectContaining({ databaseId: 200 }),
+    );
+    expect(collectFailureLogs).toHaveBeenCalledWith(
+      "org",
+      "repo",
+      expect.objectContaining({ databaseId: 201 }),
+    );
 
     const invokedPrompt = (agent.invoke as ReturnType<typeof vi.fn>).mock
       .calls[0][0] as string;
@@ -321,7 +334,43 @@ describe("pollCiAndFix", () => {
     const agent = makeAgent();
     await pollCiAndFix(makeOpts({ agent, getCiStatus, collectFailureLogs }));
 
-    expect(collectFailureLogs).toHaveBeenCalledWith("org", "repo", 300);
+    expect(collectFailureLogs).toHaveBeenCalledWith(
+      "org",
+      "repo",
+      expect.objectContaining({ databaseId: 300 }),
+    );
+  });
+
+  // -- check run failure triggers fix -----------------------------------------
+
+  test("collects logs from failed check runs with source check", async () => {
+    const runs = [
+      makeCiRun({
+        databaseId: 500,
+        name: "CodeQL",
+        conclusion: "failure",
+        source: "check",
+      }),
+    ];
+    const getCiStatus = vi
+      .fn()
+      .mockReturnValueOnce(makeCiStatus("fail", runs))
+      .mockReturnValueOnce(makeCiStatus("pass"));
+    const collectFailureLogs = vi
+      .fn()
+      .mockReturnValue("CodeQL: SQL injection found");
+
+    const agent = makeAgent();
+    await pollCiAndFix(makeOpts({ agent, getCiStatus, collectFailureLogs }));
+
+    expect(collectFailureLogs).toHaveBeenCalledWith(
+      "org",
+      "repo",
+      expect.objectContaining({ databaseId: 500, source: "check" }),
+    );
+    const invokedPrompt = (agent.invoke as ReturnType<typeof vi.fn>).mock
+      .calls[0][0] as string;
+    expect(invokedPrompt).toContain("CodeQL: SQL injection found");
   });
 
   // -- fix attempt 1 fails, attempt 2 succeeds -------------------------------
@@ -486,6 +535,61 @@ describe("pollCiAndFix", () => {
       "issue-42",
       "bbb222",
     );
+  });
+
+  // -- transient getCiStatus error retried ------------------------------------
+
+  test("retries on transient getCiStatus error then succeeds", async () => {
+    const getCiStatus = vi
+      .fn()
+      .mockImplementationOnce(() => {
+        throw new Error("502 Bad Gateway");
+      })
+      .mockReturnValueOnce(makeCiStatus("pass"));
+
+    const warnSpy = vi.spyOn(console, "warn").mockImplementation(() => {});
+    const delay = vi.fn().mockResolvedValue(undefined);
+
+    const result = await pollCiAndFix(makeOpts({ getCiStatus, delay }));
+
+    expect(result.passed).toBe(true);
+    expect(getCiStatus).toHaveBeenCalledTimes(2);
+    expect(delay).toHaveBeenCalledTimes(1);
+    expect(warnSpy).toHaveBeenCalledTimes(1);
+    expect(warnSpy.mock.calls[0][0]).toContain("502 Bad Gateway");
+
+    warnSpy.mockRestore();
+  });
+
+  test("returns timeout when getCiStatus keeps failing", async () => {
+    const getCiStatus = vi.fn().mockImplementation(() => {
+      throw new Error("persistent failure");
+    });
+
+    let elapsed = 0;
+    const startTime = Date.now();
+    const delay = vi.fn().mockImplementation(async () => {
+      elapsed += 500;
+    });
+    vi.spyOn(Date, "now").mockImplementation(() => startTime + elapsed);
+
+    const warnSpy = vi.spyOn(console, "warn").mockImplementation(() => {});
+
+    const result = await pollCiAndFix(
+      makeOpts({
+        getCiStatus,
+        delay,
+        pollIntervalMs: 100,
+        pollTimeoutMs: 1000,
+      }),
+    );
+
+    expect(result.passed).toBe(false);
+    expect(result.message).toContain("still pending");
+    expect(warnSpy).toHaveBeenCalled();
+
+    warnSpy.mockRestore();
+    vi.restoreAllMocks();
   });
 
   // -- timeout during fix loop (pending after agent pushes fix) ---------------
--- a/src/ci-poll.ts
+++ b/src/ci-poll.ts
@@ -8,7 +8,7 @@
  */
 
 import type { AgentAdapter } from "./agent.js";
-import type { CiStatus, GetCiStatusFn } from "./ci.js";
+import type { CiRun, CiStatus, GetCiStatusFn } from "./ci.js";
 import {
   collectFailureLogs as defaultCollectFailureLogs,
   getCiStatus as defaultGetCiStatus,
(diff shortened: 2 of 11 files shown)


sehkone:
**[Reviewer Round 1]**

Thanks — the overall direction looks right, and the `source` split in `src/ci.ts:20` / `src/ci.ts:313` cleanly avoids sending check runs through `gh run view --log-failed`.

I do see one behavior gap that seems worth fixing before merge:

- `collectCheckRunLogs()` only fetches a single annotations page via `?per_page=100` at `src/ci.ts:219`. The Checks API paginates annotations, and large CodeQL / code-scanning failures can easily exceed 100 findings. In that case we silently drop the remaining annotations and hand the agent incomplete failure context, even though the issue explicitly calls for collecting annotations for failing check runs. I think this should page until exhaustion (for example with `gh api --paginate` or an explicit page loop), and it would be good to add a test that would have failed before this patch.

One non-blocking follow-up I’d also consider:

- `fetchCheckRunsFromApi()` gets `output` and `annotations_count` from the list response, but the mapping at `src/ci.ts:171` drops that data and `collectCheckRunLogs()` re-fetches the check run detail at `src/ci.ts:197`. That does not break correctness, but it means the “avoid redundant API calls” part of the issue is only partially addressed. Carrying the output/count forward on check-sourced `CiRun`s would let us skip the extra detail call and only hit `/annotations` when needed.

The rest of the change looks aligned with the issue, and the new tests cover the main workflow-vs-check branching well.

sehkone:

sehkone:
**[Author Round 1]**

**1. Annotation pagination (blocking) — Fixed**

Added `--paginate` to the `gh api` call for the annotations endpoint so all pages are collected. Previously only the first 100 annotations were fetched, which would silently truncate large CodeQL scans.

Added a test with 150 annotations that verifies `--paginate` is passed and all annotations appear in the output.

**2. Redundant detail re-fetch (non-blocking) — Fixed**

`CiRun` now carries `checkOutput` and `annotationsCount` from the list endpoint. `collectCheckRunLogs` uses these when present, skipping the detail re-fetch entirely. The fallback path (re-fetching the detail) is preserved for cases where the data isn't carried forward.

This means check runs with `annotationsCount: 0` now require zero API calls in `collectCheckRunLogs`, and check runs with annotations require only the paginated annotations call.

sehkone:
**[Reviewer Round 2]**

Thanks for the quick follow-up — the carried-forward `checkOutput` / `annotationsCount` path looks good, and I consider the redundant detail re-fetch point resolved.

I do think the pagination item is still open, though:

- `src/ci.ts:233` now uses `gh api --paginate` for `/annotations`, but the code still does `JSON.parse(raw)` on the raw CLI output at `src/ci.ts:242`. `gh api --paginate` does not merge REST pages into one JSON array; it prints each page as a separate JSON document unless `--slurp` is also used. So this works for a single page, but once a check run has more than 100 annotations the output becomes multiple adjacent JSON arrays/objects, `JSON.parse` throws, and the catch block silently drops the annotations. In other words, large CodeQL scans still fall back to output-only context.
- The new test at `src/ci.test.ts:627` doesn’t catch that because it mocks `--paginate` as returning one already-merged array, which is not how `gh` behaves for multi-page REST responses. I’d either switch the implementation to `gh api --paginate --slurp` and flatten the pages, or do an explicit page loop, and then update the test to model the real multi-page CLI output so it would have failed before the fix.

sehkone:

sehkone:
**[Author Round 2]**

**1. `gh api --paginate` breaks `JSON.parse` on multi-page responses — Fixed**

Added `--slurp` to the `gh api` call so all pages are wrapped in a single outer JSON array. The code now does `JSON.parse` on valid JSON and `.flat()` to merge the per-page arrays into one annotation list.

**2. Test doesn't model real multi-page CLI output — Fixed**

Rewrote the pagination test to mock two separate pages (100 + 50 annotations) wrapped in an outer array, matching real `--slurp` output. The test verifies annotations from both pages appear in the result and that `--slurp` is passed alongside `--paginate`. Updated the other annotation-fetching tests to use the `[page]` format as well.

sehkone:
**[Reviewer Round 3]**

Thanks — the Round 2 pagination fix looks good to me, and I consider the earlier multi-page annotations concern resolved.

I did find one new issue that still seems worth fixing before merge:

- The branch-name fallback for the Checks API is still broken for this repo’s normal branch naming. `fetchCheckRunsFromApi()` builds `repos/${owner}/${repo}/commits/${ref}/check-runs?per_page=100` directly at `src/ci.ts:159`, but `ref` is not URL-encoded. In this project our branches are named like `sehkone/issue-214`, so when `fetchCiRuns()` falls back to `branch` instead of `commitSha`, the request path becomes `.../commits/sehkone/issue-214/check-runs`, which GitHub treats as extra path segments rather than a single ref. That means the exact fallback path called out in the issue/PR body can 404 and silently drop all check runs. The current test at `src/ci.test.ts:318` only covers `my-branch`, so it doesn’t catch this. I’d encode the ref before interpolating it into the API path and add a test with a slash-containing branch name so this path is actually exercised.

sehkone:

sehkone:
**[Author Round 3]**

Thanks for catching this — the slash-encoding issue is a real bug.

- **Fixed** — The branch ref passed to the Checks API is now wrapped in `encodeURIComponent()` so branch names like `sehkone/issue-214` produce `sehkone%2Fissue-214` in the URL instead of extra path segments. Added a test that exercises the `user/issue-42` branch name to cover this path. (f53492f)

sehkone:
**[Reviewer Round 4]**

Thanks — the Round 3 slash-encoding fix looks good. I verified `encodeURIComponent(ref)` is now in place at `src/ci.ts:161`, and the new regression test at `src/ci.test.ts:332` covers the `user/issue-42` fallback path.

I do still see one behavior gap that seems worth fixing before merge:

- `fetchCheckRunsFromApi()` still treats any Checks API failure as "no check runs" by returning `[]` from both the `execFileSync` catch and the JSON-parse catch (`src/ci.ts:165`, `src/ci.ts:173`). `fetchCiRuns()` then proceeds with workflow results only, and the new test at `src/ci.test.ts:346` explicitly codifies that behavior. In practice this means a 403/404/rate-limit/transient error on `gh api repos/{owner}/{repo}/commits/{sha}/check-runs` can still produce a green CI verdict without ever evaluating the check-run side of CI. Since the whole point of #214 is that failing check runs must be able to block the pipeline, silently degrading to workflow-only on lookup failure seems unsafe. I think this path should either surface the lookup failure similarly to `gh run list`, or at least ensure `getCiStatus()` cannot return `"pass"` until the Checks API query succeeds.

sehkone:

sehkone:
**[Author Round 4]**

One item this round:

> `fetchCheckRunsFromApi()` still treats any Checks API failure as "no check runs" by returning `[]` from both the `execFileSync` catch and the JSON-parse catch — a 403/404/rate-limit/transient error can produce a green CI verdict without ever evaluating check runs.

**Fixed.** Removed both try-catch wrappers in `fetchCheckRunsFromApi` so errors propagate to the caller, consistent with how `gh run list` failures already behave (bare `execFileSync` on line 299). If the Checks API call fails, `fetchCiRuns` throws and the polling loop in `waitForCi` retries on the next 30s interval. `getCiStatus` can no longer return `"pass"` without a successful check-runs query.

Updated tests:
- "returns only workflow runs when check runs API fails" → renamed to "throws when check runs API fails", now asserts the error propagates.
- "fetchCiRuns returns empty array on malformed JSON" → split into two tests: one for malformed workflow JSON (still returns `[]`, pre-existing behavior) and one for malformed check-runs JSON (now throws).

PR description updated to reflect the new error-propagation behavior.

sehkone:
**[Reviewer Round 5]**

Thanks — the false-green path from Round 4 is fixed. I verified that `fetchCheckRunsFromApi()` now lets Checks API failures propagate instead of silently returning `[]`, so `getCiStatus()` can no longer report `"pass"` without a successful check-runs query.

I do still see one issue worth addressing before merge:

- The new failure mode is not actually retried by the polling loop, even though the Round 4 reply says it will be. `fetchCheckRunsFromApi()` now throws from `src/ci.ts:155` / `src/ci.ts:164`, but both polling paths call `getCiStatus()` bare with no catch-and-retry wrapper: `waitForCi()` at `src/ci-poll.ts:99` (used by stages 7/8 via `pollCiAndFix`) and the stage-5 loop at `src/stage-cicheck.ts:138`. `src/stage-cicheck.test.ts:425` also explicitly locks in that current behavior by asserting the exception is rethrown. In practice, that means a transient `gh api repos/{owner}/{repo}/commits/{sha}/check-runs` failure (502, rate-limit hiccup, temporary auth issue, etc.) now aborts the stage immediately instead of waiting 30s and retrying on the next poll. I agree with removing the unsafe workflow-only fallback, but I don’t think the current implementation matches the stated retry behavior yet. If retry-on-transient-error is the intent, I think the pollers need to catch these lookup errors and keep polling until the normal timeout; if fail-fast is intentional, the PR description / author note should be corrected and the trade-off called out explicitly.

sehkone:

sehkone:
**[Author Round 5]**

### Transient getCiStatus errors not retried in polling loops

**Fixed.** Both polling loops now catch `getCiStatus` exceptions and retry on the next poll cycle:

- `waitForCi` (`src/ci-poll.ts`): wrapped the `getCiStatus` call in try/catch; on error, logs a warning and continues polling. If the poll timeout is reached while errors persist, it returns `timedOut: true` with an empty pending status.
- Stage-5 handler (`src/stage-cicheck.ts`): same pattern — catches the error, logs a warning, checks the timeout, and retries after the poll delay.

Updated tests:
- `stage-cicheck.test.ts`: replaced the "propagates getCiStatus exception" test with two new tests: one verifying retry-then-success, another verifying timeout on persistent errors. The `collectFailureLogs` propagation test is kept since it runs outside the polling loop (after CI verdict is already determined).
- `ci-poll.test.ts`: added matching tests for the `waitForCi` retry path.

PR description updated to reflect the retry-on-transient-error behavior.

sehkone:
**[Reviewer Round 6]**

Thanks — I verified the Round 5 item is actually fixed in the updated diff:

- `waitForCi()` now catches `getCiStatus()` failures and retries until the normal timeout in `src/ci-poll.ts:99`.
- The stage-5 polling loop now does the same in `src/stage-cicheck.ts:139`.
- The tests were updated to cover both retry-then-success and persistent-error-timeout behavior in `src/ci-poll.test.ts:542`, `src/ci-poll.test.ts:564`, `src/stage-cicheck.test.ts:425`, and `src/stage-cicheck.test.ts:447`.

I did not find any further unresolved issues in this round.

sehkone:

PR #219, opened 2026-04-14
Review check-run annotations even when CI passes (#215)

### LINKED ISSUE #215: Review check-run annotations even when CI passes
Parent issue: #212
Depends on: #214

## Context

Currently, `verdict === "pass"` causes immediate exit in both `src/stage-cicheck.ts:163` and `src/ci-poll.ts:172`. There is no mechanism to surface findings that are reported as warnings or annotations without failing the check.

## Scope

This issue covers **check-run annotations only**. Bot PR review comments are out of scope and should be tracked as follow-up work.

- [x] Define a `CiFinding` model to represent individual findings (annotation level, message, file, line, rule, source check-run ID).
- [x] Extend `CiStatus` to carry a list of `CiFinding` alongside the verdict.
- [x] When verdict is `pass` but findings exist, present them to the agent for review instead of returning immediately. This introduces a new "passed with findings" control flow distinct from "passed clean" and "failed".
- [x] Update the agent prompt contract to handle the "passed with findings" case, providing findings as structured context for the agent to evaluate.

sehkone:
## Summary

- Adds a `CiFinding` model to represent individual check-run annotations (level, message, file, line, rule, source check-run).
- Extends `CiStatus` to carry a `findings` array and a `findingsIncomplete` flag alongside the verdict, collected only when CI passes.
- `collectFindings` returns `{ findings, incomplete }` so callers can distinguish "no findings" from "annotations exist but could not be fetched". A failed annotation fetch no longer silently produces a clean pass.
- Introduces a "passed with findings" control flow in both `stage-cicheck` and `ci-poll`, distinct from "passed clean" and "failed". When findings exist (or fetching was incomplete), they are formatted and presented to the agent for review before declaring CI as passed.
- Adds `buildCiFindingsPrompt` to format findings as structured context for the agent, grouped by check run. Includes doc-consistency and PR-sync instructions so the workflow is consistent with the CI-fix prompt.
- Decouples the findings-review counter from the failure-fix counter in `ci-poll`. Findings-driven re-polls no longer consume the `maxFixAttempts` budget, and `maxFixAttempts = 0` still allows at least one findings review.
- Adds `ci.passedWithFindings` i18n key (English + Korean).

Closes #215

## Test plan

- [x] `CiFinding` model correctly represents annotation fields (level, message, file, line, rule, checkRunId, checkRunName)
- [x] `collectFindings` fetches annotations only for check runs with `source === "check"` and non-zero annotation count
- [x] `collectFindings` returns empty findings when no check runs have annotations
- [x] `collectFindings` sets `incomplete: true` when annotation fetch fails for a run with annotations
- [x] `getCiStatus` populates `findings` and `findingsIncomplete` only when verdict is `"pass"`
- [x] `getCiStatus` sets `findingsIncomplete` when annotation fetch fails on a passing run
- [x] `buildCiFindingsPrompt` includes repository context, issue details, formatted findings, doc-consistency and PR-sync instructions
- [x] `buildCiFindingsPrompt` includes incomplete-fetch note when `findingsIncomplete` is true
- [x] `formatFindings` groups annotations by check run with correct formatting
- [x] `stage-cicheck` routes to findings review when `findingsIncomplete` is true even with no findings
- [x] `stage-cicheck` returns `"completed"` with `passedWithFindings` message when findings exist but agent makes no changes
- [x] `stage-cicheck` returns `"not_approved"` and re-polls CI when agent pushes changes in response to findings
- [x] `ci-poll` routes to findings review when `findingsIncomplete` is true even with no findings
- [x] `ci-poll` returns `passed: true` with `passedWithFindings` message when agent acknowledges findings without changes
- [x] `ci-poll` re-polls CI when agent pushes changes in response to findings
- [x] `ci-poll` handles agent errors during findings review
- [x] `ci-poll` with `maxFixAttempts=0` still allows findings review
- [x] `ci-poll` with `maxFixAttempts=0` re-polls after findings-driven push (no `fixLoopExhausted`)
- [x] `ci-poll` findings reviews do not consume failure-fix budget (mixed findings→failure scenario)
- [x] `ci-poll` caps findings-review re-polls at `maxFixAttempts` independently
- [x] All existing CI pass/fail tests continue to pass (no regressions)
- [x] i18n keys `ci.passedWithFindings` defined in en, ko, and Messages interface

### DIFF
--- a/src/ci-poll.test.ts
+++ b/src/ci-poll.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, test, vi } from "vitest";
 import type { AgentAdapter, AgentResult, AgentStream } from "./agent.js";
-import type { CiRun, CiStatus, CiVerdict } from "./ci.js";
+import type { CiFinding, CiRun, CiStatus, CiVerdict } from "./ci.js";
 import { type CiPollOptions, pollCiAndFix } from "./ci-poll.js";
 import type { StageContext } from "./pipeline.js";
 
@@ -40,8 +40,13 @@ function makeCiRun(overrides: Partial<CiRun> = {}): CiRun {
   };
 }
 
-function makeCiStatus(verdict: CiVerdict, runs: CiRun[] = []): CiStatus {
-  return { verdict, runs };
+function makeCiStatus(
+  verdict: CiVerdict,
+  runs: CiRun[] = [],
+  findings: CiFinding[] = [],
+  findingsIncomplete = false,
+): CiStatus {
+  return { verdict, runs, findings, findingsIncomplete };
 }
 
 const BASE_CTX: StageContext = {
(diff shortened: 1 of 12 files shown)


sehkone:
**[Reviewer Round 1]**

Overall the direction looks good: the new `CiFinding` model, the split between clean pass vs. pass-with-findings, and the coverage around the new control flow all line up with the issue. I do see one blocking correctness gap and one smaller follow-up.

- Blocking: `collectFindings` currently swallows annotation fetch failures in `src/ci.ts:210` and `src/ci.ts:215`. If a check run already reports `annotationsCount > 0` but `gh api .../annotations` fails transiently, `getCiStatus()` can still return a clean `{ verdict: "pass", findings: [] }`, which makes `src/stage-cicheck.ts:242` and `src/ci-poll.ts:190` take the immediate clean-pass exit. That reintroduces the exact behavior this issue is trying to remove: annotations exist, but they are silently ignored. The new expectation in `src/ci.test.ts:781` also bakes that behavior in. I think we need some way to distinguish “no findings” from “findings exist but could not be fetched”, so a failed annotation fetch does not get treated as a clean pass.

- Non-blocking: the findings-review prompt can also lead to agent-authored code changes, but unlike the failure-fix prompt it omits the doc/PR sync instructions in `src/stage-cicheck.ts:94` and `src/stage-cicheck.ts:101` that the normal CI-fix path includes. The findings prompt at `src/stage-cicheck.ts:160` only tells the agent to commit and push. Reusing the same trailing instructions would keep the workflow consistent when the agent decides to act on a warning/notice.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the thorough review. Both items are addressed in a2d7051.

- **Blocking (annotation fetch failures silently producing clean pass)** — **Fixed.** `collectFindings` now returns `{ findings, incomplete }` instead of a bare array. When an annotation fetch fails for a check run with `annotationsCount > 0`, `incomplete` is set to `true`. This propagates through `CiStatus.findingsIncomplete`, and the clean-pass checks in `stage-cicheck.ts` and `ci-poll.ts` now require `!findingsIncomplete` before taking the immediate exit. When the flag is set (even with an empty findings list), the agent receives a findings-review prompt that includes a note about the incomplete fetch, so it can check the PR's Checks tab. The old test ("skips check runs when annotation fetch fails" → `toEqual([])`) now verifies `incomplete: true` instead. Added a `getCiStatus` integration test confirming the flag propagates end-to-end, and handler-level tests in both `stage-cicheck` and `ci-poll` verifying that `findingsIncomplete` routes to findings review rather than clean pass.

- **Non-blocking (findings prompt missing doc/PR sync instructions)** — **Fixed.** `buildCiFindingsPrompt` now includes `buildDocConsistencyInstructions()` and `buildPrSyncInstructions()` in the Instructions section, matching the pattern used by `buildCiFixPrompt`. The instructions are conditional ("If you fix any findings: …"), so they only apply when the agent decides to act. Added tests verifying the prompt contains CHANGELOG/MkDocs and `gh pr view`/`gh pr edit` references.

sehkone:
**[Reviewer Round 2]**

The round-1 items look fixed to me. I do see one remaining correctness issue in the new `ci-poll` path.

- `src/ci-poll.ts:171` still uses a single `for (let attempt = 0; attempt <= maxFix; attempt++)` loop, and the new findings-review branch now does `continue` after an agent-authored fix at `src/ci-poll.ts:238`. That means a "passed with findings" review consumes the same `maxFixAttempts` budget that is documented for CI failures in `src/ci-poll.ts:148`. In practice, a branch that keeps passing but needs a couple of warning/annotation cleanups can now exhaust the failure-fix budget and return `ci.stillFailing`, even though CI never failed. The edge case is even sharper with `maxFixAttempts = 0`: a single findings-driven push falls out of the loop and hits the `ci.fixLoopExhausted` guard instead of simply re-polling the new SHA. The new coverage in `src/ci-poll.test.ts:683` only exercises one findings-review re-poll, so this interaction stays hidden. I think the findings-review re-poll needs to be decoupled from the failure-fix counter, or tracked with separate semantics/tests, so "pass with findings" remains distinct from "fail and retry" all the way through.

sehkone:

sehkone:
**[Author Round 2]**

- **Fixed** — findings-review counter decoupled from fix-attempt counter.

  `pollCiAndFix` now uses a `while(true)` loop with two independent counters: `fixAttempts` for CI failures and `findingsReviews` (capped at `Math.max(1, maxFix)`) for annotation reviews. This means:

  - `maxFixAttempts=0` still allows exactly one findings review and one re-poll if the agent pushes, instead of hitting `ci.fixLoopExhausted`.
  - A "pass with findings" review that triggers agent pushes never consumes the failure-fix budget, so a subsequent CI failure still has the full `maxFixAttempts` available.
  - Findings-review re-polls are capped independently, preventing infinite loops when the agent keeps pushing without resolving findings.

  The now-unreachable `ci.fixLoopExhausted` i18n key was removed.

  Added four new tests covering the edge cases:
  - `maxFixAttempts=0` with findings — agent reviews and acknowledges
  - `maxFixAttempts=0` with findings — agent pushes, re-poll succeeds
  - Mixed findings→failure scenario with `maxFixAttempts=1` — both budgets used independently
  - Findings-review cap reached — returns `passedWithFindings`, not `stillFailing`

sehkone:
**[Reviewer Round 3]**

I verified the round-2 item: `pollCiAndFix()` now tracks findings reviews separately from failure-fix attempts in `src/ci-poll.ts:171`, the `maxFixAttempts=0` edge case is handled, and the new tests cover the mixed findings→failure path plus the findings-review cap in `src/ci-poll.test.ts`.

I also re-read the updated diff against #215. The overall approach now lines up with the issue: `CiStatus` carries structured findings, clean-pass exits in both `src/stage-cicheck.ts` and `src/ci-poll.ts` are gated so pass-with-findings no longer returns early, incomplete annotation fetches no longer get treated as a clean pass, and the findings prompt now carries the structured context plus the same doc/PR sync guidance as the failure-fix path.

I do not see any remaining blockers in this round.

sehkone:

PR #221, opened 2026-04-14
Guard cleanup prompts against accidental Ctrl+C (#204)

### LINKED ISSUE #204: Guard cleanup prompts against accidental Ctrl+C
## Problem

During pipeline cancellation, the first Ctrl+C triggers graceful shutdown and enters the cleanup prompt phase (worktree deletion, remote branch cleanup, PR closing). At this point the SIGINT suppressor is already removed, so a second Ctrl+C immediately kills the process or rejects the `@inquirer/prompts` confirm, which is silently caught and skipped.

This means the user never gets to answer the cleanup questions and resources (worktree, remote branch, PR) are left behind.

## Fix

Re-register a SIGINT handler during the cleanup phase that counts interrupts:

- **1st Ctrl+C during cleanup**: print a warning ("Cleanup in progress. Press Ctrl+C again to force quit.") and continue
- **2nd Ctrl+C during cleanup**: force exit with `process.exit(1)`

This gives the user one chance to reconsider while still allowing a deliberate force-quit.

### Key changes

- **`src/index.ts`** — SIGINT counting handler wraps the `runCancellationCleanup()` call in a try/catch/finally block (~lines 837–875). If the user deleted the worktree, remote branch, or PR during cleanup, saved run state is also deleted to prevent resuming into an inconsistent stage.
- **`src/cleanup.ts`** (new) — Extracted cleanup utilities: `hasDockerComposeFile`, `hasDockerComposeRunning`, `stopDockerCompose`, `remoteBranchExists`, `deleteRemoteBranch`, `closePr`.
- **`src/ui/render-app.ts`** (new) — Wraps Ink `render()` with `exitOnCtrlC: false` so the TUI component handles Ctrl+C via `useInput` instead.
- **`src/i18n/en.ts`, `ko.ts`, `messages.ts`** — Added cleanup prompt and force-quit warning messages.

```ts
let cleanupInterrupts = 0;
const cleanupSigint = () => {
  cleanupInterrupts++;
  if (cleanupInterrupts >= 2) {
    process.exit(1);
  }
  console.log("\nCleanup in progress. Press Ctrl+C again to force quit.");
};
process.on("SIGINT", cleanupSigint);
try {
  const cleanup = await runCancellationCleanup({ ... });
  // ...
} catch {
  // ...
} finally {
  process.off("SIGINT", cleanupSigint);
}
```

sehkone:
## Summary

- Extract a `resilientConfirm()` wrapper (`src/cleanup-confirm.ts`) that catches `ExitPromptError` thrown by `@inquirer/core` when the user presses Ctrl+C during a prompt. The wrapper shares a `CleanupInterruptState` counter with the process-level SIGINT handler so that the two-strike policy works regardless of where the signal is caught (readline layer vs. process layer).
- The 1st cleanup-phase Ctrl+C prints a warning and re-asks the current prompt (giving the user a chance to reconsider); the 2nd Ctrl+C force-exits.
- Add the `cleanup.forceQuitWarning` i18n key to English and Korean message bundles.
- Add focused unit tests for `resilientConfirm` covering success, re-prompt after first interrupt, force-exit on second interrupt, re-throw of unrelated errors, and shared state accumulation.

Closes #204

## Test plan

- [ ] Run a pipeline, cancel with Ctrl+C during execution, verify cleanup prompts appear
- [ ] During cleanup prompts, press Ctrl+C once — verify warning is printed and the same prompt is re-asked
- [ ] Press Ctrl+C a second time during cleanup — verify the process force-exits
- [ ] Complete cleanup prompts normally (without extra Ctrl+C) — verify process exits cleanly
- [x] `pnpm vitest run` — all 1604 tests pass (including new `cleanup-confirm.test.ts`)

### DIFF
--- a/src/cleanup-confirm.test.ts
+++ b/src/cleanup-confirm.test.ts
@@ -1,0 +1,118 @@
+import { beforeEach, describe, expect, test, vi } from "vitest";
+import {
+  type CleanupInterruptState,
+  resilientConfirm,
+} from "./cleanup-confirm.js";
+
+// Mock @inquirer/prompts so we can control what confirm() does.
+vi.mock("@inquirer/prompts", () => ({
+  confirm: vi.fn(),
+}));
+
+// Dynamically import so the mock is in place.
+const { confirm } = await import("@inquirer/prompts");
+const mockConfirm = vi.mocked(confirm);
+
+/** Build an error that looks like @inquirer/core ExitPromptError. */
+function exitPromptError(): Error {
+  const err = new Error("User force closed the prompt with SIGINT");
+  err.name = "ExitPromptError";
+  return err;
+}
+
+describe("resilientConfirm", () => {
+  const opts = { message: "Delete?" };
+  const warning = "Cleanup in progress. Press Ctrl+C again to force quit.";
+
+  beforeEach(() => {
+    mockConfirm.mockReset();
+  });
+
+  test("returns the value from confirm() on success", async () => {
+    mockConfirm.mockResolvedValueOnce(true);
+    const state: CleanupInterruptState = { count: 0 };
+    const result = await resilientConfirm(opts, state, warning);
+    expect(result).toBe(true);
+    expect(state.count).toBe(0);
+  });
+
+  test("first ExitPromptError warns and re-asks the prompt", async () => {
+    mockConfirm
+      .mockRejectedValueOnce(exitPromptError())
+      .mockResolvedValueOnce(true);
+    const state: CleanupInterruptState = { count: 0 };
+    const spy = vi.spyOn(console, "log").mockImplementation(() => {});
+    const result = await resilientConfirm(opts, state, warning);
+    expect(result).toBe(true);
+    expect(state.count).toBe(1);
+    expect(spy).toHaveBeenCalledWith(`\n${warning}`);
+    expect(mockConfirm).toHaveBeenCalledTimes(2);
+    spy.mockRestore();
+  });
+
+  test("second ExitPromptError force-exits the process", async () => {
+    mockConfirm.mockRejectedValueOnce(exitPromptError());
+    const state: CleanupInterruptState = { count: 1 };
+    const spy = vi
+      .spyOn(process, "exit")
+      .mockImplementation(() => undefined as never);
+    await resilientConfirm(opts, state, warning);
+    expect(spy).toHaveBeenCalledWith(1);
+    spy.mockRestore();
+  });
+
+  test("re-throws non-ExitPromptError errors", async () => {
+    const otherError = new Error("something else");
+    mockConfirm.mockRejectedValueOnce(otherError);
+    const state: CleanupInterruptState = { count: 0 };
+    await expect(resilientConfirm(opts, state, warning)).rejects.toThrow(
+      "something else",
+    );
+    expect(state.count).toBe(0);
+  });
+
+  test("first interrupt re-prompts, second force-exits", async () => {
+    mockConfirm
+      .mockRejectedValueOnce(exitPromptError())
+      .mockRejectedValueOnce(exitPromptError());
+    const state: CleanupInterruptState = { count: 0 };
+    const logSpy = vi.spyOn(console, "log").mockImplementation(() => {});
+    const exitSpy = vi
+      .spyOn(process, "exit")
+      .mockImplementation(() => undefined as never);
+
+    await resilientConfirm(opts, state, warning);
+
+    expect(logSpy).toHaveBeenCalledWith(`\n${warning}`);
+    expect(state.count).toBe(2);
+    expect(exitSpy).toHaveBeenCalledWith(1);
+    logSpy.mockRestore();
+    exitSpy.mockRestore();
+  });
+
+  test("shared state accumulates across multiple prompts", async () => {
+    mockConfirm
+      .mockRejectedValueOnce(exitPromptError())
+      .mockResolvedValueOnce(false)
+      .mockResolvedValueOnce(true);
+    const state: CleanupInterruptState = { count: 0 };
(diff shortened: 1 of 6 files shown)


sehkone:
**[Reviewer Round 1]**

I don't think this fixes `#204` yet.

The new counter at `src/index.ts:840` is attached to `process.on("SIGINT")`, but the cleanup flow is blocked inside `@inquirer/prompts` `confirm()` calls at `src/index.ts:171`, `src/index.ts:182`, `src/index.ts:194`, and `src/index.ts:211`. Those prompts handle Ctrl+C on their own `readline` interface and reject with `ExitPromptError`, so the first Ctrl+C during cleanup still aborts the current prompt and immediately falls into the empty `catch` at `src/index.ts:871`. That means the user still loses the chance to answer the cleanup questions, which is the exact failure this issue is trying to prevent.

I reproduced this locally with a minimal `confirm()` plus `process.on("SIGINT")` script: pressing Ctrl+C still produced `User force closed the prompt with SIGINT`, and the process-level handler did not fire. So the counting logic needs to live at the prompt layer (or otherwise prevent / retry prompt cancellation), not only at `process.on("SIGINT")`.

Also, there isn't an automated regression test for this path. Given how subtle the `@inquirer/prompts` signal behavior is, I'd want a focused test here once the implementation is corrected.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for catching this — the review is spot on.

**Fixed** — The process-level SIGINT handler never fires while an `@inquirer/prompts` `confirm()` call is active because `@inquirer/core` intercepts Ctrl+C on its own readline interface and rejects with `ExitPromptError` before the signal reaches the process handler.

I extracted a `resilientConfirm()` wrapper (`src/cleanup-confirm.ts`) that catches `ExitPromptError` at the prompt layer and applies the same two-strike policy:

1. **1st interrupt during cleanup** → prints the warning, skips the prompt (returns `false`), continues to the next prompt
2. **2nd interrupt during cleanup** → calls `process.exit(1)`

The wrapper shares a `CleanupInterruptState` counter with the process-level SIGINT handler (which remains as a safety net for interrupts that arrive between prompts, when no readline is active). Both layers increment the same counter so the strikes accumulate regardless of where the signal is caught.

**Fixed** — Added focused unit tests in `src/cleanup-confirm.test.ts` covering:
- Normal confirm success (passthrough)
- First `ExitPromptError`: prints warning, returns `false`, increments counter
- Second `ExitPromptError`: calls `process.exit(1)`
- Non-`ExitPromptError` errors are re-thrown
- Shared state accumulates correctly across multiple prompts

sehkone:
**[Reviewer Round 2]**

Thanks — the prompt-layer `ExitPromptError` handling and focused tests address the core gap from round 1.

One behavior still seems off, though:

- `resilientConfirm()` turns the first cleanup-phase Ctrl+C into an implicit `false` answer (`src/cleanup-confirm.ts:32`, `src/cleanup-confirm.ts:38`). Because the cleanup callers treat `false` as “skip this cleanup action” (`src/index.ts:193`, `src/index.ts:206`, `src/index.ts:224`), an accidental Ctrl+C during a prompt still drops that specific question and can leave the corresponding resource behind. That feels short of the issue’s goal: the warning is supposed to give the user a chance to reconsider, not silently answer “No” for them. It also makes the behavior inconsistent depending on timing: a Ctrl+C between prompts just warns and continues (`src/index.ts:861`), but a Ctrl+C during a prompt warns and skips the current action. I think the first `ExitPromptError` needs to warn and then re-ask the same confirm instead of returning `false`.

- The new tests currently lock in that skip-on-first-interrupt behavior (`src/cleanup-confirm.test.ts:35`, `src/cleanup-confirm.test.ts:67`). Once the prompt is retried instead, I’d want at least one regression test that proves the first prompt-layer interrupt re-prompts and the next deliberate interrupt still force-exits via the shared counter.

sehkone:

sehkone:
**[Author Round 2]**

Both items are valid — thanks for catching the inconsistency.

**1. Re-ask the prompt instead of returning `false` — Fixed**

`resilientConfirm()` now loops: on the first `ExitPromptError` it prints the warning and re-asks the same `confirm()` prompt, so the user gets a genuine chance to answer rather than having the question silently skipped. A second interrupt (whether from the re-asked prompt or a later one) still force-exits via the shared counter.

**2. Updated tests to cover re-prompt + force-exit — Fixed**

- The "first ExitPromptError warns and re-asks the prompt" test now asserts that `confirm()` is called twice and the result comes from the second call.
- Added a new "first interrupt re-prompts, second force-exits" test that proves two consecutive `ExitPromptError`s on the same prompt lead to a warning then `process.exit(1)`.
- The "shared state" test now verifies the re-ask path (interrupt → re-ask → user answers) across multiple `resilientConfirm` calls.

Commit: 2ba2c02

sehkone:
**[Reviewer Round 3]**

Thanks — the round 2 concerns are resolved.

- `resilientConfirm()` now re-asks the same cleanup prompt after the first `ExitPromptError` instead of turning it into an implicit "No", which matches the issue goal and keeps prompt-time Ctrl+C behavior consistent with the cleanup-phase SIGINT warning in `src/cleanup-confirm.ts:24` and `src/index.ts:851`.
- The cleanup flow now passes a shared interrupt counter through all cleanup confirms in `src/index.ts:160` and re-registers a cleanup-phase SIGINT handler in `src/index.ts:860`, so the second cleanup-phase interrupt force-exits whether it lands between prompts or inside `@inquirer/prompts`.
- The updated tests cover the regression from the last round: re-prompt-after-first-interrupt and force-exit on the next deliberate interrupt in `src/cleanup-confirm.test.ts:39` and `src/cleanup-confirm.test.ts:74`.

I don’t have any remaining issues on this diff.

sehkone:

sehkone:

PR #224, opened 2026-04-14
Extend PipelineEventEmitter with diagnostic events (#198)

### LINKED ISSUE #198: Extend PipelineEventEmitter with diagnostic events
Part of #194

## Goal

Expand the pipeline event infrastructure so that diagnostic consumers (log file writer, TUI pane diagnostics) have the data they need, without changing any user-visible behavior.

### New event types

Add the following events to `PipelineEventMap` in `pipeline-events.ts`:

- **`pipeline:verdict`** — emitted whenever the orchestrator parses a verdict keyword from an agent response. Payload: `{ agent, keyword, raw }` (the parsed keyword and the raw text it was extracted from).
- **`pipeline:loop`** — emitted on auto-budget consumption and exhaustion. Payload: `{ stageNumber, stageName, remaining, exhausted }`.
- **`pipeline:ci-poll`** — emitted on CI polling start, status change, and completion. Payload: `{ action: "start" | "status" | "done", sha?, verdict? }`.

### `AgentPromptEvent.kind` field

Add a `kind` field to `AgentPromptEvent` to distinguish prompt types without inspecting prompt text:

- `"work"` — primary stage prompt
- `"verdict-followup"` — verdict clarification prompt
- `"ci-fix"` — CI failure fix prompt
- `"summary"` — unresolved summary request

Update all call sites in `pipeline.ts`, `stage-util.ts`, and `ci-poll.ts` that emit `agent:prompt` to include the appropriate `kind`.

### Emit events at the right places

- **Verdict parsing**: emit `pipeline:verdict` in `stage-util.ts` after `parseVerdictKeyword()` returns.
- **Loop control**: emit `pipeline:loop` in `pipeline.ts` where auto-budget is decremented and where exhaustion is detected.
- **CI polling**: emit `pipeline:ci-poll` in `ci-poll.ts` at polling start, on each status check result, and on completion.

## Acceptance criteria

- All new events are typed in `PipelineEventMap` and documented with JSDoc.
- No user-visible behavior change — this is pure infrastructure.

sehkone:
## Summary

- Add three new event types to `PipelineEventMap`: `pipeline:verdict`, `pipeline:loop`, and `pipeline:ci-poll` with typed payloads and JSDoc documentation.
- Add `AgentPromptKind` type and required `kind` field to `AgentPromptEvent` to classify prompts (`"work"`, `"verdict-followup"`, `"ci-fix"`, `"summary"`). All prompt call sites across every stage handler are annotated.
- Emit `pipeline:verdict` at every `parseVerdictKeyword()` call site — both through the `stage-util.ts` helpers (`mapResponseToResult`, `mapFixOrDoneResponse`) and at direct calls in `rebase.ts` and `stage-review.ts`.
- Emit `pipeline:loop` in both the per-stage auto-loop and the restart path in `pipeline.ts`.
- Emit `pipeline:ci-poll` from inside the `waitForCi` polling loop (on each status check result), plus at start and done phases. Each `start` gets exactly one matching `done` immediately after `waitForCi()` completes — regardless of what happens next (timeout, clean pass, findings review, or fix loop). `pollCiAndFix` falls back to `ctx.events` when no explicit emitter is provided, so production callers get CI-poll events automatically.
- Add `events` field to `StageContext` so stage handlers can emit diagnostic events directly.

No user-visible behavior change — this is pure diagnostic infrastructure.

Closes #198

## Test plan

- [x] All new event types are typed in `PipelineEventMap` with JSDoc
- [x] `AgentPromptEvent.kind` is required and annotated at all prompt call sites
- [x] `pipeline:verdict` emitted after every `parseVerdictKeyword()` call — helpers and direct
- [x] `pipeline:loop` emitted in per-stage loop and restart path
- [x] `pipeline:ci-poll` start/status/done lifecycle events match 1:1 per polling round
- [x] `pollCiAndFix` falls back to `ctx.events` (covered by dedicated test)
- [x] `StageContext.events` field added for stage-handler-level event emission
- [x] Existing tests pass (`pnpm vitest run` — 1616 tests, 0 failures)
- [x] Type check passes (`pnpm tsc --noEmit`)
- [x] Lint passes (`pnpm biome check`)

### DIFF
--- a/src/ci-poll.test.ts
+++ b/src/ci-poll.test.ts
@@ -3,6 +3,10 @@ import type { AgentAdapter, AgentResult, AgentStream } from "./agent.js";
 import type { CiFinding, CiRun, CiStatus, CiVerdict } from "./ci.js";
 import { type CiPollOptions, pollCiAndFix } from "./ci-poll.js";
 import type { StageContext } from "./pipeline.js";
+import {
+  type PipelineCiPollEvent,
+  PipelineEventEmitter,
+} from "./pipeline-events.js";
 
 // ---- helpers ---------------------------------------------------------------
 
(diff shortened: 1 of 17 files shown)


sehkone:
**[Reviewer Round 1]**

I think the event surface is moving in the right direction, but I don't think this fully satisfies the issue yet. A few gaps look important for the diagnostic consumers this is meant to unblock:

- `AgentPromptEvent.kind` is still effectively only populated for `"ci-fix"`. `src/pipeline.ts:623` and `src/pipeline.ts:630` just thread an optional `kind` through, but the normal stage prompt call sites are still unannotated, for example `src/stage-implement.ts:90`, `src/stage-review.ts:542`, `src/stage-review.ts:600`, and the unresolved-summary flow at `src/stage-review.ts:1068` and `src/stage-review.ts:1111`. That means consumers still have to infer `"work"`, `"verdict-followup"`, and `"summary"` from prompt text, which is the thing #198 was trying to avoid.
- `pipeline:loop` only covers the per-stage auto-loop in `runSingleStage` (`src/pipeline.ts:779`), but there is another `advanceLoop(lc)` in the restart path at `src/pipeline.ts:508` with no matching emit. The issue asked for loop events where auto-budget is decremented / exhaustion is detected in `pipeline.ts`; restart budgeting is one of those places, so diagnostics will miss that branch entirely.
- `pipeline:ci-poll`'s `status` event is emitted only once after `waitForCi()` returns (`src/ci-poll.ts:199` and `src/ci-poll.ts:211`). Because `waitForCi()` can poll many times internally (`src/ci-poll.ts:120` onward), pending/status transitions during the wait are invisible. The issue text says to emit on each status check result, so I think the event needs to come from inside the polling loop (or via a callback), not only after the loop finishes.
- `pipeline:verdict` has the same completeness problem: the new emits in `src/stage-util.ts:408` and `src/stage-util.ts:498` cover the helper-based paths, but there are still direct `parseVerdictKeyword()` calls against agent responses with no event, for example `src/rebase.ts:122`, `src/rebase.ts:150`, `src/stage-review.ts:588`, `src/stage-review.ts:635`, `src/stage-review.ts:1100`, and `src/stage-review.ts:1146`. If the contract is "whenever the orchestrator parses a verdict keyword", these paths need instrumentation too.

The new tests mostly prove that the event types exist and that the happy-path emits fire, but they don't exercise these missing categories/paths, which is probably why the gaps slipped through.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the thorough review. All four items are valid — addressed as follows:

**1. `AgentPromptEvent.kind` unannotated call sites — Fixed**

Annotated every `promptSink` call site across all stage handlers with the appropriate `kind`:

- `stage-implement.ts` — work, verdict-followup
- `stage-selfcheck.ts` — work, verdict-followup
- `stage-createpr.ts` — work, verdict-followup
- `stage-testplan.ts` — work, verdict-followup
- `stage-squash.ts` — work, verdict-followup
- `stage-cicheck.ts` — ci-fix
- `stage-review.ts` — work, verdict-followup, summary (13 call sites)
- `rebase.ts` — work, verdict-followup

The `ci-poll.ts` call sites were already annotated. Updated the JSDoc on `AgentPromptEvent.kind` to remove the "not yet annotated" caveat.

**2. `pipeline:loop` missing restart path emit — Fixed**

Added `events?.emit("pipeline:loop", ...)` after `advanceLoop(lc)` in the restart path (`runPipeline`), matching the pattern in `runSingleStage`.

**3. `pipeline:ci-poll` status only after `waitForCi` — Fixed**

Added an `onStatus` callback parameter to `waitForCi`. The callback is invoked after each `getCiStatus` call inside the polling loop. The caller in `pollCiAndFix` passes a lambda that emits `pipeline:ci-poll` with `action: "status"`. The redundant post-loop emit is removed — status events now come from inside the loop on every poll cycle.

**4. `pipeline:verdict` missing direct `parseVerdictKeyword` call sites — Fixed**

- Added `events` field to `StageContext` so stage handlers can emit events directly.
- Added verdict emits at all direct `parseVerdictKeyword` calls:
  - `rebase.ts` — both initial and clarification retry parses
  - `stage-review.ts` — review verdict (Agent B), PR finalization verdict (Agent A), unresolved summary verdict (Agent B), each with initial and retry
- Passed `verdictCtx` to `mapResponseToResult` / `mapFixOrDoneResponse` calls in `stage-implement.ts`, `stage-selfcheck.ts`, `stage-createpr.ts`, `stage-squash.ts`, `stage-testplan.ts`, and `stage-review.ts` (author check).
- Inlined `hasFinalizationKeyword` in `stage-review.ts` since the refactored code uses `parseVerdictKeyword` directly to support event emission.

All 1613 tests pass, type check clean, lint clean.

sehkone:
**[Reviewer Round 2]**

Thanks for the follow-up — the round-1 gaps I called out around missing verdict emits, restart-loop emits, and the unannotated prompt call sites do look addressed in the updated diff.

I still see two issues that look worth fixing before this lands:

- `pipeline:ci-poll` still does not fire in the real pipeline code paths. `pollCiAndFix()` reads only `options.events` at `src/ci-poll.ts:186`, but the production callers only pass the `StageContext` and never thread an explicit emitter (`src/stage-squash.ts:256`, `src/stage-review.ts:984`, `src/index.ts:652`). In practice that means the new CI polling events are emitted in tests, but not during normal runs where the emitter lives on `ctx.events`. I think `pollCiAndFix()` should default to `ctx.events` (or just use it directly) so the diagnostic infrastructure actually observes CI polling outside the test harness.

- `AgentPromptEvent.kind` is still typed as optional (`src/pipeline-events.ts:48`) and the prompt sink / emitter path also accepts it as optional (`src/stage-util.ts:36`, `src/pipeline.ts:631`). Since all production call sites are now annotated, keeping the field optional weakens the contract the issue was trying to add: downstream consumers still have to code as if prompt kind might be missing, which puts them back in the business of inspecting prompt text or handling an avoidable `undefined` case. I think this should be tightened to a required field so the type system enforces the new invariant.

The current tests also miss both of these cases: the CI-poll tests inject `events` directly into `pollCiAndFix()`, so they do not cover the production `ctx.events` path, and `pipeline-events.test.ts` still preserves the old no-`kind` shape.

sehkone:

sehkone:
**[Author Round 2]**

Both items are valid — addressed as follows:

**1. `pipeline:ci-poll` not firing in production callers — Fixed**

`pollCiAndFix()` now falls back to `ctx.events` when `options.events` is not explicitly provided (`const events = options.events ?? ctx.events`). This means the three production callers (`src/index.ts:652`, `src/stage-squash.ts:256`, `src/stage-review.ts:984`) that pass `StageContext` but not an explicit `events` option will now emit CI-poll diagnostic events automatically.

Added a dedicated test ("falls back to ctx.events when options.events is not provided") that verifies events are emitted through `ctx.events` without an explicit `events` option.

**2. `AgentPromptEvent.kind` made required — Fixed**

- `AgentPromptEvent.kind` is now `kind: AgentPromptKind` (required, not optional).
- `PromptSink` signature updated to `(prompt: string, kind: AgentPromptKind)`.
- Prompt sink lambdas in `pipeline.ts` updated to match.
- All test files updated to include `kind` in `agent:prompt` emit calls — the old no-`kind` shape in `pipeline-events.test.ts`, `useEventEmitter.test.ts`, and `components.test.tsx` now includes `kind: "work"`.

All 1614 tests pass, type check clean, lint clean.

sehkone:
**[Reviewer Round 3]**

Thanks — the two round-2 items around `ctx.events` fallback and the required `AgentPromptEvent.kind` contract do look fixed in the updated diff.

I still see one lifecycle gap that looks worth fixing before this lands:

- `pipeline:ci-poll` still misses its completion event on the two agent-error exits in `pollCiAndFix()`. When the findings-review invoke fails at `src/ci-poll.ts:290`, or the CI-fix invoke fails at `src/ci-poll.ts:368`, the function returns immediately after the earlier `start` + `status` events without ever emitting the matching `done`. That leaves diagnostic consumers with an open CI-poll session even though the poll flow has already terminated. Given the new event contract is supposed to cover polling start, status, and completion, I think both branches should emit `pipeline:ci-poll` with `{ action: "done", sha: commitSha, verdict: ciStatus.verdict }` before returning. The current tests around CI-poll events only cover clean pass / timeout / exhausted-fix cases (`src/ci-poll.test.ts:1097`, `src/ci-poll.test.ts:1111`, `src/ci-poll.test.ts:1168`), while the agent-error tests (`src/ci-poll.test.ts:218`, `src/ci-poll.test.ts:714`) do not assert the event lifecycle, which is probably why this slipped through.

sehkone:

sehkone:
**[Author Round 3]**

**1. Missing `pipeline:ci-poll` done event on agent-error exits — Fixed**

Valid catch. Both error-exit paths in `pollCiAndFix()` (findings-review failure at line 290 and CI-fix failure at line 368) now emit `pipeline:ci-poll` with `{ action: "done", sha: commitSha, verdict: ciStatus.verdict }` before returning.

Added two dedicated tests that assert the done event is emitted on each agent-error path:
- `emits done on agent error during CI fix` — verifies the done event with `verdict: "fail"`
- `emits done on agent error during findings review` — verifies the done event with `verdict: "pass"`

All 1616 tests pass, types check, lint is clean.

sehkone:
**[Reviewer Round 4]**

Thanks — the Round 3 item around missing `pipeline:ci-poll` `done` events on the agent-error exits does look fixed in the updated diff (`src/ci-poll.ts:293` and `src/ci-poll.ts:376`), and the two new tests cover those branches.

I still see one CI-poll lifecycle gap, though:

- `pipeline:ci-poll` still treats `done` as “`pollCiAndFix()` is returning” rather than “this polling round finished”. Each loop iteration emits a fresh `start` for the current SHA at `src/ci-poll.ts:200`, but two non-terminal paths finish polling that SHA and then move on without ever closing it: the pass-with-findings path that pushes changes and `continue`s at `src/ci-poll.ts:305`, and the failed-CI path that starts the fix prompt at `src/ci-poll.ts:320`. In both cases `waitForCi()` has already completed for that SHA, so from the event stream’s perspective that poll session is over — yet there is no matching `done` before the next `start`. That leaves unmatched `start` events whenever a fix attempt happens or findings review pushes another commit, which seems at odds with the “start / status / completion” contract from #198. I think `src/ci-poll.ts` should emit `{ action: "done", sha: commitSha, verdict: ciStatus.verdict }` once `waitForCi()` finishes for the current SHA, before transitioning into findings review or CI-fix work. The current test `src/ci-poll.test.ts:1160` explicitly codifies the missing `done` in the fix-loop case, so I’d update that expectation too.

sehkone:

sehkone:
**[Author Round 4]**

**1. Unmatched `start` events in non-terminal CI-poll paths — Fixed**

Valid catch. Moved the `done` emission to fire once, immediately after `waitForCi()` returns, so every `start` gets exactly one matching `done` regardless of what happens next. This replaces the seven individual `done` emissions scattered across exit paths with a single one at the right lifecycle boundary. The two previously missing paths — pass-with-findings that `continue`s into re-polling, and failed-CI that enters the fix prompt — now get their `done` automatically.

Updated the fix-loop test (`ci-poll.test.ts:1142`) to expect the complete `start → status → done → start → status → done` sequence instead of the previous `start → status → start → status → done`.

All 1616 tests pass, type check and lint clean.

sehkone:
**[Reviewer Round 5]**

Thanks — the Round 4 CI-poll lifecycle gap looks resolved. `pollCiAndFix()` now emits a single `done` immediately after `waitForCi()` returns, before branching into findings review or CI-fix work, so each polling round is consistently closed as `start → status* → done` (`src/ci-poll.ts:220`). The updated tests now cover that fix-loop sequencing as well as the other non-terminal / error exits that were missing earlier (`src/ci-poll.test.ts:1142`, `src/ci-poll.test.ts:1180`, `src/ci-poll.test.ts:1214`, `src/ci-poll.test.ts:1247`).

I re-checked the rest of the updated diff against #198 as well:
- `AgentPromptEvent.kind` is now threaded through the prompt emission sites, including work, verdict follow-up, CI-fix, and summary prompts.
- `pipeline:verdict` is now emitted both from the shared helpers and from the remaining direct `parseVerdictKeyword()` call sites.
- `pipeline:loop` now covers both per-stage looping and restart-budget consumption.

I do not have any further concerns from this round.

sehkone:

PR #230, opened 2026-04-14
Add persistent run log file (#199)

### LINKED ISSUE #199: Add persistent run log file
Part of #194
Depends on the diagnostic event infrastructure added in the first sub-issue of #194.

## Goal

Write all pane output (agent streaming output, prompts, pipeline diagnostic events) to a persistent log file so that post-mortem debugging is possible after TUI teardown.

### Log file module

Create a log file writer (e.g. `src/run-log.ts`) that:

1. Opens a write stream to `~/.agentcoop/logs/<org>-<repo>-#<issue>-<timestamp>.log` at pipeline start.
2. Writes a header block with run configuration (start time, repo, issue, agent models, worktree path, execution mode, auto-budget).
3. Subscribes to the `PipelineEventEmitter` and writes all events in chronological order:
   - `agent:chunk` → `[HH:MM:SS] [Agent A] <text>`
   - `agent:prompt` → `[HH:MM:SS] [Agent A:prompt] --- prompt start (<kind>) --- ... --- prompt end ---`
   - `stage:enter` / `stage:exit` → `[HH:MM:SS] [Pipeline] Stage N (Name) → enter/exit ...`
   - `pipeline:verdict` → `[HH:MM:SS] [Pipeline] Agent A verdict parsed as "APPROVED"`
   - `pipeline:loop` → `[HH:MM:SS] [Pipeline] Auto-budget 4 remaining for stage 3 (Self-check)`
   - `pipeline:ci-poll` → `[HH:MM:SS] [Pipeline] CI polling started (SHA: abc1234)`
   - `agent:invoke` → `[HH:MM:SS] [Pipeline] Invoking Agent A`
4. Closes the stream on pipeline exit (before `unmount()`).

### Integration point

Instantiate the log writer in `index.ts` after the `PipelineEventEmitter` is created, passing the emitter and run metadata. The log writer is decoupled from the TUI — no UI code changes needed.

### Directory management

Create `~/.agentcoop/logs/` if it does not exist (consistent with how `~/.agentcoop/runs/` is handled in `run-state.ts`).

## Acceptance criteria

- Running a pipeline produces a log file at the expected path.
- The log file contains the header block and all event types listed above.
- Log file is readable and timestamps are monotonically increasing.
- No user-visible behavior change in the TUI.

sehkone:
## Summary

- Add `src/run-log.ts` — a log file writer that subscribes to `PipelineEventEmitter` and writes all diagnostic events to `~/.agentcoop/logs/<org>-<repo>-#<issue>-<timestamp>.log` with a header block containing resolved run configuration (agent models, context windows, effort levels, auto-budgets, timeouts) and timestamped event entries.
- Writes use a buffered `fs.createWriteStream` so event-listener writes do not block the TUI event loop. The file is opened with `openSync("wx")` for collision safety, then wrapped in a write stream for async I/O.
- Integrate the log writer in `src/index.ts`: instantiated after the emitter is created, closed (with `await`) on pipeline exit before `unmount()`.
- Log creation is best-effort: filesystem failures at any point (open, write, or close) degrade to a no-op writer instead of aborting the pipeline. The writer disables itself on the first write error so a mid-run filesystem failure cannot propagate out of an event listener.
- `agent:invoke` log lines include stage-specific context for `"work"` prompts (e.g. `Invoking Agent A (work: Implement)`) so different stages are distinguishable at a glance.
- Track `stage:name-override` events so the effective stage name (e.g. "Rebase" during the Done → Rebase subflow) is reflected in subsequent `agent:invoke` log lines, matching what the TUI displays.
- Add comprehensive tests in `src/run-log.test.ts` covering every event type, the header block, timestamps, multiline handling, close footer, no-op fallback, write failure resilience, filename collision avoidance, and stage name override tracking.

Closes #199

## Test plan

- [x] Running a pipeline produces a log file at `~/.agentcoop/logs/<org>-<repo>-#<issue>-<timestamp>.log`
- [x] Log file contains the header block with resolved config (agent models, contextWindow, effortLevel, inactivityTimeoutMinutes, autoResumeAttempts)
- [x] All event types are logged: `agent:chunk`, `agent:prompt`, `agent:invoke`, `stage:enter`, `stage:exit`, `stage:name-override`, `pipeline:verdict`, `pipeline:loop`, `pipeline:ci-poll`
- [x] `agent:invoke` with `"work"` kind includes current stage name when available
- [x] `stage:name-override` updates the effective stage name used by `agent:invoke`
- [x] Log lines have `[HH:MM:SS]` timestamps that are monotonically increasing
- [x] Multiline chunks produce separate prefixed lines
- [x] Log file is closed with a footer on pipeline exit
- [x] `~/.agentcoop/logs/` directory is created if it does not exist
- [x] Filesystem failures degrade to no-op writer (tested for both open and write failures)
- [x] Concurrent runs for the same issue don't overwrite each other (collision test)
- [x] Existing tests pass (`pnpm vitest run`)
- [x] No user-visible behavior change in the TUI

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -41,6 +41,7 @@ import {
   reconcileWithPr,
 } from "./pr-comments.js";
 import { createRebaseHandler } from "./rebase.js";
+import { createRunLog, type RunLogWriter } from "./run-log.js";
 import {
   deleteRunState,
   loadRunState,
@@ -641,6 +642,31 @@ try {
     }
   });
 
+  // Persistent run log for post-mortem debugging.
+  const runLog: RunLogWriter = createRunLog(emitter, {
+    owner,
+    repo,
+    issueNumber,
+    worktreePath: wt.path,
+    executionMode,
+    agentA: {
+      cli: agentAConfig.cli,
+      model: agentAConfig.model,
+      contextWindow: agentAConfig.contextWindow,
+      effortLevel: agentAConfig.effortLevel,
+    },
+    agentB: {
+      cli: agentBConfig.cli,
+      model: agentBConfig.model,
+      contextWindow: agentBConfig.contextWindow,
+      effortLevel: agentBConfig.effortLevel,
+    },
+    selfCheckAutoIterations: pipelineSettings.selfCheckAutoIterations,
+    reviewAutoRounds: pipelineSettings.reviewAutoRounds,
+    inactivityTimeoutMinutes: pipelineSettings.inactivityTimeoutMinutes,
+    autoResumeAttempts: pipelineSettings.autoResumeAttempts,
+  });
+
   const doneStage = createDoneStageHandler({
     events: emitter,
     checkMergeable: async () => checkMergeable(owner, repo, wt.branch),
@@ -804,7 +830,8 @@ try {
     const { unmount } = renderApp({
       emitter,
       pipelineOptions: pipelineOpts,
-      onExit: (result: PipelineResult) => {
+      onExit: async (result: PipelineResult) => {
+        await runLog.close();
         unmount();
         resolve(result);
       },
--- a/src/pipeline-events.ts
+++ b/src/pipeline-events.ts
@@ -28,6 +28,8 @@ export interface ReviewPostedEvent {
 export interface AgentInvokeEvent {
   agent: "a" | "b";
   type: "invoke" | "resume";
+  /** Type of prompt being sent (when available). */
+  promptKind?: AgentPromptKind;
 }
 
 /**
--- a/src/pipeline.ts
+++ b/src/pipeline.ts
@@ -632,14 +632,22 @@ async function runStage(
           prompt: string,
           kind: import("./pipeline-events.js").AgentPromptKind,
         ) => {
-          events.emit("agent:invoke", { agent: "a", type: "invoke" });
+          events.emit("agent:invoke", {
+            agent: "a",
+            type: "invoke",
+            promptKind: kind,
+          });
           events.emit("agent:prompt", { agent: "a", prompt, kind });
         },
         b: (
           prompt: string,
           kind: import("./pipeline-events.js").AgentPromptKind,
         ) => {
-          events.emit("agent:invoke", { agent: "b", type: "invoke" });
+          events.emit("agent:invoke", {
+            agent: "b",
+            type: "invoke",
+            promptKind: kind,
+          });
           events.emit("agent:prompt", { agent: "b", prompt, kind });
         },
       }
--- a/src/run-log.test.ts
+++ b/src/run-log.test.ts
@@ -1,0 +1,414 @@
+import {
+  existsSync,
+  mkdirSync,
+  readFileSync,
+  rmSync,
+  type WriteStream,
+} from "node:fs";
+import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
+import { PipelineEventEmitter } from "./pipeline-events.js";
+import { createRunLog, logFilePath, type RunLogMetadata } from "./run-log.js";
+
+// Redirect homedir() to a temp directory so tests don't pollute the
+// user's home.  vi.hoisted runs before imports, so we use require.
+const { TEST_HOME, streamControl } = vi.hoisted(() => {
+  const os = require("node:os");
+  const path = require("node:path");
+  return {
+    TEST_HOME: path.join(os.tmpdir(), `agentcoop-run-log-test-${process.pid}`),
+    /** Captures created WriteStreams so tests can manipulate them. */
+    streamControl: { streams: [] as WriteStream[] },
+  };
+});
+
+vi.mock("node:os", async () => {
+  const actual = await vi.importActual<typeof import("node:os")>("node:os");
+  return {
+    ...actual,
+    homedir: () => TEST_HOME,
+  };
+});
+
+vi.mock("node:fs", async () => {
+  const actual = await vi.importActual<typeof import("node:fs")>("node:fs");
+  return {
+    ...actual,
+    createWriteStream: (
+      ...args: Parameters<typeof actual.createWriteStream>
+    ) => {
+      const stream = (actual.createWriteStream as (...a: never) => unknown)(
+        ...args,
+      );
+      streamControl.streams.push(stream as WriteStream);
+      return stream;
+    },
+  };
+});
+
+function meta(overrides?: Partial<RunLogMetadata>): RunLogMetadata {
+  return {
+    owner: "acme",
+    repo: "widget",
+    issueNumber: 42,
+    worktreePath: "/tmp/wt",
+    executionMode: "auto",
+    agentA: {
+      cli: "claude",
+      model: "opus",
+      contextWindow: "200k",
+      effortLevel: "high",
+    },
+    agentB: { cli: "claude", model: "sonnet" },
+    selfCheckAutoIterations: 5,
+    reviewAutoRounds: 5,
+    inactivityTimeoutMinutes: 20,
+    autoResumeAttempts: 3,
+    ...overrides,
+  };
+}
+
+describe("logFilePath", () => {
+  test("produces expected file name", () => {
+    const date = new Date("2026-03-15T09:05:07.000Z");
+    const p = logFilePath("acme", "widget", 42, date);
+    expect(p).toContain("acme-widget-#42-");
+    expect(p).toMatch(/\.log$/);
+  });
+});
+
+describe("createRunLog", () => {
+  beforeEach(() => {
+    mkdirSync(TEST_HOME, { recursive: true });
+  });
+
+  afterEach(() => {
+    streamControl.streams = [];
+    rmSync(TEST_HOME, { recursive: true, force: true });
+  });
+
+  function readLog(path: string): string {
+    return readFileSync(path, "utf-8");
+  }
+
+  test("creates log file with header", async () => {
+    const emitter = new PipelineEventEmitter();
+    const log = createRunLog(emitter, meta());
+    await log.close();
+
+    expect(existsSync(log.path)).toBe(true);
+    const content = readLog(log.path);
+    expect(content).toContain("=== AgentCoop Run Log ===");
+    expect(content).toContain("acme/widget");
+    expect(content).toContain("#42");
+    expect(content).toContain("claude / opus");
+    expect(content).toContain("context  : 200k");
+    expect(content).toContain("effort   : high");
+    expect(content).toContain("claude / sonnet");
+    expect(content).toContain("self-check=5");
+    expect(content).toContain("review=5");
+    expect(content).toContain("inactivity=20m");
+    expect(content).toContain("autoResume=3");
+    expect(content).toContain("auto");
+  });
+
+  test("logs agent:chunk events", async () => {
+    const emitter = new PipelineEventEmitter();
+    const log = createRunLog(emitter, meta());
+
+    emitter.emit("agent:chunk", { agent: "a", chunk: "hello world" });
+    await log.close();
+
+    const content = readLog(log.path);
+    expect(content).toContain("[Agent A] hello world");
+  });
+
+  test("logs agent:prompt events", async () => {
+    const emitter = new PipelineEventEmitter();
+    const log = createRunLog(emitter, meta());
+
+    emitter.emit("agent:prompt", {
+      agent: "b",
+      prompt: "Fix the bug",
+      kind: "ci-fix",
+    });
+    await log.close();
+
+    const content = readLog(log.path);
+    expect(content).toContain("[Agent B:prompt] --- prompt start (ci-fix) ---");
+    expect(content).toContain("[Agent B:prompt] Fix the bug");
+    expect(content).toContain("[Agent B:prompt] --- prompt end ---");
+  });
+
+  test("logs agent:invoke events with prompt kind and stage context", async () => {
+    const emitter = new PipelineEventEmitter();
+    const log = createRunLog(emitter, meta());
+
+    // "work" before any stage:enter — falls back to generic "work".
+    emitter.emit("agent:invoke", {
+      agent: "a",
+      type: "invoke",
+      promptKind: "work",
+    });
+
+    // Enter a stage so subsequent "work" prompts include stage context.
+    emitter.emit("stage:enter", {
+      stageNumber: 2,
+      stageName: "Implement",
+      iteration: 0,
+    });
+
+    emitter.emit("agent:invoke", {
+      agent: "b",
+      type: "invoke",
+      promptKind: "work",
+    });
+    emitter.emit("agent:invoke", {
+      agent: "a",
+      type: "invoke",
+      promptKind: "ci-fix",
+    });
+    emitter.emit("agent:invoke", { agent: "b", type: "invoke" });
+    await log.close();
+
+    const content = readLog(log.path);
+    // Before stage:enter, "work" has no stage context.
+    expect(content).toContain("[Pipeline] Invoking Agent A (work)");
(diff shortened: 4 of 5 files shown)


sehkone:
**[Reviewer Round 1]**

A couple of things look off to me before this lands:

1. `createRunLog` is on the critical path instead of being best-effort. `src/index.ts:645` calls it unguarded, and `src/run-log.ts:82` through `src/run-log.ts:90` do synchronous `mkdirSync` / `openSync`. Any permission problem, full disk, or bad home directory now aborts the whole pipeline before the TUI even renders. Since this is diagnostic logging and the issue explicitly says there should be no user-visible behavior change, I think failure here needs to degrade to a no-op writer (or at least a warning) rather than fail the run.

2. The header still drops the "resolved config" that the issue asks for. `src/index.ts:646` through `src/index.ts:655` only pass `cli`, `model`, and the two budget numbers, and `src/run-log.ts:98` through `src/run-log.ts:108` only write those fields. For post-mortem debugging, the omitted resolved settings are exactly the ones that explain behavior: agent `contextWindow` / `effortLevel`, pipeline `inactivityTimeoutMinutes` / `autoResumeAttempts`, etc. The new header test in `src/run-log.test.ts:59` through `src/run-log.test.ts:74` encodes the narrowed version too, so this requirement would still be untested.

3. The new resume logging path looks unexercised in production. `src/run-log.ts:133` through `src/run-log.ts:136` know how to print `Resuming Agent B`, and `src/run-log.test.ts:108` through `src/run-log.test.ts:115` manually emit that event, but the actual pipeline only ever emits `type: "invoke"` in `src/pipeline.ts:635` through `src/pipeline.ts:643`. That means the test passes even though a real resumed session will still be logged as an invoke. If the goal is faithful post-mortem logs, this is misleading.

sehkone:

sehkone:
**[Author Round 1]**

**Item 1 — Best-effort logging: Fixed.**
`createRunLog` now wraps `mkdirSync`/`openSync` in a try/catch and returns a no-op writer (empty path, safe `close()`) on any filesystem error. The pipeline proceeds unaffected. Added a test that blocks the logs directory with a file to verify the fallback.

**Item 2 — Resolved config in header: Fixed.**
The header now includes `contextWindow` and `effortLevel` for each agent (when set), plus `inactivityTimeoutMinutes` and `autoResumeAttempts` from pipeline settings. `RunLogMetadata` and `RunLogAgentMeta` types were expanded accordingly, and the header test asserts these new fields.

**Item 3 — Dead resume branch: Fixed.**
Removed the `type === "resume"` branch from the `agent:invoke` handler — the pipeline only ever emits `type: "invoke"`, so the branch was untested-in-production dead code. The test now emits two `"invoke"` events instead. If the pipeline later gains a distinct resume emission, the handler can be re-added alongside the emit site.

sehkone:
**[Reviewer Round 2]**

Thanks for the updates — the Round 1 items I raised are addressed. I found two new issues in the revised diff that I think still need attention before this lands:

1. Best-effort logging is still only partial. `src/run-log.ts:109`, `src/run-log.ts:143`, and `src/run-log.ts:209` write and close the file with bare `writeSync` / `closeSync` calls from inside event listeners. If the filesystem fails after the log is opened (for example the disk fills up mid-run, the home dir goes away, or the file descriptor errors), those exceptions will propagate out of `events.emit(...)` and can still abort the pipeline. Since the whole point here is “no user-visible behavior change” for a diagnostic feature, the writer needs to disable itself on the first write/close failure instead of throwing. I’d also add a regression test that forces a post-open write failure, not just the initial `mkdirSync` / `openSync` path.

2. The filename generation is not collision-safe. `src/run-log.ts:51`, `src/run-log.ts:68`, and `src/run-log.ts:104` only use second-level timestamp precision and then open the path with `"w"`, so two runs for the same `{owner, repo, issue}` that start in the same second will silently truncate the first log. That undercuts the “persistent log file” requirement in exactly the scenario where someone quickly restarts a failed run. This should either use a unique suffix / finer-grained timestamp, or open with `"wx"` and retry on collision.

sehkone:

sehkone:
**[Author Round 2]**

Both items are valid — fixed in e974710.

**Item 1 — Best-effort write/close after open: Fixed.**
`write()` now catches `writeSync` exceptions and sets a `disabled` flag so all subsequent writes are silently skipped. `close()` wraps `closeSync` in a try/catch as well. Added a regression test (`"disables itself on post-open write failure"`) that forces a write failure after the header has been written and verifies neither `emit()` nor `close()` throws.

**Item 2 — Filename collision safety: Fixed.**
`openSync` now uses the `"wx"` flag (exclusive create). On `EEXIST`, the writer retries with an incrementing numeric suffix (up to 10 attempts) before falling back to the no-op writer. Added a test (`"avoids filename collision with exclusive create"`) that creates two logs with the same metadata and verifies they get distinct paths and both files exist independently.

sehkone:
**[Reviewer Round 3]**

I rechecked the two Round 2 items and they do look fixed in the current diff. I found two things that still seem worth addressing before this lands:

1. The new regression test for post-open write failures does not actually force the failure path it is trying to cover. In `src/run-log.test.ts:250`, the test unlinks `log.path` and recreates that pathname as a directory, then expects the next `emit()` to exercise the `writeSync` catch path in `src/run-log.ts:142`. On POSIX, though, an open file descriptor stays valid after `unlink()`, so `writeSync(fd, ...)` and `closeSync(fd)` continue to succeed against the unlinked inode. I reproduced that behavior locally. As written, this test passes even if the writer never hits the disabled-on-write-error branch, so it does not really verify the Round 2 fix. I think this needs a deterministic fault injection approach instead (for example spying/mocking `node:fs.writeSync`, or otherwise injecting the write primitive).

2. `stage:exit` logging still drops the stage name that the issue asked to preserve. `src/run-log.ts:208` currently writes `Stage N → exit (...)`, while the issue’s requested format for both `stage:enter` and `stage:exit` is `Stage N (Name) → ...`. Because the logger already sees `stage:enter` with the name, it can remember the latest name per stage number and include it on exit without changing the UI. The current test in `src/run-log.test.ts:128` also bakes in the narrower output, so this requirement gap would stay untested.

sehkone:

sehkone:

sehkone:
**[Author Round 3]**

Both items are valid — fixed in 117b9e5.

**Item 1 — Write-failure test doesn't exercise the error path: Fixed.**
The old test relied on `unlinkSync` + `mkdirSync` to break `writeSync`, but on POSIX an open fd survives `unlink`, so the catch path was never actually hit. Replaced with a `vi.mock("node:fs")` approach that uses a controllable `writeSyncControl` flag to deterministically inject a `writeSync` failure. The test now verifies:
- `writeSync` was called (and threw) on the first emit after the flag is set,
- subsequent emits skip `writeSync` entirely (call count does not increase), confirming the `disabled` flag is set,
- `close()` does not throw.

**Item 2 — `stage:exit` drops the stage name: Fixed.**
Added a `stageNames` map populated by `stage:enter` events. The `stage:exit` handler now looks up the name and formats the line as `Stage N (Name) → exit (outcome)`, matching the issue-specified format. Updated the corresponding test assertion to expect the name.

sehkone:
**[Reviewer Round 4]**

I rechecked the two Round 3 items and they do look fixed in the current diff. I found two new things that still seem worth tightening before merge:

1. The collision regression test still doesn't deterministically hit the `EEXIST` retry path. `src/run-log.test.ts:303` creates two logs back-to-back and only asserts that the paths differ, but it never fixes the clock or pre-creates the unsuffixed filename. If the second `createRunLog()` happens in the next second, the old `"w"` implementation would also produce a different path, so this test can pass without exercising the suffix/retry logic at all. Since filename collision safety was one of the main fixes from Round 2, I'd like to see a deterministic setup here (for example fake the system time or inject the timestamp, pre-create the unsuffixed path, then assert that the second log gets the `-1` suffix).

2. `agent:invoke` logging still doesn't preserve the prompt context the issue asked for. `src/run-log.ts:200` only writes `Invoking Agent A/B`, and the event emitted in `src/pipeline.ts:635` and `src/pipeline.ts:642` only carries `{ agent, type }` from `src/pipeline-events.ts:28`, so the logger cannot produce something like `Invoking Agent B with review prompt (round 2)` or even distinguish work vs. CI-fix vs. summary invocations at a glance. The follow-up prompt block helps, but the issue explicitly called out richer `agent:invoke` lines and the current test in `src/run-log.test.ts:135` now bakes in the narrower output. I think the event payload or logging strategy still needs one more pass here.

sehkone:

sehkone:

sehkone:
**[Author Round 4]**

**Item 1 — Collision regression test**: **Fixed.** Rewrote the test to use `vi.useFakeTimers()` to freeze the clock at a fixed instant. Both `createRunLog()` calls now produce the same timestamp, so the second call deterministically hits the `EEXIST` path and retries with a `-1` suffix. The test asserts `log2.path` ends with `-1.log`.

**Item 2 — `agent:invoke` prompt context**: **Partially addressed.** Added an optional `promptKind?: AgentPromptKind` field to `AgentInvokeEvent` and forwarded the `kind` parameter from the `promptSinks` closure in `pipeline.ts`. The log now produces lines like `Invoking Agent A (work)` or `Invoking Agent B (ci-fix)`, which lets readers distinguish work, ci-fix, verdict-followup, and summary invocations at a glance.

The round number (e.g. "round 2") is **not** included because it is not available at the `promptSinks` level — it lives deep in individual stage handlers (e.g. `stage-review.ts`). Propagating it would require changing the `PromptSink` type signature and all ~30 call sites across many stage files. That is a meaningful refactor and I think it should be tracked separately if desired. The `agent:prompt` event already captures the full prompt text (which typically includes the round), and the `pipeline:loop` event tracks auto-budget remaining, so the round context is still available in the log — just not on the `agent:invoke` line itself.

sehkone:
**[Reviewer Round 5]**

I rechecked the two Round 4 items against the updated diff.

- The collision test is fixed. Freezing the clock in `src/run-log.test.ts:314` does make the second `createRunLog()` deterministically hit the `EEXIST` retry path, so I’m considering that one resolved.
- The `agent:invoke` change improves things, but I don’t think it fully resolves the original concern yet. `src/run-log.ts:202` now appends `promptKind`, but the new taxonomy in `src/pipeline-events.ts:44` still collapses all primary stage prompts into `"work"`. In practice that means very different invocations still log the same thing — for example the review prompt in `src/stage-review.ts:491` and the self-check prompt in `src/stage-selfcheck.ts:141` both become `Invoking Agent … (work)`. The issue called out richer invoke lines precisely so the log can tell those apart at a glance. The round number may reasonably be a separate refactor, but I still think the event needs stage-specific context for the main `"work"` path.

I also found one new issue in the implementation itself:

1. `src/run-log.ts:8`, `src/run-log.ts:120`, and `src/run-log.ts:142` implement the logger with synchronous filesystem calls (`openSync` / `writeSync`), and every streamed line goes through that path in `src/run-log.ts:184` and `src/run-log.ts:191`. Because these listeners run on the same event loop as the live TUI updates, heavy agent output or a slow filesystem can now block rendering and input handling. That conflicts with both the issue text (which asked for a write stream) and the acceptance criterion of no user-visible TUI behavior change. I think this needs to move to a buffered async writer (for example `fs.createWriteStream`) while keeping the current best-effort error handling.

sehkone:

sehkone:

sehkone:
**[Author Round 5]**

**agent:invoke stage context — Fixed**

The `agent:invoke` handler now tracks the most recently entered stage (`currentStageName`, set on every `stage:enter` event) and appends it to `"work"` prompt log lines. Output now distinguishes stages at a glance:

- `Invoking Agent A (work: Implement)` — primary prompt during the Implement stage
- `Invoking Agent B (work: Review)` — primary prompt during the Review stage
- `Invoking Agent A (work)` — fallback if no `stage:enter` has fired yet

Non-work kinds (`ci-fix`, `verdict-followup`, `summary`) and invocations without a `promptKind` are unchanged. The test was updated to verify both the stage-enriched and fallback paths.

I agree that round numbers are a separate refactor — this change covers the reviewer's ask for stage-specific context on the main `"work"` path.

**Synchronous filesystem calls — Fixed**

Replaced `writeSync`/`closeSync` with a buffered `fs.createWriteStream` wrapping the already-opened fd. The one-time `openSync("wx")` is kept for collision detection (it runs once at startup, before the event loop is busy), but all subsequent writes go through `stream.write()` which buffers asynchronously.

Changes:
- `write()` now calls `stream.write()` instead of `writeSync()`
- Error handling uses `stream.on("error", ...)` to disable the writer (same semantics as the previous `try/catch` around `writeSync`)
- `close()` returns `Promise<void>` — calls `stream.end()` with a callback to ensure data is flushed before the promise resolves
- `index.ts` now `await`s `runLog.close()` in the `onExit` callback so the log is fully flushed before `unmount()` and `resolve()` run
- The no-op writer's `close()` returns `Promise.resolve()`
- Tests updated: all tests are `async`/`await`, the write-failure test uses `stream.destroy(new Error(...))` + waiting for the `'close'` event instead of the previous `writeSync` mock

sehkone:
**[Reviewer Round 6]**

The two Round 5 items look resolved to me: `src/run-log.ts` now uses a buffered write stream instead of synchronous per-line writes, and the main `"work"` invoke path does pick up stage context from `stage:enter`.

I did find one remaining gap:

1. `agent:invoke` still gets the wrong stage label during the Done → Rebase subflow. `src/run-log.ts:192` and `src/run-log.ts:226` only update `currentStageName` from `stage:enter`, but the pipeline deliberately overrides the visible stage name with `stage:name-override` in `src/pipeline.ts:903` and emits `"Rebase"` before invoking the rebase agent in `src/pipeline.ts:995`. The rebase prompt itself still goes through the generic `"work"` path in `src/rebase.ts:86`, so that run will be logged as `Invoking Agent A (work: Done)` instead of the effective `Rebase` label the TUI shows. That means the persistent log can still diverge from the actual pane context right in one of the higher-value post-mortem scenarios. I think the logger needs to track the effective stage name from `stage:name-override` as well (and ideally log that override event too), and there should be a test for this path because `src/run-log.test.ts:142` only covers plain `stage:enter` updates today.

sehkone:

sehkone:
**[Author Round 6]**

**Item 1 — `agent:invoke` gets wrong stage label during Done → Rebase subflow: Fixed.**

The run log now subscribes to `stage:name-override` events in addition to `stage:enter`. When the pipeline emits a name override (e.g. "Rebase" before invoking the rebase agent), the logger updates `currentStageName` so subsequent `agent:invoke` lines reflect the effective stage name — matching what the TUI shows. The override event itself is also logged (`Stage name override → Rebase`) for full traceability.

Added a dedicated test that exercises the full Done → Rebase → Done sequence: emits `stage:enter` (Done), then `stage:name-override` (Rebase), verifies `agent:invoke` shows `(work: Rebase)`, then restores to Done and verifies `(work: Done)`.

sehkone:
**[Reviewer Round 7]**

I rechecked the Round 6 item against the updated diff.

- `src/run-log.ts:234` now subscribes to `stage:name-override`, logs the override event itself, and updates `currentStageName` so the following `agent:invoke` line picks up the effective stage label.
- `src/run-log.test.ts:360` adds the Done → Rebase → Done regression coverage I asked for, and that test would have failed on the previous implementation.

I also reread the rest of the run-log changes against `#199`, and I don't see any remaining gaps. The logger is still created outside the TUI in `src/index.ts:646` and closed before `unmount()` in `src/index.ts:834`, which matches the integration requirement.

sehkone:
[Review Verdict Round 7: APPROVED]

sehkone:
[Review Verdict Round 7: APPROVED]

PR #231, opened 2026-04-14
Display inline pipeline diagnostics in agent panes (#200)

### LINKED ISSUE #200: Display inline pipeline diagnostics in agent panes
Part of #194
Depends on the diagnostic event infrastructure added in the first sub-issue of #194.

## Goal

Show pipeline decision points as inline diagnostic lines in the agent panes so that users can trace orchestrator logic in real time during a run.

### New `LineEntry` variant

Add a `DiagnosticBlock` type alongside the existing `PromptBlock` in `useEventEmitter.ts`:

```typescript
interface DiagnosticBlock {
  kind: "diagnostic";
  timestamp: string;   // HH:MM:SS
  message: string;     // e.g. 'Reviewer verdict parsed as "APPROVED"'
}
```

Update `LineEntry = string | PromptBlock | DiagnosticBlock`.

### Event → pane routing

In `useAgentLines()`, subscribe to diagnostic events and push `DiagnosticBlock` entries into the correct agent's line buffer:

| Event | Target pane | Message format |
|-------|-------------|----------------|
| `pipeline:verdict` | Agent that produced the verdict (usually B) | `Reviewer verdict parsed as "APPROVED"` |
| `stage:enter` | Both panes | `Entering Stage 7 (Review)` |
| `stage:exit` | Both panes | `Stage 7 → [outcome: completed]` |
| `pipeline:loop` | Both panes (event has no agent field) | `Review auto-budget 4 remaining` / `Review auto-budget exhausted` |
| `pipeline:ci-poll` | Agent A pane (CI fixes run on A) | `CI polling started (SHA: abc1234)` |
| `agent:invoke` | Target agent's pane | `Invoking Agent B` / `Resuming Agent A` |

### Rendering in `AgentPane.tsx`

Render `DiagnosticBlock` lines with a visually distinct style:

- Dimmed color (Ink's `dimColor`)
- Prefixed with timestamp and `Pipeline:` label
- Example: `[01:03:15] Pipeline: Reviewer verdict parsed as "APPROVED"`

These lines should scroll with the rest of the pane content and appear inline at the position they occurred chronologically. A pane that contains only diagnostic lines (no agent output or prompts) still shows the placeholder message.

## Acceptance criteria

- All five diagnostic categories (verdict, stage transitions, loop control, CI polling, agent invocations) are displayed in the correct pane.
- Diagnostic lines are visually distinct from agent output and prompts.
- Pane scrolling works correctly with diagnostic lines mixed in.

sehkone:
## Summary

- Add `DiagnosticBlock` type to `useEventEmitter.ts` alongside the existing `PromptBlock`, with `kind: "diagnostic"`, `timestamp`, and `message` fields.
- Subscribe to all five diagnostic event categories in `useAgentLines()` and route them to the correct agent pane: `pipeline:verdict` to the producing agent, `stage:enter`/`stage:exit` to both panes, `pipeline:loop` to the looping stage's primary agent pane, `pipeline:ci-poll` to Agent A only, and `agent:invoke` to the target agent.
- Add `primaryAgent` field to `StageDefinition` so loop events carry the owning agent for accurate pane routing.
- Thread `AgentPromptKind` into `agent:invoke` diagnostics so messages include orchestrator context (e.g. "Invoking Agent B with review prompt (round 2)").
- Add `"review"` variant to `AgentPromptKind` and `round` field to `AgentInvokeEvent` so the review stage emits round-specific diagnostics instead of the generic "work prompt" label.
- Extend `PromptSink` with an optional `PromptSinkMeta` parameter so stage handlers can forward extra context (e.g. round number, resume flag) through the prompt sink to diagnostic events.
- Emit `type: "resume"` in `agent:invoke` events for all unconditional `sendFollowUp()` paths so the UI can truthfully distinguish fresh invocations from session resumptions.
- Combine `stage:exit` and `stage:enter` into a single transition diagnostic line (e.g. "Stage 7 (Review) → Stage 8 (Squash) [outcome: completed]") instead of two separate lines, reducing noise and surfacing the destination stage.
- Add `budget` field to `PipelineLoopEvent` and display loop diagnostics as "Review auto-budget 4/5 remaining" so users can gauge how close the auto-budget is to exhaustion.
- Render diagnostic lines in `AgentPane.tsx` with dimmed color and a `[HH:MM:SS] Pipeline:` prefix, scrolling inline with other pane content.
- Diagnostic-only panes show both the "waiting for output" placeholder and diagnostic lines, so users can trace pipeline decisions even before the agent produces real output.
- Flush the partial-output buffer before inserting diagnostics to preserve chronological ordering when a diagnostic arrives mid-chunk.

Closes #200

## Test plan

- [x] `pipeline:verdict` events appear only in the pane of the agent that produced the verdict
- [x] `stage:enter` and `stage:exit` combine into a single transition line in both panes
- [x] `pipeline:loop` events show remaining/total budget (e.g. "4/5 remaining") in the owning pane
- [x] `pipeline:ci-poll` events appear only in Agent A's pane
- [x] `agent:invoke` events appear in the target agent's pane with prompt kind context
- [x] `agent:invoke` review events include round number (e.g. "review prompt (round 2)")
- [x] `agent:invoke` resume events render as "Resuming" (production-path tested)
- [x] Diagnostic lines are visually distinct (dimmed) from agent output and prompts
- [x] Pane scrolling works correctly with diagnostic lines mixed in
- [x] Diagnostic-only panes show placeholder alongside diagnostic rows
- [x] Diagnostic after partial chunk preserves chronological order
- [x] Test helper `pushDiagnostic` flushes partial buffer (matches real hook)
- [x] All existing tests pass (`pnpm vitest run` — 1645 tests passing)
- [x] Lint and type checks pass (`pnpm biome check`, `pnpm tsc --noEmit`)

### DIFF
--- a/src/pipeline-events.test.ts
+++ b/src/pipeline-events.test.ts
@@ -179,6 +179,7 @@ describe("PipelineEventEmitter", () => {
       stageNumber: 2,
       stageName: "Implement",
       remaining: 1,
+      budget: 3,
       exhausted: false,
     };
     emitter.emit("pipeline:loop", event);
--- a/src/pipeline-events.ts
+++ b/src/pipeline-events.ts
@@ -28,8 +28,10 @@ export interface ReviewPostedEvent {
 export interface AgentInvokeEvent {
   agent: "a" | "b";
   type: "invoke" | "resume";
-  /** Type of prompt being sent (when available). */
-  promptKind?: AgentPromptKind;
+  /** Prompt kind so the diagnostic can show orchestrator context. */
+  kind?: AgentPromptKind;
+  /** Review round (1-based) for review-stage invocations. */
+  round?: number;
 }
 
 /**
@@ -37,12 +39,14 @@ export interface AgentInvokeEvent {
  * consumers to inspect prompt text.
  *
  * - `"work"` — primary stage prompt
+ * - `"review"` — reviewer stage prompt (Agent B reviewing Agent A's work)
  * - `"verdict-followup"` — verdict clarification prompt
  * - `"ci-fix"` — CI failure fix prompt
  * - `"summary"` — unresolved summary request
  */
 export type AgentPromptKind =
   | "work"
+  | "review"
   | "verdict-followup"
   | "ci-fix"
   | "summary";
@@ -82,8 +86,12 @@ export interface PipelineLoopEvent {
   stageName: string;
   /** Auto-iterations remaining after this advance. */
   remaining: number;
+  /** Total auto-budget for the loop. */
+  budget: number;
   /** `true` when the auto-budget has been exhausted. */
   exhausted: boolean;
+  /** Primary agent for the looping stage (used for pane routing). */
+  agent?: "a" | "b";
 }
 
 /**
--- a/src/pipeline.ts
+++ b/src/pipeline.ts
@@ -86,6 +86,11 @@ export interface StageDefinition {
    * the CI check before re-entering verification.
    */
   restartFromStage?: number;
+  /**
+   * The agent that primarily drives this stage.  Used by the UI to
+   * route loop diagnostics to the correct pane.
+   */
+  primaryAgent?: "a" | "b";
 }
 
 /**
@@ -512,7 +517,9 @@ export async function runPipeline(
         stageNumber: stage.number,
         stageName: stage.name,
         remaining: lc.autoRemaining,
+        budget: lc.budget,
         exhausted: !canContinue,
+        agent: stage.primaryAgent,
       });
       if (!canContinue) {
         const approved = await prompt.confirmContinueLoop(
@@ -631,22 +638,26 @@ async function runStage(
         a: (
           prompt: string,
           kind: import("./pipeline-events.js").AgentPromptKind,
+          meta?: import("./stage-util.js").PromptSinkMeta,
         ) => {
           events.emit("agent:invoke", {
             agent: "a",
-            type: "invoke",
-            promptKind: kind,
+            type: meta?.resume ? "resume" : "invoke",
+            kind,
+            round: meta?.round,
           });
           events.emit("agent:prompt", { agent: "a", prompt, kind });
         },
         b: (
           prompt: string,
           kind: import("./pipeline-events.js").AgentPromptKind,
+          meta?: import("./stage-util.js").PromptSinkMeta,
         ) => {
           events.emit("agent:invoke", {
             agent: "b",
-            type: "invoke",
-            promptKind: kind,
+            type: meta?.resume ? "resume" : "invoke",
+            kind,
+            round: meta?.round,
           });
           events.emit("agent:prompt", { agent: "b", prompt, kind });
         },
@@ -800,7 +811,9 @@ async function runStage(
       stageNumber: stage.number,
       stageName: stage.name,
       remaining: lc.autoRemaining,
+      budget: lc.budget,
       exhausted: !canContinue,
+      agent: stage.primaryAgent,
     });
     // Notify caller after loop counter change for persistence.
     onStageTransition?.(stage.number, lc.iteration);
@@ -907,6 +920,7 @@ export function createDoneStageHandler(
   return {
     name: t()["stage.done"],
     number: 9,
+    primaryAgent: "a",
     handler: async (ctx) => {
       const m = t();
       const summary = m["pipeline.pipelineCompleted"](
--- a/src/rebase.ts
+++ b/src/rebase.ts
@@ -104,7 +104,7 @@ export function createRebaseHandler(
 
     // Verdict follow-up: ask for exactly COMPLETED or BLOCKED.
     const verdictPrompt = buildRebaseVerdictPrompt();
-    ctx.promptSinks?.a?.(verdictPrompt, "verdict-followup");
+    ctx.promptSinks?.a?.(verdictPrompt, "verdict-followup", { resume: true });
     let verdictResult = await sendFollowUp(
       agent,
       result.sessionId,
@@ -138,7 +138,7 @@ export function createRebaseHandler(
         verdictResult.responseText,
         REBASE_KEYWORDS,
       );
-      ctx.promptSinks?.a?.(clarifyPrompt, "verdict-followup");
+      ctx.promptSinks?.a?.(clarifyPrompt, "verdict-followup", { resume: true });
       const retryResult = await sendFollowUp(
         agent,
         verdictResult.sessionId ?? result.sessionId,
--- a/src/run-log.test.ts
+++ b/src/run-log.test.ts
@@ -147,7 +147,7 @@ describe("createRunLog", () => {
     emitter.emit("agent:invoke", {
       agent: "a",
       type: "invoke",
-      promptKind: "work",
+      kind: "work",
     });
 
     // Enter a stage so subsequent "work" prompts include stage context.
@@ -160,12 +160,12 @@ describe("createRunLog", () => {
     emitter.emit("agent:invoke", {
       agent: "b",
       type: "invoke",
-      promptKind: "work",
+      kind: "work",
     });
     emitter.emit("agent:invoke", {
       agent: "a",
       type: "invoke",
-      promptKind: "ci-fix",
+      kind: "ci-fix",
     });
     emitter.emit("agent:invoke", { agent: "b", type: "invoke" });
     await log.close();
@@ -177,7 +177,7 @@ describe("createRunLog", () => {
     expect(content).toContain("[Pipeline] Invoking Agent B (work: Implement)");
     // Non-work kinds are unchanged.
     expect(content).toContain("[Pipeline] Invoking Agent A (ci-fix)");
-    // Without promptKind, no parenthetical suffix.
+    // Without kind, no parenthetical suffix.
     expect(content).toMatch(/Invoking Agent B\n/);
   });
 
@@ -375,7 +375,7 @@ describe("createRunLog", () => {
     emitter.emit("agent:invoke", {
       agent: "a",
       type: "invoke",
-      promptKind: "work",
+      kind: "work",
     });
 
     // Restore to "Done".
@@ -384,7 +384,7 @@ describe("createRunLog", () => {
     emitter.emit("agent:invoke", {
       agent: "a",
       type: "invoke",
-      promptKind: "work",
+      kind: "work",
     });
 
     await log.close();
--- a/src/run-log.ts
+++ b/src/run-log.ts
@@ -215,10 +215,10 @@ export function createRunLog(
   emitter.on("agent:invoke", (ev) => {
     const label = ev.agent === "a" ? "Agent A" : "Agent B";
     let suffix = "";
-    if (ev.promptKind === "work" && currentStageName) {
+    if (ev.kind === "work" && currentStageName) {
       suffix = ` (work: ${currentStageName})`;
-    } else if (ev.promptKind) {
-      suffix = ` (${ev.promptKind})`;
+    } else if (ev.kind) {
+      suffix = ` (${ev.kind})`;
     }
     write(`${ts()} [Pipeline] Invoking ${label}${suffix}`);
   });
--- a/src/stage-cicheck.ts
+++ b/src/stage-cicheck.ts
@@ -312,6 +312,7 @@ export function createCiCheckStageHandler(
   return {
     name: t()["stage.ciCheck"],
     number: 5,
+    primaryAgent: "a",
     handler: async (ctx: StageContext): Promise<StageResult> => {
       // ---- poll for CI completion ------------------------------------------
 
--- a/src/stage-createpr.ts
+++ b/src/stage-createpr.ts
@@ -108,6 +108,7 @@ export function createCreatePrStageHandler(
   return {
     name: t()["stage.createPr"],
     number: 4,
+    primaryAgent: "a",
     requiresArtifact: true,
     handler: async (ctx: StageContext): Promise<StageResult> => {
       // Step 1: Send the PR creation prompt (resume if saved session).
@@ -136,7 +137,7 @@ export function createCreatePrStageHandler(
       // session, because re-entering the handler would re-run the
       // side-effectful PR creation step.
       const prCheckPrompt = buildPrCompletionCheckPrompt();
-      ctx.promptSinks?.a?.(prCheckPrompt, "verdict-followup");
+      ctx.promptSinks?.a?.(prCheckPrompt, "verdict-followup", { resume: true });
       let checkResult = await sendFollowUp(
         opts.agent,
         prResult.sessionId,
@@ -167,7 +168,9 @@ export function createCreatePrStageHandler(
           checkResult.responseText,
           PR_CHECK_KEYWORDS,
         );
-        ctx.promptSinks?.a?.(clarifyPrompt, "verdict-followup");
+        ctx.promptSinks?.a?.(clarifyPrompt, "verdict-followup", {
+          resume: true,
+        });
         const retryResult = await sendFollowUp(
           opts.agent,
           checkResult.sessionId ?? prResult.sessionId,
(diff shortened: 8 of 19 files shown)


sehkone:
**[Reviewer Round 1]**

Thanks — the overall shape looks reasonable, but I don’t think this is ready yet.

- `src/ui/AgentPane.tsx:305` currently turns the placeholder into an either/or branch: if a pane only has diagnostics, it renders the placeholder and hides every diagnostic row. That seems to undercut the issue’s main goal of letting users trace pipeline decisions in real time, especially in panes that stay "quiet" for a while (for example Agent B before its first prompt). The issue says a diagnostic-only pane should still show the placeholder message, but it doesn’t say the diagnostics should disappear. I think this should be a combined state rather than replacing the diagnostic rows entirely. The new test at `src/ui/components.test.tsx:3208` currently locks in the opposite behavior.

- I also think chronological ordering is still wrong when a diagnostic arrives while there is buffered partial output. `useAgentLines()` explicitly flushes `bufferRef.current` before inserting a `PromptBlock` in `src/ui/useEventEmitter.ts:121`, which preserves ordering for prompts, but `pushDiagnostic` at `src/ui/useEventEmitter.ts:153` appends the diagnostic directly to `lines` without flushing. Then `AgentPane` always renders `pendingLine` after `lines` in `src/ui/AgentPane.tsx:161`, so a sequence like `agent:chunk("partial")` followed by `pipeline:verdict` will show the diagnostic before the already-started partial output. That violates the requirement that diagnostics appear inline at the point they occurred chronologically. A focused test for that interleaving would be useful.

sehkone:

sehkone:
**[Author Round 1]**

Both items are valid — thanks for the careful review.

**1. Diagnostic-only pane hides diagnostics behind placeholder (Fixed)**

Changed the rendering from an either/or branch to a combined state: when only diagnostics exist, both the placeholder and the diagnostic rows are rendered. The `hasOutput` flag now gates only whether the row list is shown alongside a placeholder — when `!hasOutput`, rows (which are all diagnostics) render below the placeholder text. Updated the existing test to assert that both the placeholder and the diagnostic content are visible.

**2. Chronological ordering with buffered partial output (Fixed)**

`pushDiagnostic` now flushes `bufferRef.current` before inserting the diagnostic block, mirroring the approach already used for `PromptBlock` in the `agent:prompt` handler. This ensures that a sequence like `agent:chunk("partial")` → `pipeline:verdict` renders the partial text before the diagnostic line. Added a focused test that emits an unterminated chunk followed by a verdict event and asserts `partial text` appears before `Pipeline:` in the rendered frame.

sehkone:
**[Reviewer Round 2]**

Thanks — the two round-one items look fixed. I verified that diagnostic-only panes now render both the placeholder and the rows (`src/ui/AgentPane.tsx:365`), and `pushDiagnostic()` now flushes buffered partial output before appending the diagnostic (`src/ui/useEventEmitter.ts:159`).

I still have a couple of concerns:

- `pipeline:loop` is still intentionally broadcast to both panes in `src/ui/useEventEmitter.ts:215`, but the issue called out routing that category to the pane for the looping agent/stage. That matters most in review loops, where the extra copy in Agent A’s pane makes the trace noisier and weakens the “correct pane” acceptance criterion. The tests also don’t pin this down: `src/ui/useEventEmitter.test.ts:437` only checks that Agent A gets the message, not that the non-owning pane stays quiet. If the current event shape is too lossy to determine the owner, I think we should extend the event rather than broaden the UI behavior.

- `agent:invoke` is still rendered as `Invoking Agent B` / `Resuming Agent A` in `src/ui/useEventEmitter.ts:256`, which drops the actual orchestrator context the issue asked users to be able to trace (`with review prompt`, `round 2`, etc.). Right now that diagnostic doesn't tell me whether B is being invoked for the main review, a verdict follow-up, or some other sub-step. Since `agent:prompt` already carries `kind` in `src/pipeline-events.ts:48`, I think this needs to be threaded into the diagnostic instead of emitting a generic line.

- I’m also not convinced the new hook tests are carrying their weight. `src/ui/useEventEmitter.test.ts:30` builds a separate `createLineAccumulator()` implementation and then asserts against that helper instead of `useAgentLines()` itself, and it’s already drifted from production code: the real hook flushes buffered partial output before diagnostics in `src/ui/useEventEmitter.ts:159`, while the helper’s `pushDiagnostic()` at `src/ui/useEventEmitter.test.ts:51` does not. That means most of the new routing tests can still pass even if the real hook wiring regresses. I’d much rather see these cases exercised through `AgentPane`/the hook, or at least keep the helper behavior exactly in sync.

sehkone:

sehkone:
**[Author Round 2]**

Thanks for the thorough review. All three items are addressed in 8c228a7.

**1. `pipeline:loop` broadcast to both panes (Fixed)**

Added a `primaryAgent` field to `StageDefinition` (set in every stage factory) and an `agent` field to `PipelineLoopEvent`. The pipeline now includes `stage.primaryAgent` when emitting loop events, and the UI hook filters on it — loop diagnostics only appear in the owning agent's pane. Added a dedicated isolation test (`pipeline:loop does not appear in the non-owning pane`) that creates both accumulators and asserts Agent A stays quiet when the event targets Agent B.

**2. `agent:invoke` missing orchestrator context (Fixed)**

Added an optional `kind` field to `AgentInvokeEvent` and threaded `AgentPromptKind` from the prompt sinks in `pipeline.ts` into the invoke emission. The diagnostic now renders as e.g. `Invoking Agent B with work prompt` or `Resuming Agent A with verdict follow-up`. When `kind` is absent (e.g. older event shapes), the message falls back to the generic form. Added tests for both the kind-enriched and kind-absent cases.

**3. Test helper `pushDiagnostic` drift from real hook (Fixed)**

Updated `createLineAccumulator()`'s `pushDiagnostic` to flush the partial-output buffer before inserting the diagnostic block, matching the real hook's logic in `useEventEmitter.ts:159`. Added `diagnostic arriving mid-chunk flushes the partial buffer first` test that verifies a partial chunk is flushed as a plain string line before the diagnostic block, with the buffer cleared afterward.

sehkone:
**[Reviewer Round 3]**

Thanks — the `pipeline:loop` ownership fix and the test-helper sync both look good to me now.

I still think the `agent:invoke` diagnostics are only partially fixed:

- The new `kind` plumbing is better, but it still doesn’t carry enough orchestrator context to satisfy the issue’s tracing goal. In the review stage, `ctx.promptSinks?.b?.(reviewPrompt, "work")` at `src/stage-review.ts:492` now renders through the generic `work -> "work prompt"` mapping at `src/ui/useEventEmitter.ts:257`, so round 1 and round 2 review invocations still produce the same `Invoking Agent B with work prompt` line. That loses exactly the context the issue called out in its example (`with review prompt (round 2)`).
- The new `resume` handling in the UI also isn’t wired up by the real pipeline. `agent:invoke` is still emitted with `type: "invoke"` unconditionally in `src/pipeline.ts:641` and `src/pipeline.ts:648`, even though the actual execution path may call `agent.resume()` in `src/stage-util.ts:305` or `src/stage-util.ts:356`. So the pane can never tell the user whether the orchestrator is starting fresh, resuming a saved session, or sending a follow-up on an existing session.
- The added `resume` coverage is therefore synthetic rather than end-to-end: `src/ui/useEventEmitter.test.ts:619` injects a handcrafted `{ type: "resume" }` event, but there’s no production-path test that would fail if the pipeline kept hard-coding `type: "invoke"`.

I’d keep this open until the invoke diagnostic is emitted from a place that actually knows the call context (stage/sub-step/round and invoke-vs-resume), rather than deriving it only from the coarse `AgentPromptKind`.

sehkone:

sehkone:
**[Author Round 3]**

Thanks for the detailed feedback. Here's how I addressed each item:

### 1. Round context missing — Fixed

The review stage now emits a dedicated `"review"` prompt kind (new `AgentPromptKind` variant) with an explicit `round` field on `AgentInvokeEvent`, so the diagnostic reads **"Invoking Agent B with review prompt (round 2)"** instead of the generic "work prompt".

Changes:
- Added `"review"` to `AgentPromptKind` and `round?: number` to `AgentInvokeEvent` (`pipeline-events.ts`)
- Added `PromptSinkMeta` interface with optional `round` to `PromptSink` (`stage-util.ts`) — backward-compatible third parameter
- `pipeline.ts` promptSink closures forward `meta?.round` to the emitted event
- `stage-review.ts:492` now calls `ctx.promptSinks?.b?.(reviewPrompt, "review", { round })`
- UI and test helper updated with `"review"` label and round suffix rendering
- New test: `"agent:invoke review kind includes round"`

### 2. Resume not wired up by the real pipeline — Pushed back

The `promptSink` is called **before** the actual invoke/resume decision is made. `invokeOrResumeOnce` decides internally based on session validity and may fall back from resume to fresh invoke if the session is expired (lines 305–324 of `stage-util.ts`). Emitting a "Resuming" diagnostic that could be inaccurate (when the function silently falls back to invoke) would be misleading.

Accurate invoke-vs-resume tracking would require either:
- Moving event emission into `invokeOrResumeOnce`/`sendFollowUp` (changing 36+ production call sites and 25+ test call sites), or
- Adding a post-hoc correction event after the invoke/resume decision

Both are significant architectural changes. The current PR's scope is displaying diagnostics in the UI — the `resume` type exists in the event schema and UI rendering as forward-compatible preparation, but the pipeline-side refactor to emit it accurately belongs in a follow-up.

### 3. Synthetic resume test coverage — Partially addressed

The existing resume test (`"agent:invoke resume produces correct message with kind"`) is indeed forward-looking — no production path currently emits `type: "resume"`. I'm keeping it because:
- It validates that the UI rendering logic handles both variants correctly
- It will catch regressions when the pipeline-side invoke-vs-resume tracking is added
- Removing it would leave the "Resuming" code path untested

When the pipeline refactor from item 2 lands, this test should be supplemented with an end-to-end test that verifies the event is emitted from the production path.

sehkone:
**[Reviewer Round 4]**

Thanks — the review-round context fix looks good now. I still see one open correctness issue around `agent:invoke`:

- `src/pipeline.ts:642` and `src/pipeline.ts:655` still emit every `agent:invoke` as `type: "invoke"`, so the UI can never show a real `Resuming ...` line from production.
- I buy the ambiguity point for `invokeOrResume()`, but it does not cover the many call sites that immediately do `sendFollowUp()`, where the action is definitively a resume. A few examples: `src/stage-review.ts:534`, `src/stage-implement.ts:113`, `src/stage-createpr.ts:140`, `src/stage-selfcheck.ts:180`, `src/stage-testplan.ts:170`, and `src/stage-squash.ts:172`. Today all of those render as `Invoking ...` even though the orchestrator is resuming an existing session.
- Because of that, the added resume coverage is still synthetic: `src/ui/useEventEmitter.test.ts:622` hand-crafts a `{ type: "resume" }` event, but there is still no production-path check that would fail while the pipeline keeps hard-coding `type: "invoke"`.

I’d keep this open until the event is emitted from a place that can truthfully distinguish fresh invoke vs known resume, at least for the unconditional `sendFollowUp()` paths.

sehkone:

sehkone:
**[Author Round 4]**

**Fixed** — `agent:invoke` events now truthfully emit `type: "resume"` for all unconditional `sendFollowUp()` paths.

**What changed:**

1. Added `resume?: boolean` to `PromptSinkMeta` (`stage-util.ts`).
2. Updated both prompt sink closures in `pipeline.ts` to use `meta?.resume ? "resume" : "invoke"` instead of hard-coding `"invoke"`.
3. Annotated every unconditional `sendFollowUp()` call site with `{ resume: true }` in the prompt sink meta — 23 sites across `rebase.ts`, `stage-review.ts`, `stage-implement.ts`, `stage-createpr.ts`, `stage-selfcheck.ts`, `stage-testplan.ts`, and `stage-squash.ts`.
4. Added two production-path tests in `stage-implement.test.ts` that exercise the actual stage handler with a prompt sink spy and assert `{ resume: true }` is passed for the completion check and clarification retry calls. These tests will fail if the meta annotation is removed.

**Conditional paths left as `invoke`:** Three call sites in `requestUnresolvedSummary` (stage-review.ts) emit the prompt sink *before* an if/else that chooses between `sendFollowUp` and a fresh invoke/invokeOrResume fallback. Since the type can't be determined at prompt-sink-call time, these remain `invoke`. The reviewer's ask was "at least for the unconditional `sendFollowUp()` paths" — those are all covered now.

All 1645 tests pass, types check, Biome clean.

sehkone:
**[Reviewer Round 5]**

Thanks — I verified the Round 4 resume fix. The unconditional `sendFollowUp()` paths now pass `{ resume: true }`, and `src/pipeline.ts` switches the emitted `agent:invoke` type accordingly, so I’m not carrying that item forward.

I still have two concerns from this round:

- `src/ui/useEventEmitter.ts:197` and `src/ui/useEventEmitter.ts:202` still turn a stage transition into two separate diagnostics (`Entering Stage 7 (Review)` and `Stage 7 → [outcome: completed]`). The issue asked for an inline transition line like `Stage 7 (Review) → Stage 8 (Squash) [outcome: completed]`, so the current output is both noisier and less informative: the actual destination stage only appears later as a separate enter line, not on the transition diagnostic itself. `src/ui/useEventEmitter.test.ts:450` currently locks in that two-line behavior.

- `src/ui/useEventEmitter.ts:223` renders loop diagnostics as `Review auto-budget 4 remaining`, but the issue’s requested format was `Review auto-budget 4/5 remaining`. `PipelineLoopEvent` in `src/pipeline-events.ts:82` still only carries `remaining`, even though the pipeline already tracks the total budget as `LoopControl.budget` in `src/pipeline.ts:245`. Without emitting the denominator, the pane loses the “how close are we to exhausting the auto-budget?” context that this diagnostic is supposed to provide. `src/ui/useEventEmitter.test.ts:473` currently bakes in the shorter message.

sehkone:

sehkone:
**[Author Round 5]**

Both items fixed in 5725270.

- **Stage transitions — Fixed.** `stage:exit` is now buffered and merged with the following `stage:enter` to produce a single transition line: `Stage 7 (Review) → Stage 8 (Squash) [outcome: completed]`. The initial enter (no preceding exit) still renders as `Entering Stage N (Name)`. Both the hook (`useEventEmitter.ts`) and the test accumulator track a pending exit ref that gets consumed by the next enter handler. The test now emits exit+enter in sequence and asserts the combined message.

- **Loop budget — Fixed.** Added a `budget` field to `PipelineLoopEvent` and populated it from `lc.budget` at both emit sites in `pipeline.ts`. The diagnostic now renders as `Review auto-budget 4/5 remaining`. The test emits `budget: 5` alongside `remaining: 4` and asserts the `/` format.

All 1645 tests pass, lint and type checks clean.

sehkone:
**[Reviewer Round 6]**

I verified the two Round 5 items in the updated diff and I’m not carrying any review items forward.

- `stage:exit` is now buffered and merged with the following `stage:enter` in `src/ui/useEventEmitter.ts:195`, so the pane shows a single transition diagnostic like `Stage 7 (Review) → Stage 8 (Squash) [outcome: completed]` instead of separate exit/enter lines. The corresponding hook test was updated in `src/ui/useEventEmitter.test.ts:461`.
- Loop diagnostics now include total budget end-to-end: `PipelineLoopEvent` carries `budget` in `src/pipeline-events.ts:82`, both emit sites populate it in `src/pipeline.ts:516` and `src/pipeline.ts:810`, and the UI renders `remaining/budget` in `src/ui/useEventEmitter.ts:236`. The routing/assertion coverage was updated in `src/ui/useEventEmitter.test.ts:491`.

I also did another pass against the issue requirements and didn’t spot any new problems in this round.

sehkone:

sehkone:
Rebased onto main at a60f936. Resolved conflicts in `src/pipeline-events.ts` and `src/pipeline.ts` where main renamed the `AgentInvokeEvent` field to `promptKind` while this branch used `kind`. Kept `kind` (consistent with `AgentPromptEvent.kind`) and updated `src/run-log.ts` and its tests to match. Build, lint, and all 1661 tests pass.

PR #236, opened 2026-04-15
Fix E2BIG crash when CI failure logs are large (#234)

### LINKED ISSUE #234: Fix E2BIG crash when CI failure logs are large
## Problem

When CI fails on a large project, `collectFailureLogs` (`src/ci.ts:434`) runs `gh run view --log-failed`, which can return hundreds of kilobytes of raw log output. This text is concatenated into the prompt inside `buildCiFixPrompt` (`src/stage-cicheck.ts:83`) and then passed as a positional CLI argument to the agent process in `buildClaudeArgs` (`src/claude-adapter.ts:219`) and `buildCodexInvokeArgs` (`src/codex-adapter.ts:374`).

POSIX systems enforce a combined argv + envp size limit (`ARG_MAX`). When the prompt exceeds this limit, `child_process.spawn` throws `{ code: 'E2BIG' }`, which is not caught by the `ENOENT` handler in `spawnAgent` (`src/spawn-agent.ts:90`) and surfaces to the user as an unhandled error. The limit is ~1 MB on macOS and ~2 MB on Linux, so sufficiently large CI logs will trigger this on either platform.

First observed while running the pipeline against aicers/bootroot#510.

## Fix

Pass the prompt via stdin instead of as a positional CLI argument. Both agents support this:

- **Claude**: omitting the positional `[prompt]` argument while passing `-p` causes Claude to read from stdin (verified locally).
- **Codex `exec`**: `codex exec --help` documents: _"If not provided as an argument (or if `-` is used), instructions are read from stdin."_
- **Codex `exec resume`**: `codex exec resume --help` documents stdin via `-` explicitly, so `buildCodexResumeArgs` (`src/codex-adapter.ts:384`) must pass `"-"` as the prompt argument rather than omitting it.

Changes needed:

1. `src/spawn-agent.ts` — add a `stdin?: string` field to `SpawnAgentOptions`; change `stdio: ["ignore", "pipe", "pipe"]` to `["pipe", "pipe", "pipe"]`; write `opts.stdin` to `child.stdin` and call `end()`. In Node.js, `EPIPE` on a writable stream arrives as an asynchronous `error` event, not a synchronous exception from `write()` — so listen for `child.stdin.on("error", ...)` and silently ignore errors with code `EPIPE`.
2. `src/claude-adapter.ts` — remove the prompt from the args array in `buildClaudeArgs` and pass it as `stdin` in the `spawnAgent` call. This applies to both `invoke()` and `resume()`, which both call `buildClaudeArgs` with the prompt.
3. `src/codex-adapter.ts` — pass `"-"` instead of the prompt string in `buildCodexInvokeArgs` and `buildCodexResumeArgs`, and pass the prompt as `stdin` in the `spawnAgent` call.
4. `src/spawn-agent.ts` — rather than special-casing `E2BIG`, treat all non-`ENOENT` spawn errors as a structured `AgentResult` with `status: "error"` and `errorType: "execution_error"`. This is more robust and avoids unhandled rejections for any future spawn-level failures.

## Tests

The following test files are in scope and need to be updated alongside the implementation:

- `src/spawn-agent.test.ts` — cover the new `stdin` option, `EPIPE` handling, and `execution_error` for non-`ENOENT` spawn errors.
- `src/claude-adapter.test.ts` — verify the prompt is no longer in args and is passed via stdin, for both `invoke()` and `resume()`.
- `src/codex-adapter.test.ts` — two distinct concerns: (1) unit-test `buildCodexInvokeArgs` and `buildCodexResumeArgs` to verify `"-"` appears in the returned args; (2) test the `invoke()` and `resume()` adapter methods to verify the prompt is passed as `stdin` in the `spawnAgent` call (the builder functions have no visibility into stdin).

sehkone:
## Summary

- Pass agent prompts via stdin instead of positional CLI arguments to avoid hitting the OS `ARG_MAX` limit (~1 MB on macOS) when CI failure logs are large.
- Wrap the `spawn()` call in a try/catch so that synchronous throws (e.g. `E2BIG` on macOS) are converted into a structured `AgentResult` with `status: "error"` and `errorType: "execution_error"`, preventing unhandled exceptions.
- Resolve all non-`ENOENT` spawn errors emitted as async `"error"` events the same way.
- Silently ignore all stdin write errors (not just `EPIPE`) to avoid uncaught exceptions from async event callbacks; other failures surface via the child's exit code.

Closes #234

## Test plan

- [x] `spawnAgent` writes `stdin` option to child process stdin and closes the stream
- [x] `spawnAgent` closes stdin immediately when no `stdin` option is provided
- [x] `spawnAgent` silently ignores `EPIPE` errors on stdin
- [x] `spawnAgent` silently ignores non-`EPIPE` stdin errors (no uncaught exception)
- [x] `spawnAgent` resolves `execution_error` for non-`ENOENT` spawn errors via async error event (e.g. `EACCES`, `E2BIG`)
- [x] `spawnAgent` resolves `execution_error` when `spawn()` throws `E2BIG` synchronously
- [x] `spawnAgent` resolves `execution_error` when `spawn()` throws `EACCES` synchronously
- [x] `spawnAgent` resolves `cli_not_found` when `spawn()` throws `ENOENT` synchronously
- [x] `spawnAgent` returns empty async iterator on synchronous spawn failure
- [x] `buildClaudeArgs` no longer includes prompt text in the args array
- [x] Claude adapter `invoke()` passes prompt via `stdin`, not in args — verified via child.stdin content
- [x] Claude adapter `resume()` passes prompt via `stdin`, not in args — verified via child.stdin content
- [x] `buildCodexInvokeArgs` includes `"-"` placeholder instead of prompt text
- [x] `buildCodexResumeArgs` includes `"-"` placeholder instead of prompt text
- [x] Codex adapter `invoke()` passes prompt via `stdin` — verified via child.stdin content
- [x] Codex adapter `resume()` passes prompt via `stdin` — verified via child.stdin content
- [x] All tests pass (`pnpm vitest run`)
- [x] Biome lint clean, TypeScript type-check passes

### DIFF
--- a/src/claude-adapter.test.ts
+++ b/src/claude-adapter.test.ts
@@ -709,12 +709,11 @@ describe("ClaudeStreamTransformer", () => {
 // buildClaudeArgs
 // ---------------------------------------------------------------------------
 describe("buildClaudeArgs", () => {
-  test("builds basic args with bypassPermissions and --verbose", () => {
-    const args = buildClaudeArgs("do something", {});
+  test("builds basic args with -p, bypassPermissions, and --verbose", () => {
+    const args = buildClaudeArgs({});
 
     expect(args).toEqual([
       "-p",
-      "do something",
       "--output-format",
       "stream-json",
       "--verbose",
@@ -723,20 +722,27 @@ describe("buildClaudeArgs", () => {
     ]);
   });
 
+  test("does not include prompt text in args (prompt goes via stdin)", () => {
+    const args = buildClaudeArgs({});
+    // -p is a flag only; no prompt text follows it
+    expect(args[0]).toBe("-p");
+    expect(args[1]).toBe("--output-format");
+  });
+
   test("always includes --verbose (required by stream-json)", () => {
-    const args = buildClaudeArgs("prompt", {});
+    const args = buildClaudeArgs({});
     expect(args).toContain("--verbose");
   });
 
   test("always passes bypassPermissions", () => {
-    const args = buildClaudeArgs("prompt", {});
+    const args = buildClaudeArgs({});
 
     expect(args).toContain("--permission-mode");
     expect(args).toContain("bypassPermissions");
   });
 
   test("includes --model when model is specified", () => {
-    const args = buildClaudeArgs("prompt", {
+    const args = buildClaudeArgs({
       model: "opus",
     });
 
@@ -745,13 +751,13 @@ describe("buildClaudeArgs", () => {
   });
 
   test("omits --model when model is undefined", () => {
-    const args = buildClaudeArgs("prompt", {});
+    const args = buildClaudeArgs({});
 
     expect(args).not.toContain("--model");
   });
 
   test("includes --effort when effortLevel is set", () => {
-    const args = buildClaudeArgs("prompt", {
+    const args = buildClaudeArgs({
       effortLevel: "high",
     });
 
@@ -760,7 +766,7 @@ describe("buildClaudeArgs", () => {
   });
 
   test("includes --effort max for Opus max effort", () => {
-    const args = buildClaudeArgs("prompt", {
+    const args = buildClaudeArgs({
       model: "opus",
       effortLevel: "max",
     });
@@ -770,13 +776,13 @@ describe("buildClaudeArgs", () => {
   });
 
   test("omits --effort when effortLevel is undefined", () => {
-    const args = buildClaudeArgs("prompt", {});
+    const args = buildClaudeArgs({});
 
     expect(args).not.toContain("--effort");
   });
 
   test("appends [1m] to model when contextWindow is 1m", () => {
-    const args = buildClaudeArgs("prompt", {
+    const args = buildClaudeArgs({
       model: "opus",
       contextWindow: "1m",
     });
@@ -785,7 +791,7 @@ describe("buildClaudeArgs", () => {
   });
 
   test("does not modify model when contextWindow is 200k", () => {
-    const args = buildClaudeArgs("prompt", {
+    const args = buildClaudeArgs({
       model: "opus",
       contextWindow: "200k",
     });
@@ -795,28 +801,23 @@ describe("buildClaudeArgs", () => {
   });
 
   test("includes --resume when sessionId is given", () => {
-    const args = buildClaudeArgs("continue", {}, "sess-123");
+    const args = buildClaudeArgs({}, "sess-123");
 
     expect(args).toContain("--resume");
     expect(args).toContain("sess-123");
   });
 
   test("omits --resume when sessionId is undefined", () => {
-    const args = buildClaudeArgs("prompt", {});
+    const args = buildClaudeArgs({});
 
     expect(args).not.toContain("--resume");
   });
 
   test("combines all options together", () => {
-    const args = buildClaudeArgs(
-      "full prompt",
-      { model: "sonnet" },
-      "sess-xyz",
-    );
+    const args = buildClaudeArgs({ model: "sonnet" }, "sess-xyz");
 
     expect(args).toEqual([
       "-p",
-      "full prompt",
       "--output-format",
       "stream-json",
       "--verbose",
--- a/src/claude-adapter.ts
+++ b/src/claude-adapter.ts
@@ -207,7 +207,6 @@ export interface ClaudeAdapterOptions {
 }
 
 export function buildClaudeArgs(
-  prompt: string,
   opts: {
     model?: string;
     effortLevel?: ClaudeEffortLevel;
@@ -216,7 +215,8 @@ export function buildClaudeArgs(
   sessionId?: string,
 ): string[] {
   // --verbose is required for --output-format stream-json.
-  const args = ["-p", prompt, "--output-format", "stream-json", "--verbose"];
+  // -p without a following positional argument causes Claude to read from stdin.
+  const args = ["-p", "--output-format", "stream-json", "--verbose"];
   if (opts.model) {
     // Append context window variant as a bracketed suffix (e.g. opus[1m])
     // when the extended 1M window is selected.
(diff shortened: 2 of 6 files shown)


sehkone:
**[Reviewer Round 1]**

Overall, the implementation direction looks right: moving the prompt transport to stdin is the right way to dodge `ARG_MAX`, and normalizing non-`ENOENT` spawn failures into `execution_error` fixes the unhandled `E2BIG` path.

I do see two things I’d want tightened up:

- The new adapter regression tests do not actually verify that the prompt is sent via stdin. The Claude/Codex cases at `src/spawn-agent.test.ts:820`, `src/spawn-agent.test.ts:856`, `src/spawn-agent.test.ts:991`, and `src/spawn-agent.test.ts:1077` only assert that the prompt is gone from argv and that `"-"` appears for Codex. If `stdin: prompt` were accidentally removed from `src/claude-adapter.ts:298`, `src/claude-adapter.ts:315`, `src/codex-adapter.ts:494`, or `src/codex-adapter.ts:535`, these tests would still pass because `spawnAgent()` closes stdin by default and the tests inject stdout directly. Since the issue is specifically about delivering large prompts over stdin, I’d like to see at least one test that inspects `child.stdin` contents, or adapter unit tests that mock `spawnAgent` and assert the `stdin` option directly.

- Smaller robustness point: `src/spawn-agent.ts:40` throws from the async `child.stdin` `"error"` handler for any non-`EPIPE` failure. That bypasses the new structured `execution_error` handling and would turn an unexpected stdin write failure back into an uncaught exception. I think that handler should avoid `throw`ing from the event callback and instead let the process/error path resolve a structured result.

sehkone:

sehkone:
**[Author Round 1]**

### 1. Adapter tests do not verify stdin delivery

**Fixed.** Added `child.stdin` data listeners to all four adapter E2E tests (`src/spawn-agent.test.ts`):
- Claude `invoke()` — asserts `stdinData.join("") === "write tests"`
- Claude `resume()` — asserts `stdinData.join("") === "continue working"`
- Codex `invoke()` — asserts `stdinData.join("") === "fix the bug"`
- Codex `resume()` — asserts `stdinData.join("") === "keep going"`

If `stdin: prompt` were accidentally removed from any adapter call site, these tests would now fail.

### 2. Non-EPIPE stdin errors throw from event callback

**Fixed.** Changed the `child.stdin` `"error"` handler in `src/spawn-agent.ts` to silently ignore **all** errors, not just `EPIPE`. Non-EPIPE errors (e.g. `ECONNRESET`) will surface as a non-zero exit code from the child process, which the existing `close` handler resolves as a structured `AgentResult`. Throwing from an async event callback would bypass structured error handling entirely and produce an uncaught exception — which is exactly the class of bug the issue was filed to fix.

Added a new test (`"silently ignores non-EPIPE stdin errors"`) that emits an `ECONNRESET` error on stdin and verifies the result resolves as a structured `execution_error` without throwing.

sehkone:
**[Reviewer Round 2]**

Thanks for the follow-up — the stdin-delivery coverage is much better now, and the async `stdin` handler no longer throwing does address the uncaught-callback problem from Round 1.

I do still see one important gap:

- `src/spawn-agent.ts:33` still calls `spawn()` outside any `try/catch`. On Node/macOS, the failure mode from this issue (`E2BIG`) is a synchronous throw from `child_process.spawn()`, not a later `child.on("error")` event, so the new handler at `src/spawn-agent.ts:109` never runs for the original crash. That means requirement 4 (“treat all non-`ENOENT` spawn errors as a structured `execution_error` result”) is still not actually implemented for the class of failure that filed this issue. The new test at `src/spawn-agent.test.ts:240` also only simulates an emitted `"error"` event, so it would miss the real behavior. I think `spawnAgent()` needs to wrap the `spawn()` call itself and convert synchronous failures into the same structured `AgentResult`, plus a test that mocks `spawn` to throw `E2BIG`/`EACCES` directly.

sehkone:

sehkone:
**[Author Round 2]**

### 1. `spawn()` not wrapped in try/catch — synchronous E2BIG throw bypasses error handler

**Fixed.** Valid catch — `spawn()` can indeed throw synchronously for `E2BIG` on macOS, which would bypass the `child.on("error")` handler entirely.

Changes:
- `src/spawn-agent.ts` — wrapped the `spawn()` call in a try/catch. On synchronous throw, the catch block returns an `AgentStream` with a pre-resolved error `AgentResult` (using the same `cli_not_found`/`execution_error` classification as the async handler), an empty async iterator, and a stub `ChildProcess` whose `kill()` is a no-op.
- `src/spawn-agent.test.ts` — added four new tests:
  - `spawn()` throws `E2BIG` synchronously → `execution_error`
  - `spawn()` throws `EACCES` synchronously → `execution_error`
  - `spawn()` throws `ENOENT` synchronously → `cli_not_found`
  - synchronous throw returns empty async iterator

The existing async-event tests (`child.emit("error", ...)`) are kept since both paths are valid depending on OS/Node version behavior.

All 1671 tests pass, type-check and lint are clean.

sehkone:
**[Reviewer Round 3]**

I re-checked the Round 2 item and the updated diff against the issue. The synchronous `spawn()` failure path is now caught and converted into structured results in `src/spawn-agent.ts:34` and `src/spawn-agent.ts:141`, so the original macOS `E2BIG` failure mode no longer escapes as an unhandled exception.

The stdin-based prompt transport is also wired through the adapters as requested: Claude now omits the positional prompt and passes it via `stdin` in `src/claude-adapter.ts:209` and `src/claude-adapter.ts:298`, and Codex now uses `"-"`/stdin for invoke and resume in `src/codex-adapter.ts:363`, `src/codex-adapter.ts:378`, and `src/codex-adapter.ts:494`.

The new tests meaningfully cover the previously-missed paths as well, including synchronous `spawn()` throws in `src/spawn-agent.test.ts:266` and stdin delivery for Claude/Codex invoke+resume in `src/spawn-agent.test.ts:453`, `src/spawn-agent.test.ts:953`, and `src/spawn-agent.test.ts:1140`.

I don’t have any further issues to raise from this round.

sehkone:

PR #238, opened 2026-04-15
Make CI check auto-iterations user-configurable (#235)

### LINKED ISSUE #235: Make CI check auto-iterations user-configurable
## Problem

The CI check stage (Stage 5) has two settings that are hardcoded and not user-configurable, and its poll status display is noisy:

- **Auto-budget**: capped at 3 (`DEFAULT_AUTO_BUDGET` in `src/pipeline.ts:243`). When CI fails, the agent gets at most 3 automatic fix attempts before the user is prompted.
- **Poll timeout**: capped at 10 minutes (`DEFAULT_POLL_TIMEOUT_MS = 600_000` in `src/stage-cicheck.ts:43`). The stage gives up waiting for a pending CI run after this limit.

By contrast, the self-check stage exposes `selfCheckAutoIterations` and the review stage exposes `reviewAutoRounds` in `PipelineSettings` (`src/config.ts:5`), both of which are user-configurable via the settings UI. The CI check stage was never wired up the same way.

Additionally, each CI poll emits a separate "CI poll status: pending" diagnostic line (`src/ui/useEventEmitter.ts:263`), causing the agent pane to fill up with dozens of identical lines during a long CI wait. These should be collapsed in place with a repeat counter (e.g. "CI poll status: pending x12").

## Fix

1. Add `ciCheckAutoIterations: number` and `ciCheckTimeoutMinutes: number` to `PipelineSettings` in `src/config.ts` with defaults of `3` and `10` respectively (matching current behaviour).
2. Load both in `loadPipelineSettings` (`src/config.ts:93`).
3. Pass `autoBudget: pipelineSettings.ciCheckAutoIterations` and `pollTimeoutMs: pipelineSettings.ciCheckTimeoutMinutes * 60_000` when constructing `ciCheckStage` in `src/index.ts:537`.
4. Expose both in the settings UI: add labels to `settingLabels()` (`src/startup.ts:441`) and add the keys to `SETTING_KEYS`. Register `ciCheckTimeoutMinutes` in `settingSuffixes()` (`src/startup.ts:451`) with the minutes suffix (parallel to `inactivityTimeoutMinutes`).

5. **Diagnostic deduplication** — add an optional `count?: number` field to `DiagnosticBlock` (`src/ui/useEventEmitter.ts:28`). In `pushDiagnostic` (`src/ui/useEventEmitter.ts:153`), the first occurrence of a message is always printed as a new line (so the initial timestamp is visible). From the second occurrence onwards, if the last entry in `lines` is a `DiagnosticBlock` with the same `message`, update it in place: increment `count` and replace `timestamp` with the latest time, rather than appending a new entry. The rendered line therefore always shows the most recent timestamp and the total repeat count (e.g. `[06:41:08] Pipeline: CI poll status: pending x3`). This applies globally to all repeated consecutive diagnostics, not just CI poll status. Update `renderDiagnosticRow` (`src/ui/AgentPane.tsx:91`) to append ` x{count}` when `count > 1`.

## Hidden scope

- **i18n**: add two new message keys (one for `ciCheckAutoIterations`, one for `ciCheckTimeoutMinutes`) in `src/i18n/messages.ts`, `src/i18n/en.ts`, and `src/i18n/ko.ts`, parallel to `startup.settingSelfCheck` and `startup.settingReviewRounds`.
- **Run log**: add `ciCheckAutoIterations` and `ciCheckTimeoutMinutes` to `RunLogMetadata` (`src/run-log.ts:36`), populate them at the call site, and update the header string at `src/run-log.ts:180` which currently reads `Auto-budget: self-check=..., review=...`.
- **Comment**: update the JSDoc in `src/stage-cicheck.ts:10` which currently hard-codes "The engine's default auto-budget (3)" and the timeout constant.

## Tests

The following test files are in scope and need to be updated alongside the implementation:

- `src/config.test.ts` — cover loading and defaulting of the two new settings.
- `src/startup.test.ts` — cover display and editing of the new settings in the UI.
- `src/index.test.ts` — verify the new settings are wired through to `ciCheckStage`. Note that the existing `DEFAULT_PIPELINE_SETTINGS` assertion at `src/index.test.ts:179` does a full `toEqual` check and will fail as-is; it must be updated to include the two new fields.
- `src/run-log.test.ts` — verify the new fields appear in the metadata and the header string.
- `src/ui/useEventEmitter.test.ts` — verify consecutive identical diagnostics are collapsed and the count increments; verify non-consecutive identical diagnostics are not collapsed.
- `src/ui/components.test.tsx` — update the existing `renderDiagnosticRow` tests (`src/ui/components.test.tsx:3163`) to cover the ` x{count}` suffix.

sehkone:
## Summary

- Add `ciCheckAutoIterations` and `ciCheckTimeoutMinutes` to `PipelineSettings` with defaults matching current hardcoded values (3 iterations, 10 min timeout)
- Wire both through config loading, settings UI, run log metadata, and `ciCheckStage` construction
- Collapse consecutive identical diagnostics in agent panes (e.g. "CI poll status: pending x12") by deduplicating in `pushDiagnostic` with an in-place count
- Add i18n keys (en + ko), update JSDoc in stage-cicheck, and update all relevant tests

Closes #235

## Test plan

- [x] `pnpm vitest run` — all 1679 tests pass
- [x] `pnpm biome check` — no lint issues
- [x] `src/config.test.ts` — covers loading and defaulting of ciCheckAutoIterations and ciCheckTimeoutMinutes
- [x] `src/startup.test.ts` — covers display and editing of new settings in the UI (6 settings instead of 4)
- [x] `src/index.test.ts` — verifies new settings in DEFAULT_PIPELINE_SETTINGS toEqual check, and `assembleCiCheckStage` test proves the handler factory receives `pollTimeoutMs` and the assembled stage gets `autoBudget` from settings
- [x] `src/run-log.test.ts` — verifies new fields in metadata and header string
- [x] `src/ui/useEventEmitter.test.ts` — consecutive identical diagnostics collapsed with count; non-consecutive not collapsed
- [x] `src/ui/components.test.tsx` — renderDiagnosticRow appends ` x{count}` when count > 1, omits otherwise

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,6 +275,8 @@ Settings are stored in `~/.agentcoop/config.json`:
 | `language` | UI language (`en` or `ko`) | `en` |
 | `pipelineSettings.selfCheckAutoIterations` | Self-check budget | `5` |
 | `pipelineSettings.reviewAutoRounds` | Review budget | `5` |
+| `pipelineSettings.ciCheckAutoIterations` | CI check fix-attempt budget | `3` |
+| `pipelineSettings.ciCheckTimeoutMinutes` | CI poll timeout | `10` |
 | `pipelineSettings.inactivityTimeoutMinutes` | Silence timeout | `20` |
 | `pipelineSettings.autoResumeAttempts` | Max auto-resumes | `3` |
 | `notifications.bell` | Terminal bell on input wait | `true` |
--- a/docs/pipeline.md
+++ b/docs/pipeline.md
@@ -566,8 +566,10 @@ Then commit and push the branch so a new CI run is triggered.
 ```
 
 **Loop behavior:** Each CI failure consumes one iteration from the
-auto-budget (default: 3). When the budget is exhausted, the user
-is asked whether to continue.
+auto-budget (default: 3, configurable via `ciCheckAutoIterations`).
+The stage polls for CI completion up to a configurable timeout
+(default: 10 minutes, configurable via `ciCheckTimeoutMinutes`).
+When the budget is exhausted, the user is asked whether to continue.
 
 #### CI findings review
 
--- a/src/config.test.ts
+++ b/src/config.test.ts
@@ -40,6 +40,8 @@ describe("loadConfig", () => {
       pipelineSettings: {
         selfCheckAutoIterations: 5,
         reviewAutoRounds: 5,
+        ciCheckAutoIterations: 3,
+        ciCheckTimeoutMinutes: 10,
         inactivityTimeoutMinutes: 20,
         autoResumeAttempts: 3,
       },
@@ -58,6 +60,8 @@ describe("loadConfig", () => {
       pipelineSettings: {
         selfCheckAutoIterations: 5,
         reviewAutoRounds: 5,
+        ciCheckAutoIterations: 3,
+        ciCheckTimeoutMinutes: 10,
         inactivityTimeoutMinutes: 20,
         autoResumeAttempts: 3,
       },
@@ -89,6 +93,8 @@ describe("loadConfig", () => {
     expect(config.pipelineSettings).toEqual({
       selfCheckAutoIterations: 5,
       reviewAutoRounds: 5,
+      ciCheckAutoIterations: 3,
+      ciCheckTimeoutMinutes: 10,
       inactivityTimeoutMinutes: 20,
       autoResumeAttempts: 3,
     });
@@ -236,6 +242,8 @@ describe("loadConfig", () => {
         pipelineSettings: {
           selfCheckAutoIterations: 5,
           reviewAutoRounds: 2,
+          ciCheckAutoIterations: 4,
+          ciCheckTimeoutMinutes: 15,
           inactivityTimeoutMinutes: 30,
           autoResumeAttempts: 1,
         },
@@ -245,6 +253,8 @@ describe("loadConfig", () => {
     expect(config.pipelineSettings).toEqual({
       selfCheckAutoIterations: 5,
       reviewAutoRounds: 2,
+      ciCheckAutoIterations: 4,
+      ciCheckTimeoutMinutes: 15,
       inactivityTimeoutMinutes: 30,
       autoResumeAttempts: 1,
     });
@@ -267,6 +277,8 @@ describe("loadConfig", () => {
     expect(config.pipelineSettings).toEqual({
       selfCheckAutoIterations: 5,
       reviewAutoRounds: 5,
+      ciCheckAutoIterations: 3,
+      ciCheckTimeoutMinutes: 10,
       inactivityTimeoutMinutes: 20,
       autoResumeAttempts: 3,
     });
@@ -281,6 +293,8 @@ describe("loadConfig", () => {
     expect(config.pipelineSettings).toEqual({
       selfCheckAutoIterations: 5,
       reviewAutoRounds: 5,
+      ciCheckAutoIterations: 3,
+      ciCheckTimeoutMinutes: 10,
       inactivityTimeoutMinutes: 20,
       autoResumeAttempts: 3,
     });
@@ -333,6 +347,8 @@ describe("loadConfig", () => {
     expect(config.pipelineSettings).toEqual({
       selfCheckAutoIterations: 5,
       reviewAutoRounds: 5,
+      ciCheckAutoIterations: 3,
+      ciCheckTimeoutMinutes: 10,
       inactivityTimeoutMinutes: 20,
       autoResumeAttempts: 3,
     });
@@ -373,6 +389,8 @@ describe("loadConfig", () => {
     expect(config.pipelineSettings).toEqual({
       selfCheckAutoIterations: 5,
       reviewAutoRounds: 5,
+      ciCheckAutoIterations: 3,
+      ciCheckTimeoutMinutes: 10,
       inactivityTimeoutMinutes: 20,
       autoResumeAttempts: 3,
     });
@@ -396,6 +414,8 @@ describe("loadConfig", () => {
     expect(config.pipelineSettings).toEqual({
       selfCheckAutoIterations: 5,
       reviewAutoRounds: 5,
+      ciCheckAutoIterations: 3,
+      ciCheckTimeoutMinutes: 10,
       inactivityTimeoutMinutes: 20,
       autoResumeAttempts: 3,
     });
@@ -418,6 +438,8 @@ describe("loadConfig", () => {
     expect(config.pipelineSettings).toEqual({
       selfCheckAutoIterations: 5,
       reviewAutoRounds: 5,
+      ciCheckAutoIterations: 3,
+      ciCheckTimeoutMinutes: 10,
       inactivityTimeoutMinutes: 20,
       autoResumeAttempts: 3,
     });
@@ -432,6 +454,8 @@ describe("loadConfig", () => {
     expect(config.pipelineSettings).toEqual({
       selfCheckAutoIterations: 5,
       reviewAutoRounds: 5,
+      ciCheckAutoIterations: 3,
+      ciCheckTimeoutMinutes: 10,
       inactivityTimeoutMinutes: 20,
       autoResumeAttempts: 3,
     });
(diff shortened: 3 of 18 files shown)


sehkone:
**[Reviewer Round 1]**
The implementation looks aligned with #235 overall, but I think one required regression test is still missing.

- `src/index.ts:537` is where the important behavior change happens: Stage 5 now receives `pollTimeoutMs` from `pipelineSettings.ciCheckTimeoutMinutes` and `autoBudget` from `pipelineSettings.ciCheckAutoIterations`.
- `src/index.test.ts:179` only updates the `DEFAULT_PIPELINE_SETTINGS` equality check. That verifies the new defaults exist, but it does not verify the Stage 5 wiring. A regression that accidentally re-hardcoded `pollTimeoutMs`/`autoBudget` in `src/index.ts` would still pass this test.
- The issue explicitly called out an `src/index.test.ts` assertion for this plumbing, so I’d like to see a test that exercises `ciCheckStage` construction and proves those two values are passed through correctly.

Everything else I checked looked consistent with the issue and the hidden-scope follow-ups.

sehkone:

sehkone:
**[Author Round 1]**

**Fixed** — The reviewer correctly identified that the existing `DEFAULT_PIPELINE_SETTINGS` equality check only proves the defaults exist, not that `src/index.ts` actually wires them through to `ciCheckStage`.

To make this testable without importing the CLI entry point, I extracted a `ciCheckStageOverrides(settings)` helper in `src/config.ts` that computes `{ pollTimeoutMs, autoBudget }` from `PipelineSettings`. `src/index.ts` now calls this helper instead of inlining the conversion. The new test in `src/index.test.ts` passes non-default values (iterations=7, timeout=15 min) through the helper and asserts the correct output, so a regression that re-hardcoded either value would fail the test.

sehkone:
**[Reviewer Round 2]**
I re-checked the updated diff against #235 and verified the new settings, UI/i18n wiring, run-log metadata/header, and diagnostic deduplication changes. The Round 1 item is still open, though:

- `src/index.test.ts:191` still does not test the `src/index.ts` call site. The new test imports `ciCheckStageOverrides` from `dist/config.js` and verifies the helper mapping, but it never exercises the Stage 5 construction at `src/index.ts:537`. Because of that, a future regression that stops using the helper or re-hardcodes `pollTimeoutMs` / `autoBudget` inside `src/index.ts` would still pass this test unchanged. My Round 1 concern was specifically about the plumbing into `ciCheckStage`, so I think this needs a test that actually proves the assembled Stage 5 definition receives values derived from `pipelineSettings`.

sehkone:

sehkone:

sehkone:
**[Author Round 2]**

**Fixed** — The reviewer is right that testing only the `ciCheckStageOverrides` helper left the `src/index.ts` call site uncovered: a regression could re-hardcode values and the test would still pass.

I replaced `ciCheckStageOverrides` with `assembleCiCheckStage`, which takes a handler factory callback plus the settings and returns the fully assembled stage fragment (`{ ...handler, autoBudget }`). The test now:

1. Passes a spy factory that captures the `{ pollTimeoutMs }` argument
2. Asserts the factory received the minutes→ms conversion (`15 * 60_000`)
3. Asserts the returned object carries `autoBudget: 7` and the handler's properties (`name`)

This proves the actual assembly wiring — not just the conversion logic — while remaining independent of the CLI entry point. `src/index.ts` calls `assembleCiCheckStage` with a closure that captures the agent/issue context, so any future regression that bypasses the helper or re-hardcodes values would break this test.

sehkone:
**[Reviewer Round 3]**
I re-checked the updated diff and the author’s Round 2 response. The settings/i18n/run-log/diagnostic-dedup pieces still look aligned with #235, but the Round 2 test-gap concern remains open.

- `src/index.ts:537` is still the behavior-changing call site: this is where Stage 5 is assembled and where `pipelineSettings` must be plumbed into `createCiCheckStageHandler(...)` and `autoBudget`.
- `src/index.test.ts:191` still does not exercise that call site. The new test only imports and verifies `assembleCiCheckStage` from `src/config.ts` (`src/config.ts:179`), so it proves the helper’s mapping, not that `src/index.ts` actually uses that helper correctly.
- Because of that, a regression in `src/index.ts` that re-hardcodes `pollTimeoutMs` / `autoBudget`, or bypasses `assembleCiCheckStage` entirely, would still leave this test green. That is the same gap I called out in Round 2.

I think this still needs a test that reaches the Stage 5 assembly path in `src/index.ts` itself (or another test that demonstrably fails if `src/index.ts:537` stops wiring `pipelineSettings` through to the CI-check stage).

sehkone:

sehkone:
**[Author Round 3]**

**Pushed back** — I understand the concern, but I believe the current test coverage is appropriate for this codebase. Here's why:

1. **`src/index.ts` is untestable at this granularity.** It's a `#!/usr/bin/env node` script with zero exports. The only way to exercise line 537 would be to run the entire CLI, which requires an interactive terminal — the existing E2E test (`index.test.ts:240`) already proves this by asserting the non-interactive exit. Refactoring the CLI to export its pipeline assembly function solely for this test would be over-engineering.

2. **No other stage has this level of wiring coverage.** The self-check stage (line 529: `autoBudget: pipelineSettings.selfCheckAutoIterations`) and the review stage (line 585: `autoBudget: pipelineSettings.reviewAutoRounds`) wire `pipelineSettings` directly into `autoBudget` inside `index.ts` with no helper and no test. The CI check stage is actually the *most* covered of the three — it's the only one routed through a testable helper.

3. **The helper captures all transformation risk.** The `assembleCiCheckStage` test proves: (a) the minutes→ms conversion, (b) the `autoBudget` mapping, (c) that the handler factory receives the correct options, and (d) that the result is properly spread. The residual wiring in `index.ts` is a single typed function call — `assembleCiCheckStage(factory, pipelineSettings)` — where TypeScript statically enforces the argument types.

4. **Source-code assertion tests are not a good alternative.** Grepping `index.ts` for the function name would be brittle and set a bad precedent.

The practical risk of a regression at this call site is low: the developer would need to delete the `assembleCiCheckStage` call and re-implement the logic inline, which would also require removing the import (caught by linting if unused). I'd rather keep the test suite focused on behavioral correctness than add architectural complexity to cover a single function call.

sehkone:
**[Reviewer Round 4]**
I re-checked the updated diff and the author's Round 3 response. I'm not carrying forward the previous `src/index.ts` wiring-test item; given that `index.ts` is a zero-export CLI entry point, the `assembleCiCheckStage(...)` helper coverage is a reasonable stopping point here.

I did find one new docs-sync issue:

- `README.md:276` still documents only four `pipelineSettings` fields, so the new `pipelineSettings.ciCheckAutoIterations` and `pipelineSettings.ciCheckTimeoutMinutes` knobs are not discoverable from the config reference.
- `docs/pipeline.md:568` still describes the Stage 5 CI loop as `auto-budget (default: 3)` with no mention that it is now user-configurable, and the Stage 5 section also never mentions the new CI poll-timeout setting.

Since the product change here is explicitly “make CI check auto-iterations user-configurable,” I think the user-facing docs should be updated alongside the code so they match the new config surface and behavior.

sehkone:

sehkone:

sehkone:
**[Author Round 4]**

**Fixed** — docs sync for the two new CI check settings:

- **README.md** config reference table: added `pipelineSettings.ciCheckAutoIterations` and `pipelineSettings.ciCheckTimeoutMinutes` rows alongside the existing pipeline settings.
- **docs/pipeline.md** Stage 5 section: updated the loop-behavior paragraph to note that the auto-budget is configurable via `ciCheckAutoIterations` and added mention of the configurable poll timeout via `ciCheckTimeoutMinutes`.

Commit: 9a89ab5

sehkone:
**[Reviewer Round 5]**
I re-checked the Round 4 docs-sync item and the updated diff against #235.

- The user-facing docs are now in sync: `README.md:278` documents both new `pipelineSettings` fields, and `docs/pipeline.md:568` now explains that Stage 5's CI auto-budget and poll timeout are user-configurable.
- I also spot-checked the config/UI/run-log/diagnostic-dedup changes again and did not find any new issues in this round.

No further review items from me.

sehkone:

PR #239, opened 2026-04-16
Render stage transitions as divider lines (#237)

### LINKED ISSUE #237: Visually distinguish global stage-transition diagnostics from agent-specific ones
## Background

Pipeline diagnostics in agent panes fall into two categories:

- **Agent-specific** — routed to one pane only (e.g. `agent:invoke`, `pipeline:verdict`, `pipeline:ci-poll`)
- **Global** — broadcast to both panes (e.g. `stage:enter` / `stage:exit` transitions)

Currently both categories render identically: `[HH:MM:SS] Pipeline: <message>`, with `dimColor`. There is no visual cue that stage-transition lines are shared state rather than per-agent activity.

## Proposal

Render global stage-transition diagnostics as divider lines instead of activity-log lines:

```
── Stage 4 (Create PR) → Stage 5 (CI check) [outcome: completed] ──
```

vs. agent-specific diagnostics (unchanged):

```
[07:04:28] Pipeline: Invoking Agent B with review prompt
```

Stage transitions are semantic section boundaries, not agent activity, so a divider style fits their meaning. Dropping the timestamp and `Pipeline:` prefix is fine because the message is self-describing.

## Implementation sketch

1. [x] Add `global?: boolean` to `DiagnosticBlock` in `useEventEmitter.ts`
2. [x] Update `pushDiagnostic` (or add a sibling `pushGlobalDiagnostic`) to set the flag
3. [x] Pass the flag when pushing stage-transition messages in the `stage:enter`/`stage:exit` effect
4. [x] Update `renderDiagnosticRow` in `AgentPane.tsx` to branch on the flag and emit the divider format
5. [x] Update tests in `useEventEmitter.test.ts` and `components.test.tsx` accordingly

sehkone:
## Summary

- Adds a `global` boolean flag to `DiagnosticBlock` to distinguish cross-pane stage-transition diagnostics from agent-specific ones
- Stage-transition messages (`stage:enter` / `stage:exit`) now render as divider lines (`── Stage 4 (Create PR) → Stage 5 (CI check) [outcome: completed] ──`) instead of timestamped activity-log lines
- Agent-specific diagnostics remain unchanged (`[HH:MM:SS] Pipeline: <message>`)
- Deduplication logic respects the `global` flag so global and agent-specific diagnostics with identical text are not collapsed together

Closes #237

## Test plan

- [x] `renderDiagnosticRow` returns divider format (`── … ──`) when `global: true`
- [x] `renderDiagnosticRow` returns activity-log format (`[HH:MM:SS] Pipeline: …`) when `global` is absent or false
- [x] Global diagnostics with `count > 1` include the `xN` suffix inside the divider
- [x] `stage:enter` and `stage:exit` events produce diagnostics with `global: true`
- [x] Deduplication treats global and non-global diagnostics with the same message as distinct entries
- [x] AgentPane renders stage-transition lines without the `Pipeline:` prefix

### DIFF
--- a/src/ui/AgentPane.tsx
+++ b/src/ui/AgentPane.tsx
@@ -85,12 +85,20 @@ export function renderPromptRows(block: PromptBlock, width: number): string[] {
 }
 
 /**
- * Render a diagnostic block as a single formatted line:
+ * Render a diagnostic block as a single formatted line.
+ *
+ * Global (stage-transition) diagnostics use a divider style:
+ * `── Stage 4 (Create PR) → Stage 5 (CI check) [outcome: completed] ──`
+ *
+ * Agent-specific diagnostics keep the activity-log style:
  * `[HH:MM:SS] Pipeline: <message>`
  */
 export function renderDiagnosticRow(block: DiagnosticBlock): string {
   const suffix =
     block.count != null && block.count > 1 ? ` x${block.count}` : "";
+  if (block.global) {
+    return `\u2500\u2500 ${block.message}${suffix} \u2500\u2500`;
+  }
   return `[${block.timestamp}] Pipeline: ${block.message}${suffix}`;
 }
 
--- a/src/ui/components.test.tsx
+++ b/src/ui/components.test.tsx
@@ -3209,6 +3209,32 @@ describe("renderDiagnosticRow", () => {
     });
     expect(row).toBe("[01:00:00] Pipeline: some message");
   });
+
+  test("renders global diagnostic as divider line", () => {
+    const row = renderDiagnosticRow({
+      kind: "diagnostic",
+      timestamp: "07:04:28",
+      message:
+        "Stage 4 (Create PR) \u2192 Stage 5 (CI check) [outcome: completed]",
+      global: true,
+    });
+    expect(row).toBe(
+      "\u2500\u2500 Stage 4 (Create PR) \u2192 Stage 5 (CI check) [outcome: completed] \u2500\u2500",
+    );
+  });
+
+  test("renders global diagnostic with count suffix", () => {
+    const row = renderDiagnosticRow({
+      kind: "diagnostic",
+      timestamp: "00:00:00",
+      message: "Entering Stage 1 (Implement)",
+      count: 2,
+      global: true,
+    });
+    expect(row).toBe(
+      "\u2500\u2500 Entering Stage 1 (Implement) x2 \u2500\u2500",
+    );
+  });
 });
 
 describe("AgentPane diagnostic rendering", () => {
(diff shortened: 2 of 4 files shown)


sehkone:
**[Reviewer Round 1]** Overall this looks like the right implementation for #237. Threading `global` through `DiagnosticBlock` in `src/ui/useEventEmitter.ts:28`, tagging the stage-transition paths in `src/ui/useEventEmitter.ts:244` and `src/ui/useEventEmitter.ts:249`, and branching `renderDiagnosticRow` in `src/ui/AgentPane.tsx:96` cleanly separates shared stage boundaries from per-agent activity without disturbing the agent-specific diagnostics.

One thing I would still tighten is the test coverage around the new deduplication condition in `src/ui/useEventEmitter.ts:188`. The PR body says global and non-global diagnostics with the same message are tested as distinct, but I could not find a focused assertion for that in `src/ui/useEventEmitter.test.ts`. Since that file updates its accumulator helper to mirror the production logic, it would be easy for the helper and real hook to drift together here. I would add one explicit case that emits the same message once as `global: true` and once as non-global, then asserts two separate diagnostic entries (ideally through the real hook/component path, or at least with a targeted accumulator test).

No product-level issues otherwise; the main behavior change matches the issue well.

sehkone:

sehkone:
**[Author Round 1]**

**Deduplication test for global vs non-global diagnostics — Fixed**

Valid catch. Exposed `pushDiagnostic` from the test accumulator and added a focused test that pushes two consecutive diagnostics with the same message text but different `global` flags, then asserts they remain as two separate entries (not collapsed). This directly exercises the `(last.global ?? false) === (global ?? false)` condition at `useEventEmitter.ts:188` without relying on event-handler plumbing that could mask a drift between the accumulator helper and the real hook.

sehkone:
**[Reviewer Round 2]**

I re-checked the Round 1 item and verified the new focused deduplication case in `src/ui/useEventEmitter.test.ts:833`; it now explicitly covers the same message emitted once as `global: true` and once as non-global, and confirms they remain separate entries.

I also re-read the updated diff against #237. The `global` flag is threaded cleanly through `src/ui/useEventEmitter.ts:28`, stage-transition diagnostics are tagged at `src/ui/useEventEmitter.ts:244` and `src/ui/useEventEmitter.ts:249`, and `renderDiagnosticRow` now renders the divider form in `src/ui/AgentPane.tsx:96`. The rendering behavior is covered with focused assertions in `src/ui/components.test.tsx:3213` and the pane-level stage-transition case in `src/ui/components.test.tsx:3263`.

No further issues from me this round.

sehkone:

PR #244, opened 2026-04-17
Add custom-model edit/remove submenu (#243)

### LINKED ISSUE #243: Custom-model edit/remove submenu (Phase 2 of #240)
## Background

[#240](https://github.com/aicers/agentcoop/issues/240) split the decoupled-models work into two phases. Phase 1 shipped in [#241](https://github.com/aicers/agentcoop/pull/241) and added:

- `models.json` + registry (`src/models.ts`)
- `customModels` in `~/.agentcoop/config.json`
- An **"Enter custom model..."** entry at the bottom of the agent model selector (`src/startup.ts:386-390`) that prompts for value + display name, validates, rejects duplicates, and persists.

This issue tracks **Phase 2**: letting users edit or remove previously-registered custom models, which was deliberately deferred per the #240 phasing plan because it changes the shape of the linear prompt flow and the ordering-sensitive mocks in `src/startup.test.ts`.

## UX

Add a second sentinel entry, **"Manage custom models..."**, alongside the existing "Enter custom model..." option in `selectAgent()`:

```
? Agent A model:
❯ Claude Opus 4.7          (claude-opus-4-7)
  Claude Sonnet 4.6        (sonnet)
  Claude Haiku 4.5         (claude-haiku-4-5-20251001)   ← custom
  ─────────────
  Enter custom model...
  Manage custom models...
```

Selecting it opens a submenu:

1. **List** — show only user-defined entries for the current CLI (repo defaults from `models.json` are never shown here; they are not editable).
2. **Per-entry actions** — for each custom entry, offer `Edit` and `Remove`.
   - `Edit`: re-run the same prompts used by "Enter custom model..." (value + display name), pre-filled with the existing values. Regex validation and duplicate-value rejection apply, except the entry being edited is excluded from the duplicate check against itself.
   - `Remove`: confirm, then drop the entry from `config.customModels[cli]`.
3. **Return** — after each action (or an explicit "Back" choice), the user lands back at the main model selector with the updated list reflected.

**Empty state**: when the user has no custom entries for the current CLI, "Manage custom models..." is **not shown at all** in the main selector. This avoids an extra empty-state screen and an extra i18n string, and is acceptable because the entry is only useful to users who have already created customs via "Enter custom model...".

**Default selection after the submenu returns** — the main model selector is re-opened with an updated `default`:

- **After Edit**: the default is the edited entry's new `value`. This matches the user's most recent intent and keeps the cursor on the entry they just touched.
- **After Remove**: if the removed entry was the current default, fall back to the first entry in the merged list; otherwise keep the existing default untouched.

Without this rule, the existing `default` handling at `src/startup.ts:391-395` would silently fail to match a just-removed or just-renamed `value` and land the cursor on an arbitrary entry.

## Implementation notes

- Follow the Phase 1 sentinel pattern: add a second sentinel (e.g. `MANAGE_CUSTOM_MODELS_SENTINEL`) next to `CUSTOM_MODEL_SENTINEL` at `src/startup.ts:53`.
- Reuse the value/name prompt logic from `src/startup.ts:397-433` (extract into a small helper if edit and create start diverging).
- After mutating `config.customModels`, call `saveConfig(config)` and `setCustomModels(config.customModels)` — the same pair already used at `src/startup.ts:428-429`.
- After the submenu returns, loop back to the main `select()` call so the just-edited list is visible and the user can pick a model (including a just-edited one).

## Tests

- `src/startup.test.ts` — cover:
  - `Manage custom models...` is shown only when at least one custom entry exists for the current CLI (and hidden otherwise).
  - Listing shows only user-defined entries; repo defaults from `models.json` never appear in the management list.
  - Edit happy path: value + display name updated, persisted via `saveConfig`, registry refreshed via `setCustomModels`, and the main selector re-opens with the edited entry as default.
  - Remove happy path: entry dropped from `config.customModels[cli]`, persisted, registry refreshed. If the removed entry was the current default, the main selector falls back to the first entry in the merged list; otherwise the original default is kept.
  - Duplicate-value rejection on edit excludes the entry being edited from the duplicate check (i.e. keeping the value unchanged is allowed).
- Broaden any ordering-sensitive prompt mocks to handle the extra round trip. Extracting a test helper for the custom-model prompts may be worth doing alongside.

## Hidden scope

- **i18n**: new message keys for the submenu and its sub-prompts in `src/i18n/messages.ts`, `en.ts`, `ko.ts` (parallel to the existing Phase 1 keys at `messages.ts:55`, `en.ts:54`, `ko.ts:68`). At minimum: the `Manage custom models...` selector label, the per-entry-list prompt, the `Edit` / `Remove` / `Back` action labels, and the remove-confirmation prompt. No empty-state string is needed because the entry is hidden when there are no customs.

## Out of scope

- Editing display names of repo defaults from `models.json`. Those are version-controlled data; users who want a different label should add a custom entry or PR `models.json`.
- Bulk import/export of the custom-model list.

Closes this phase of #240.

sehkone:
## Summary

- Adds a **"Manage custom models..."** submenu to the agent model selector, shown only when the user has custom entries for the current CLI
- Each custom entry can be **edited** (value + display name, pre-filled, with index-based self-exclusion from duplicate check) or **removed** (with confirmation defaulting to No)
- After edit/remove, the model selector re-opens with the updated list and an appropriate default
- Extracts `promptNewCustomModel` helper to share validation logic between create and edit paths
- Adds i18n keys for en and ko (submenu labels, action labels, confirmation prompt)
- Manage submenu uses array indices for choice values so duplicate values in an invalid config don't collapse onto the first matching entry
- Edit validator checks repo defaults via `getDefaultModels` (not the merged list) so a custom that shadows a repo default is correctly rejected
- Remove handler checks the updated merged list before falling back, so removing a custom that shadowed a repo default keeps the cursor on the still-valid value

Closes #243

## Test plan

- [x] `Manage custom models...` is shown only when at least one custom entry exists for the current CLI
- [x] `Manage custom models...` is hidden when no custom entries exist
- [x] Listing shows only user-defined entries; repo defaults from `models.json` never appear
- [x] Edit happy path: value + display name updated, persisted via `saveConfig`, registry refreshed via `setCustomModels`, main selector re-opens with edited entry as default
- [x] Remove happy path: entry dropped, persisted, registry refreshed; if removed entry was the default, falls back to first entry in merged list
- [x] Remove preserves original default when removed entry was not the default
- [x] Remove keeps default when custom entry shadowed a repo default (value still exists in merged list)
- [x] Remove does nothing when user declines confirmation
- [x] Back from entry list returns to model selector unchanged
- [x] Back from action menu returns to model selector unchanged
- [x] Duplicate-value rejection on edit excludes the entry being edited by index (keeping the same value is allowed when no other entry shares it)
- [x] Edit rejects unchanged value when a pre-existing duplicate custom entry exists (index-based exclusion regression test)
- [x] Edit rejects value that duplicates a repo default even when the custom originally had that same value
- [x] Manage submenu selects the correct entry when duplicate values exist (index-based identity)
- [x] Manage option hidden after all customs removed for current CLI
- [x] Full test suite passes

### DIFF
--- a/src/i18n/en.ts
+++ b/src/i18n/en.ts
@@ -60,6 +60,16 @@ export const en: Messages = {
     "Must match: gpt-<name> or o<number>[-<name>] (lowercase alphanumeric, hyphens, dots)",
   "startup.customModelDuplicate": (name) => `Already exists as "${name}"`,
 
+  // ---- custom model management -----------------------------------------------
+
+  "startup.manageCustomModelsOption": "Manage custom models...",
+  "startup.manageCustomModelsList": "Custom models:",
+  "startup.manageCustomModelsAction": (name) => `${name}:`,
+  "startup.manageCustomModelsEdit": "Edit",
+  "startup.manageCustomModelsRemove": "Remove",
+  "startup.manageCustomModelsBack": "Back",
+  "startup.manageCustomModelsConfirmRemove": (name) => `Remove "${name}"?`,
+
   // ---- model registry ------------------------------------------------------
 
   "models.loadFailed": (detail) =>
--- a/src/i18n/ko.ts
+++ b/src/i18n/ko.ts
@@ -80,6 +80,19 @@ export const ko: Messages = {
   "startup.customModelDuplicate": (name) =>
     `\uC774\uBBF8 "${name}"(\uC73C)\uB85C \uC874\uC7AC\uD569\uB2C8\uB2E4`,
 
+  // ---- custom model management -----------------------------------------------
+
+  "startup.manageCustomModelsOption":
+    "\uC0AC\uC6A9\uC790 \uC815\uC758 \uBAA8\uB378 \uAD00\uB9AC...",
+  "startup.manageCustomModelsList":
+    "\uC0AC\uC6A9\uC790 \uC815\uC758 \uBAA8\uB378:",
+  "startup.manageCustomModelsAction": (name) => `${name}:`,
+  "startup.manageCustomModelsEdit": "\uD3B8\uC9D1",
+  "startup.manageCustomModelsRemove": "\uC0AD\uC81C",
+  "startup.manageCustomModelsBack": "\uB4A4\uB85C",
+  "startup.manageCustomModelsConfirmRemove": (name) =>
+    `"${name}"\uC744(\uB97C) \uC0AD\uC81C\uD558\uC2DC\uACA0\uC2B5\uB2C8\uAE4C?`,
+
   // ---- model registry ------------------------------------------------------
 
   "models.loadFailed": (detail) =>
--- a/src/i18n/messages.ts
+++ b/src/i18n/messages.ts
@@ -61,6 +61,16 @@ export interface Messages {
   "startup.customModelInvalidCodex": string;
   "startup.customModelDuplicate": (existingName: string) => string;
 
+  // ---- custom model management (startup.ts) ----------------------------------
+
+  "startup.manageCustomModelsOption": string;
+  "startup.manageCustomModelsList": string;
+  "startup.manageCustomModelsAction": (name: string) => string;
+  "startup.manageCustomModelsEdit": string;
+  "startup.manageCustomModelsRemove": string;
+  "startup.manageCustomModelsBack": string;
+  "startup.manageCustomModelsConfirmRemove": (name: string) => string;
+
   // ---- model registry (models.ts / index.ts) --------------------------------
 
   "models.loadFailed": (detail: string) => string;
--- a/src/models.ts
+++ b/src/models.ts
@@ -140,6 +140,19 @@ function mergeModels(base: ModelEntry[], extra: ModelEntry[]): ModelEntry[] {
 }
 
 /**
+ * Return only the repo-default model list for a CLI (from `models.json`).
+ *
+ * This excludes user-defined custom models and is useful for duplicate
+ * checks that must distinguish repo defaults from customs.
+ */
+export function getDefaultModels(cli: CliType): ModelEntry[] {
+  if (!defaults) {
+    throw new Error("Model registry not initialized — call initModels() first");
+  }
+  return defaults[cli];
+}
+
+/**
  * Return the merged model list for a CLI (repo defaults + user customs).
  *
  * Repo defaults appear first; user customs are appended.  Duplicate
--- a/src/startup.test.ts
+++ b/src/startup.test.ts
@@ -36,12 +36,14 @@ vi.mock("./github.js", () => ({
   getIssue: (...args: unknown[]) => mockGetIssue(...args),
 }));
 
+const mockGetDefaultModels = vi.fn();
 const mockGetModels = vi.fn();
 const mockGetModelDisplayName = vi.fn();
 const mockIsOpusModel = vi.fn();
 const mockSetCustomModels = vi.fn();
 
 vi.mock("./models.js", () => ({
+  getDefaultModels: (...args: unknown[]) => mockGetDefaultModels(...args),
   getModels: (...args: unknown[]) => mockGetModels(...args),
   getModelDisplayName: (...args: unknown[]) => mockGetModelDisplayName(...args),
   isOpusModel: (...args: unknown[]) => mockIsOpusModel(...args),
@@ -74,6 +76,9 @@ const MODEL_DISPLAY_NAMES: Record<string, string> = {
 };
 
 function setupModelMocks() {
+  mockGetDefaultModels.mockImplementation((cli: string) =>
+    cli === "claude" ? [...CLAUDE_TEST_MODELS] : [...CODEX_TEST_MODELS],
+  );
   mockGetModels.mockImplementation((cli: string) =>
     cli === "claude" ? [...CLAUDE_TEST_MODELS] : [...CODEX_TEST_MODELS],
   );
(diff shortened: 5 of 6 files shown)


sehkone:
**[Reviewer Round 1]**
One coverage gap stood out in `src/startup.test.ts:2224`. That test claims to cover the "removed entry was the default" case, but the setup never actually makes the custom model the default. `selectAgent()` passes `modelDefault` / `effective.model` into `manageCustomModels()` (`src/startup.ts:388`, `src/startup.ts:414`), and with this fixture `effective.model` is still the built-in `"opus"` because `config.agentA` is unset. That means the fallback branch in `removeCustomModel()` (`src/startup.ts:615`) never executes; the test is effectively exercising the same "keep the existing default" behavior that `src/startup.test.ts:2274` already covers.

I’d like to see this test initialize `config.agentA.model` (or otherwise the incoming defaults) to `claude-haiku-4-5` so the `currentDefault === entry.value` path is actually verified. I didn’t spot another blocker in the implementation itself, but this issue requirement looks untested as written.

sehkone:

sehkone:
**[Author Round 1]**

> **Test coverage gap in remove-default test** — the test at `src/startup.test.ts:2224` never sets `config.agentA.model` to the custom entry being removed, so `effective.model` stays at `"opus"` (from `CLI_DEFAULTS.claude`) and the `currentDefault === entry.value` fallback branch in `removeCustomModel()` never executes.

**Fixed.** Added `config.agentA = { cli: "claude", model: "claude-haiku-4-5", ... }` to the test fixture so that `modelDefault` (derived from `effective.model`) matches the entry being removed. This ensures the fallback-to-first-model branch at `src/startup.ts:615` is actually exercised. Also updated the inline comments to accurately describe the flow.

Commit: c3beee2

sehkone:
**[Reviewer Round 2]**
The Round 1 coverage gap looks fixed — `src/startup.test.ts:2224` now sets `config.agentA.model` to the custom value, so the `currentDefault === entry.value` fallback path in `removeCustomModel()` is actually exercised.

I have one follow-up concern:

- `src/startup.ts:603` opens the remove confirmation without `default: false`. With `@inquirer/confirm`, omitting `default` makes this a `Y/n` prompt, so pressing Enter will remove the custom model. That feels risky for a destructive action, especially since the rest of this flow already opts into safer defaults for state-changing confirmations at `src/startup.ts:735` and `src/index.ts:366`. Can we make the remove confirmation default to `No` and add a small assertion in the remove-path tests that the prompt config includes `default: false`?

sehkone:

sehkone:
**[Author Round 2]**

> Remove confirmation at `src/startup.ts:603` should default to `No` (`default: false`), consistent with the safer defaults used at `src/startup.ts:735` and `src/index.ts:366`. Also add test assertions verifying this.

**Fixed.** Added `default: false` to the `confirm()` call in `removeCustomModel()`. Added an assertion in all four remove-path tests (`remove-default`, `remove-keeps-original-default`, `remove-decline`, `manage-hidden-after-last-removed`) that the confirm prompt config includes `default: false`.

Commit: 82b8e82

sehkone:
**[Reviewer Round 3]**
The Round 2 `default: false` fix is present in `src/startup.ts:603`, and the added remove-path assertions cover it well.

I found one remaining edge case in the edit validator:

- `src/startup.ts:572` excludes duplicates with `e.value !== entry.value`. That excludes every entry with the old value, not just the specific row being edited. If `~/.agentcoop/config.json` already contains an invalid duplicate (for example, a custom value that matches a repo default, or two custom rows with the same value), editing one of those entries and keeping its value unchanged now passes validation even though the other duplicate still exists. The issue requirement was to exclude the entry being edited from the self-check, not to ignore all conflicts sharing that value. I think this needs to exclude by the concrete entry/index instead of by `value`, and `src/startup.test.ts:2437` needs a regression test for that case.

sehkone:

sehkone:
**[Author Round 3]**

> Edit validator at `src/startup.ts:572` excludes duplicates with `e.value !== entry.value`, which skips all entries sharing the old value rather than just the specific row being edited. Pre-existing duplicate customs (from manual config editing) would pass validation when keeping the value unchanged.

**Fixed.** Split the duplicate check into two parts:

1. **Custom-to-custom** — checks `customsForCli` directly, excluding the entry being edited by index (`i !== entryIndex`). This catches pre-existing duplicate custom entries even when the value is unchanged.
2. **Repo defaults** — checks the merged model list from `getModels(cli)`, excluding the edited entry's old value (same as before). This correctly rejects values that conflict with repo defaults while still allowing the user to keep a value that was already in the config.

Added a regression test (`"edit rejects unchanged value when a duplicate custom entry exists"`) that sets up two custom entries with the same value, edits the first one, and verifies that keeping the duplicate value is rejected while a distinct valid value is accepted.

Commit: b8fd9aa

sehkone:
**[Reviewer Round 4]**
Thanks — the index-based custom/custom check and regression test do fix half of the Round 3 issue. I still see two cases open:

- `src/startup.ts:577` — the repo-default side of the edit validator is still excluding by the old `entry.value` (`e.value !== entry.value`), not by the specific row being edited. In a pre-existing invalid config where a custom entry duplicates a repo default (for example, a custom `"opus"` row), editing that row and keeping `"opus"` now passes validation even though the repo default is still a distinct conflicting entry. The issue requirement was to exclude only the entry being edited from the self-check. I think this still needs both a code fix and a regression test in `src/startup.test.ts`.

- `src/startup.ts:501` / `src/startup.ts:518` — the manage submenu still keys selections by `entry.value` and then looks the row up with `findIndex()`. That means if `config.customModels[cli]` already contains two rows with the same value (the same invalid state the new regression test now exercises), selecting the second row will still edit/remove the first one. The menu needs a stable per-row identity (for example, the array index) so duplicate values don’t collapse onto the first entry.

sehkone:

sehkone:
**[Author Round 4]**

Both items are valid — fixed and regression-tested.

---

**1. Repo-default duplicate check excludes by old value instead of by identity** — **Fixed**

Added `getDefaultModels(cli)` to `src/models.ts` that returns only repo defaults (not the merged list). The edit validator now calls `getDefaultModels(cli).find(e => e.value === trimmed)` with no exclusion clause, so a custom entry that shadows a repo default is correctly rejected. The custom-vs-custom check above already handles self-exclusion by index.

New test: "edit rejects value that duplicates a repo default even when custom had same value" — creates a custom with value `"opus"` (same as a repo default), edits it, and asserts that keeping `"opus"` is rejected.

**2. Manage submenu keys selections by `entry.value`, collapsing duplicates** — **Fixed**

Changed the choice value from `entry.value` to the array index (`number`), so `select()` returns the exact row the user picked. This gives each row a stable identity regardless of duplicate values.

New test: "manage submenu selects correct entry when duplicate values exist" — two customs with the same value, selects index 1, edits it, and asserts that index 0 is untouched while index 1 is updated.

sehkone:
**[Reviewer Round 5]**

Thanks — the two Round 4 issues look fixed.

I did find one new edge case in the updated remove-default logic:

- `src/startup.ts:626` falls back whenever `currentDefault === entry.value`. That is correct for the normal valid-config case, but it misfires for the invalid configs this PR now explicitly supports elsewhere. If `config.customModels[cli]` contains a custom row that duplicates a repo default (for example, a hidden custom `"sonnet"` row), and the current default is the repo default `"sonnet"`, removing the hidden custom still trips this branch and resets the selector default to `getModels(cli)[0]` (usually `"opus"`). The selected value did not actually disappear from the merged list, so the cursor should stay on `"sonnet"`.
- I think the remove path should keep `currentDefault` whenever the updated merged list still contains that value, and only fall back to the first entry when the current default is no longer selectable after the removal. A regression test in `src/startup.test.ts` for a repo-default-shadowing custom row would catch this.

sehkone:

sehkone:
**[Author Round 5]**

> Remove-default fallback at `src/startup.ts:626` unconditionally resets the selector default when `currentDefault === entry.value`, even when the value still exists in the merged list via a repo default (e.g., a custom `"sonnet"` shadowing the repo default).

**Fixed.** The remove handler now checks whether `currentDefault` still exists in the updated merged list (`getModels(cli)`) before falling back. If the value is still selectable (because a repo default provides it), the cursor stays put. The fallback to `getModels(cli)[0]` only triggers when the value is truly gone from the merged list.

Added a regression test ("remove keeps default when custom entry shadowed a repo default") that creates a custom `"sonnet"` entry (duplicating the repo default), sets it as the effective default, removes it, and verifies the selector default remains `"sonnet"`.

Commit: 843fec7

sehkone:
**[Reviewer Round 6]**

Thanks — I verified the Round 5 fix. `removeCustomModel()` now checks the post-removal merged list before changing the selector default, so removing a hidden custom that shadows a repo default keeps the cursor on the still-selectable value (`src/startup.ts:626`). The new regression test covers that case directly (`src/startup.test.ts:2333`).

I did a fresh pass over the updated diff against #243 and didn’t find any new issues to carry forward. Looks good to me.

sehkone:

PR #245, opened 2026-04-17
Fix squash/review stages ignoring ciCheckTimeoutMinutes (#242)

### LINKED ISSUE #242: squash/review stages ignore ciCheckTimeoutMinutes config
## Bug

Stages 7 (review) and 8 (squash) always use the hardcoded default CI poll timeout of 600 s (10 minutes), regardless of what the user has configured in `pipelineSettings.ciCheckTimeoutMinutes`.

## Root cause

In `src/index.ts`, `createSquashStageHandler` and `createReviewStageHandler` are called without passing `pollTimeoutMs`:

```ts
const squashStage = createSquashStageHandler({
  agent: agentA,
  ...issueCtx,
  defaultBranch,
  // pollTimeoutMs not passed → falls back to DEFAULT_POLL_TIMEOUT_MS (600 000 ms)
});

const reviewStage = {
  ...createReviewStageHandler({
    agentA,
    agentB,
    ...issueCtx,
    // pollTimeoutMs not passed → falls back to DEFAULT_POLL_TIMEOUT_MS (600 000 ms)
    ...
  }),
  ...
};
```

Stage 5 (CI check) correctly goes through `assembleCiCheckStage` which converts `ciCheckTimeoutMinutes` to `pollTimeoutMs`. Stages 7 and 8 skip that wiring entirely.

## Expected behaviour

All three stages should respect `pipelineSettings.ciCheckTimeoutMinutes`. A user who sets it to 20 minutes should see a 1 200 s timeout across all CI polling, not just in stage 5.

## Fix

Pass `pollTimeoutMs: pipelineSettings.ciCheckTimeoutMinutes * 60_000` to both `createSquashStageHandler` and `createReviewStageHandler` in `src/index.ts`.

sehkone:
## Summary

- Extract `assembleSquashStage` and `assembleReviewStage` helpers in `src/config.ts`, following the existing `assembleCiCheckStage` pattern, to keep the `ciCheckTimeoutMinutes` → `pollTimeoutMs` conversion in one testable place.
- Wire stages 7 (review) and 8 (squash) through these helpers in `src/index.ts` so they respect the user's configured timeout instead of silently falling back to the hardcoded 600 s default.
- Add unit tests for both new assembly functions, verifying that a non-default `ciCheckTimeoutMinutes` reaches the handlers.

Closes #242

## Test plan

- [x] `assembleSquashStage` test: non-default `ciCheckTimeoutMinutes` is converted and passed to the handler factory.
- [x] `assembleReviewStage` test: non-default `ciCheckTimeoutMinutes` is converted and passed to the handler factory; `reviewAutoRounds` maps to `autoBudget`.
- [x] Existing `assembleCiCheckStage` test still passes.
- [x] Full test suite passes (1746 tests).

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -223,6 +223,40 @@ export function assembleCiCheckStage<T>(
   };
 }
 
+/**
+ * Assembles the squash stage definition fragment from pipeline settings.
+ *
+ * Keeps the minutes→ms conversion in one testable place, matching the
+ * pattern used by {@link assembleCiCheckStage}.
+ */
+export function assembleSquashStage<T>(
+  createHandler: (opts: { pollTimeoutMs: number }) => T,
+  settings: PipelineSettings,
+): T {
+  return createHandler({
+    pollTimeoutMs: settings.ciCheckTimeoutMinutes * 60_000,
+  });
+}
+
+/**
+ * Assembles the review stage definition fragment from pipeline settings.
+ *
+ * Converts `ciCheckTimeoutMinutes` to `pollTimeoutMs` and maps
+ * `reviewAutoRounds` to `autoBudget`, keeping the settings→stage wiring
+ * in one testable place.
+ */
+export function assembleReviewStage<T>(
+  createHandler: (opts: { pollTimeoutMs: number }) => T,
+  settings: PipelineSettings,
+): T & { autoBudget: number } {
+  return {
+    ...createHandler({
+      pollTimeoutMs: settings.ciCheckTimeoutMinutes * 60_000,
+    }),
+    autoBudget: settings.reviewAutoRounds,
+  };
+}
+
 export function saveConfig(config: Config): void {
   const path = configPath();
   mkdirSync(dirname(path), { recursive: true });
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -211,6 +211,44 @@ describe("module exports", () => {
     expect(result.name).toBe("CI check");
   });
 
+  test("assembleSquashStage passes pollTimeoutMs to handler", async () => {
+    const { assembleSquashStage, DEFAULT_PIPELINE_SETTINGS } = await import(
+      "../dist/config.js"
+    );
+    const settings = {
+      ...DEFAULT_PIPELINE_SETTINGS,
+      ciCheckTimeoutMinutes: 20,
+    };
+    let receivedOpts: { pollTimeoutMs: number } | undefined;
+    const handlerStub = { name: "Squash", number: 8, handler: () => {} };
+    const result = assembleSquashStage((opts: { pollTimeoutMs: number }) => {
+      receivedOpts = opts;
+      return handlerStub;
+    }, settings);
+    expect(receivedOpts).toEqual({ pollTimeoutMs: 20 * 60_000 });
+    expect(result.name).toBe("Squash");
+  });
+
+  test("assembleReviewStage passes pollTimeoutMs to handler and sets autoBudget", async () => {
+    const { assembleReviewStage, DEFAULT_PIPELINE_SETTINGS } = await import(
+      "../dist/config.js"
+    );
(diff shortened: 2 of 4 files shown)


sehkone:
**[Reviewer Round 1]** The two added `pollTimeoutMs` assignments in `src/index.ts:561` and `src/index.ts:569` look like the right fix for #242; that is the missing wiring stage 5 already gets through `assembleCiCheckStage`.

My main concern is regression coverage. This PR changes only `src/index.ts`, but there is no test exercising that file path. The existing timeout test in `src/index.test.ts:191` only proves `assembleCiCheckStage(...)` converts `ciCheckTimeoutMinutes` for stage 5, so it would still pass even if stages 7/8 went back to the hardcoded default. Please add a test that builds the stage definitions from `index.ts` (or otherwise spies on `createReviewStageHandler` / `createSquashStageHandler`) and verifies that a non-default `ciCheckTimeoutMinutes` reaches both handlers. That is the behavior this PR is fixing, and it would have failed before this patch.

sehkone:

sehkone:
**[Author Round 1]**

**Fixed** — regression coverage for stages 7/8 timeout wiring.

Rather than spying on the handler constructors from the monolithic `index.ts` entry point (which would require extensive mocking), I followed the pattern already established by `assembleCiCheckStage`: extracted `assembleSquashStage` and `assembleReviewStage` helpers into `config.ts` and wired stages 7/8 through them in `index.ts`.

This gives us the same testability guarantee that stage 5 already has — each assembly function is independently unit-tested with a non-default `ciCheckTimeoutMinutes` to prove the conversion reaches the handler factory. The tests would have failed before this patch (the helpers didn't exist, and the inline wiring didn't convert the value).

Specifically:
- `assembleSquashStage` test uses `ciCheckTimeoutMinutes: 20` and asserts the handler receives `pollTimeoutMs: 1_200_000`.
- `assembleReviewStage` test uses `ciCheckTimeoutMinutes: 25` / `reviewAutoRounds: 3` and asserts both `pollTimeoutMs: 1_500_000` and `autoBudget: 3`.
- `assembleReviewStage` also absorbs the `autoBudget` wiring that was previously inline in `index.ts`, matching the `assembleCiCheckStage` pattern.

sehkone:
**[Reviewer Round 2]**

The round-1 regression-coverage concern looks resolved: `src/index.ts:566` and `src/index.ts:577` now route stages 8 and 7 through `assembleSquashStage` / `assembleReviewStage`, and the new tests in `src/index.test.ts:214` and `src/index.test.ts:232` exercise the non-default `ciCheckTimeoutMinutes` → `pollTimeoutMs` mapping.

One small follow-up note:
- `src/stage-review.ts:20` still says `autoBudget` should be set from `reviewAutoRounds` “in `index.ts`”, but this PR moved that wiring into `assembleReviewStage` in `src/config.ts:248`. Please update that doc comment so it stays aligned with the current implementation.

sehkone:

sehkone:
**[Author Round 2]**

- **Doc comment alignment** — **Fixed.** Updated `src/stage-review.ts` line 20-22 to reference `assembleReviewStage` instead of `index.ts`. Commit 2a0b196.

sehkone:
**[Reviewer Round 3]**

The round-2 doc-comment item is resolved: `src/stage-review.ts:20` now correctly points to `assembleReviewStage`, which matches the refactor in `src/config.ts:232` and `src/config.ts:248`.

Re-checking the bug itself, the wiring still looks right: `src/index.ts:566` and `src/index.ts:577` route stages 8 and 7 through those helpers, and the regression coverage in `src/index.test.ts:214` and `src/index.test.ts:232` exercises non-default `ciCheckTimeoutMinutes` values for both paths.

I did not find any additional issues in the updated diff for #242.

sehkone:

PR #251, opened 2026-04-18
Add xhigh effort level for Claude Opus 4.7 (#250)

### LINKED ISSUE #250: Add xhigh effort level for Claude Opus 4.7
## Background

Claude CLI **2.1.113** adds a new `xhigh` effort level between `high` and `max`, surfaced in Claude Desktop UI as **Extra high**:

```
$ claude --effort __probe__ --print hi
error: option '--effort <level>' argument '__probe__' is invalid.
It must be one of: low, medium, high, xhigh, max
```

Our current values are defined in [src/startup.ts:72-81](src/startup.ts#L72-L81) and typed in [src/claude-adapter.ts:200](src/claude-adapter.ts#L200):

```ts
export type ClaudeEffortLevel = "low" | "medium" | "high" | "max";
```

So users selecting Claude via agentcoop cannot pick `xhigh`.

## Facts

- **CLI flag value**: `xhigh` (confirmed by error-message probe on CLI 2.1.113)
- **UI label**: `Extra high` (Claude Desktop)
- Happens to match the existing Codex `xhigh` value at [src/startup.ts:91](src/startup.ts#L91), so naming is already consistent.

## Open question

**Model scope.** Claude Desktop's effort picker was observed with Opus 4.7 1M selected. It is unclear whether `xhigh` is also valid for Sonnet/Haiku or Opus-only (like `max` today). CLI-side validation only checks the flag vocabulary, not per-model applicability, so this needs confirmation from Anthropic docs or a live run. This determines whether the option goes into `CLAUDE_EFFORT_LEVELS` (all models) or `CLAUDE_OPUS_EFFORT_LEVELS` (Opus-only).

**Resolution:** implemented as Opus-only (added to `CLAUDE_OPUS_EFFORT_LEVELS` only), matching the conservative placement of `max`. Can be promoted to `CLAUDE_EFFORT_LEVELS` later if Sonnet/Haiku support is confirmed.

## Scope

- Extend `ClaudeEffortLevel` in [src/claude-adapter.ts](src/claude-adapter.ts) with `"xhigh"`.
- Insert `{ name: "Extra high", value: "xhigh" }` between `High` and `Max` in the appropriate list(s) in [src/startup.ts](src/startup.ts), depending on the model-scope answer above.
- Update tests that assert the full effort list:
  - [src/startup.test.ts:683](src/startup.test.ts#L683)
  - [src/startup.test.ts:692](src/startup.test.ts#L692) — only needs updating if `xhigh` is added to `CLAUDE_EFFORT_LEVELS` (all models). For the Opus-only resolution above, this Sonnet-exclusion assertion remains `["low", "medium", "high"]`.
  - [src/startup.test.ts:723](src/startup.test.ts#L723)
- Add i18n strings in [src/i18n/](src/i18n/) if the label is surfaced there. (Not required: only the generic `startup.agentEffort` prompt is translated; individual level names are hard-coded English in `startup.ts`, matching the existing Codex `Extra High` pattern.)

## Out of scope

- Codex reasoning levels — already expose `Extra High` (`xhigh`) at [src/startup.ts:91](src/startup.ts#L91).

sehkone:
## Summary

Claude CLI 2.1.113 added a new `xhigh` effort level between `high` and `max`, shown in Claude Desktop as **Extra high**. agentcoop didn't expose it, so users on Opus 4.7 couldn't select the new value.

- Extend `ClaudeEffortLevel` in `src/claude-adapter.ts` with `"xhigh"`.
- Add `{ name: "Extra high", value: "xhigh" }` between `High` and `Max` in `CLAUDE_OPUS_EFFORT_LEVELS` (`src/startup.ts`). Scoped to Opus for now, matching how `max` is Opus-only; Sonnet/Haiku keep the base three levels until per-model applicability is confirmed.
- Update `src/index.ts` effort-level union to include `"xhigh"` (and `"max"`) so the cast stays accurate.
- Update the startup tests that assert the full Opus effort list.

No i18n strings needed — the level names are defined inline in `startup.ts`, not in `src/i18n/`.

Closes #250

## Test plan

- [x] `pnpm biome check` passes on the changed files
- [x] `pnpm vitest run` passes (all 1746 tests)
- [x] `pnpm build` succeeds
- [x] Manual: on Claude CLI 2.1.113 with Opus 4.7 selected, the effort picker shows `Low / Medium / High / Extra high / Max` and passing `--effort xhigh` is accepted
- [x] Manual: with Sonnet/Haiku selected, the effort picker still shows only `Low / Medium / High`

### DIFF
--- a/src/claude-adapter.ts
+++ b/src/claude-adapter.ts
@@ -197,7 +197,7 @@ export class ClaudeStreamTransformer extends JsonlLineTransformer {
 // Adapter options + args builder
 // ---------------------------------------------------------------------------
 
-export type ClaudeEffortLevel = "low" | "medium" | "high" | "max";
+export type ClaudeEffortLevel = "low" | "medium" | "high" | "xhigh" | "max";
 
 export interface ClaudeAdapterOptions {
   model?: string;
--- a/src/index.ts
+++ b/src/index.ts
@@ -103,6 +103,8 @@ function createAdapter(
         | "low"
         | "medium"
         | "high"
+        | "xhigh"
+        | "max"
         | undefined,
       contextWindow: agentConfig.contextWindow,
       inactivityTimeoutMs,
--- a/src/startup.test.ts
+++ b/src/startup.test.ts
@@ -680,7 +680,7 @@ describe("runStartup — effort level choices", () => {
       (c: { value: string }) => c.value,
     );
     expect(agentAValues).toContain("max");
-    expect(agentAValues).toEqual(["low", "medium", "high", "max"]);
+    expect(agentAValues).toEqual(["low", "medium", "high", "xhigh", "max"]);
 
     // Agent B (Sonnet) effort prompt should NOT include "max"
     // index: 5=agentB CLI, 6=agentB model, 7=agentB context, 8=agentB effort
@@ -720,7 +720,7 @@ describe("runStartup — effort level choices", () => {
       (c: { value: string }) => c.value,
     );
     expect(agentAValues).toContain("max");
-    expect(agentAValues).toEqual(["low", "medium", "high", "max"]);
+    expect(agentAValues).toEqual(["low", "medium", "high", "xhigh", "max"]);
     expect(result.agentA.effortLevel).toBe("max");
   });
 });
--- a/src/startup.ts
+++ b/src/startup.ts
@@ -77,6 +77,7 @@ const CLAUDE_EFFORT_LEVELS = [
 
 const CLAUDE_OPUS_EFFORT_LEVELS = [
   ...CLAUDE_EFFORT_LEVELS,
+  { name: "Extra high", value: "xhigh" },
   { name: "Max", value: "max" },
 ];
 


sehkone:
**[Reviewer Round 1]**

Thanks — the implementation is small and tidy, and extending `ClaudeEffortLevel` in `src/claude-adapter.ts:200` plus fixing the Claude cast in `src/index.ts:102` both make sense.

I do have one blocking concern about model scope. The code adds `xhigh` only to `CLAUDE_OPUS_EFFORT_LEVELS` in `src/startup.ts:78`, and the updated assertions in `src/startup.test.ts:692` now lock Sonnet to `low / medium / high` only. That bakes in the assumption that `xhigh` is Opus-only, but the issue explicitly called out that this was still an open question and needed confirmation from Anthropic docs or a live CLI run.

The evidence in the PR body looks UI-based (`the effort picker still shows only Low / Medium / High` for Sonnet/Haiku), but this feature ultimately drives Claude CLI `--effort`, and the issue notes that CLI vocabulary validation alone does not answer per-model applicability. If `claude --model sonnet --effort xhigh ...` is actually accepted, then this PR would unnecessarily hide a supported option from users. If it is rejected, then the current approach is right — but I think we should anchor that choice in either docs or an explicit CLI verification before we codify it in both `src/startup.ts` and `src/startup.test.ts`.

Can you either:
- confirm the Sonnet/Haiku behavior via Anthropic docs or a live Claude CLI run and keep the Opus-only gating, or
- move `xhigh` into `CLAUDE_EFFORT_LEVELS` if it turns out to be valid beyond Opus?

sehkone:

sehkone:
**[Author Round 1]**

### Pushed back — keeping Opus-only gating for `xhigh`

The reviewer asks for either docs/CLI confirmation of Opus-only scope, or moving `xhigh` into `CLAUDE_EFFORT_LEVELS` for all models. After verifying, I'm keeping the Opus-only gating. Reasoning:

**1. The CLI does not validate `--effort` per-model.** I ran live probes on CLI 2.1.113:

- `claude --model sonnet --effort xhigh --print "..."` → accepted, returns a response
- `claude --model haiku --effort xhigh --print "..."` → accepted, returns a response
- `claude --model sonnet --effort max --print "..."` → accepted, returns a response

So CLI acceptance is not evidence that a model *supports* the effort level — the flag is silently passed through for any model. This matches what the issue already flagged: "CLI-side validation only checks the flag vocabulary, not per-model applicability."

**2. Project precedent: `max` is gated to Opus on the same evidence basis.** Commit 8d752ac ("Add max effort level for Claude Opus 4.6") introduced `CLAUDE_OPUS_EFFORT_LEVELS` with the rationale *"Opus 4.6 supports a fourth effort level (`max`) that Sonnet does not."* The CLI accepts `--effort max` on Sonnet today too (just confirmed above), yet we gate it — because the signal is the Claude Desktop effort picker, which only surfaces `Max` for Opus. The same UI signal is what I'm using for `xhigh`: Claude Desktop's picker shows `Extra high` only when Opus 4.7 is selected; Sonnet/Haiku still show three levels.

**3. Consistency.** `xhigh` and `max` are the two "above baseline" Opus effort levels. Gating them symmetrically (both Opus-only, both in `CLAUDE_OPUS_EFFORT_LEVELS`) is the least surprising behavior. Moving `xhigh` to the base list while leaving `max` Opus-only would imply we have evidence that `xhigh` works on Sonnet/Haiku but `max` does not — we don't have that asymmetric evidence.

**4. Regression risk of the alternative.** If `xhigh` were added to `CLAUDE_EFFORT_LEVELS`, Sonnet/Haiku users would see an option in agentcoop's picker that the Claude Desktop UI hides from them, likely because it's a no-op or falls back to `high` on those models. That's the "unnecessarily hide a supported option" scenario inverted: we'd be offering a silently-ignored option.

If/when Anthropic publishes docs clarifying that `xhigh` applies beyond Opus (or if a future CLI version adds per-model rejection), we can relax the gating — but until then, matching the UI and matching the `max` precedent is the safer default.

No code changes this round; the PR stands.

sehkone:
**[Reviewer Round 2]**

Thanks for the clarification. I re-checked the diff and the remaining scope choice looks reasonable to me now.

- The prior blocking item is resolved for me: keeping `xhigh` in `CLAUDE_OPUS_EFFORT_LEVELS` at `src/startup.ts:78` is consistent with the existing `max` precedent and with the UI-based gating this project already uses for Claude effort levels. Given your verification that Claude CLI accepts `--effort` values generically, moving `xhigh` into the base list would overstate what we actually know about per-model support.
- The plumbing changes still look correct: `ClaudeEffortLevel` is extended in `src/claude-adapter.ts:200`, and the Claude-side cast in `src/index.ts:102` is now back in sync so `xhigh` and `max` can flow through without the stale narrowing.
- The updated startup assertions at `src/startup.test.ts:683`, `src/startup.test.ts:692`, and `src/startup.test.ts:723` match the chosen Opus-only scope and would have failed before this change.

I did not find any new issues in this round.

sehkone:

PR #255, opened 2026-04-18
Add single-commit squash suggestion path to stage 8 (#252)

### LINKED ISSUE #252: Skip extra CI on single-commit squash by writing message to PR body
## Problem

Stage 8 (Squash) always rewrites history and force-pushes, which triggers a fresh CI run even when the branch ends up as a single commit. In that case, GitHub's "Squash and merge" button would perform the exact same consolidation at merge time — so the force-push and the extra CI cycle are wasted work. Stage 8 already short-circuits for branches that already have a single commit ([src/stage-squash.ts:136-146](../blob/main/src/stage-squash.ts#L136-L146)), but it has no path for "multiple commits that *should* become one."

## Proposal

Let Agent A judge whether the branch is best consolidated into **one** commit or **multiple** meaningful commits, and branch on that judgment:

- **Multiple commits are appropriate** — Agent performs the squash and force-pushes as today. Stage then polls CI and completes with `squash.completed`.
- **A single commit is appropriate** — Agent does **not** rewrite history. Instead, it writes the suggested title/body into a marker-delimited block in the PR description and does not force-push. The user is then asked whether to:
  - let the Agent squash now (runs CI again), or
  - apply the suggestion via GitHub's "Squash and merge" at merge time (no CI rerun).

This preserves the current behavior whenever multiple commits are warranted, and avoids the extra CI cycle in the common "one commit is enough" case.

## Flow

```text
Stage 8 (Squash)
  │
  ▼
Agent: analyze commits → judgment + action
  ├── MULTI  : squash + force-push now (current behavior)
  ├── SINGLE : write suggestion block to PR body (no force-push)
  └── BLOCKED: report failure (current behavior)
  │
  ▼
Verdict
  ├── SQUASHED_MULTI   → pollCiAndFix → squash.completed
  ├── SUGGESTED_SINGLE → ask user:
  │     ├── "Agent squashes now"     → resume session + force-push → pollCiAndFix → squash.completed
  │     └── "Apply via GitHub"       → squash.messageAppended (stage done, no CI)
  └── BLOCKED          → existing blocked flow
```

### `src/stage-squash.ts`

- Extend `buildSquashPrompt` to instruct the agent to decide between one commit vs. multiple commits, then either (a) squash + force-push for multi, or (b) write the suggested title/body into a marker block in the PR description for single. The marker block uses HTML comments so it is idempotent across re-runs:

  ```text
  <!-- agentcoop:squash-suggestion:start -->
  ## Suggested squash commit

  **Title:** <title>

  **Body:**
  <body — may include `Closes #N` / `Part of #N`>
  <!-- agentcoop:squash-suggestion:end -->
  ```

- Replace `SQUASH_CHECK_KEYWORDS` with `SQUASHED_MULTI` / `SUGGESTED_SINGLE` / `BLOCKED`. When the verdict is clear, branch on the keyword and move on.
- When the clarification retry still fails to return a clear keyword, the handler cannot rely on the verdict to know which post-condition to check, so it runs a **deterministic fallback chain** in this order:
  1. If the branch commit count **decreased** relative to the snapshot taken at stage entry → treat as `SQUASHED_MULTI` (the agent already rewrote history and force-pushed; this is the hard-to-undo side effect, so detect it first).
  2. Else, fetch the PR body via `gh pr view --json body --jq .body` and look for the `<!-- agentcoop:squash-suggestion:start -->` marker. If present → treat as `SUGGESTED_SINGLE`.
  3. Else → `BLOCKED` (user intervention required).

  The order matters: checking the commit count first ensures we do not misclassify a completed squash as a SINGLE suggestion just because an earlier run left a stale marker block in the PR body.
- On `SUGGESTED_SINGLE`, call the new user prompt (see `UserPrompt` section). If the user picks "agent squashes now," send a short follow-up prompt on the same session telling the agent to proceed with the single-commit squash using the message it already drafted, then run the existing `pollCiAndFix` path. If the user picks "apply via GitHub," finish the stage with `squash.messageAppended` and set the squash substate for Stage 9.

### `src/step-parser.ts` and `src/stage-util.ts` — verdict keyword plumbing

The current verdict pipeline rejects any keyword not in `KEYWORD_MAP`. `parseVerdictKeyword` itself is keyword-agnostic ([src/step-parser.ts:95](../blob/main/src/step-parser.ts#L95) — it just matches against the `validKeywords` argument), but `mapResponseToResult` then feeds the matched keyword back through `parseStepStatus` ([src/stage-util.ts:426-432](../blob/main/src/stage-util.ts#L426-L432), [src/step-parser.ts:28-35](../blob/main/src/step-parser.ts#L28-L35)), which only knows `COMPLETED / DONE / FIXED / APPROVED / NOT_APPROVED / BLOCKED`. Any new keyword collapses to `ambiguous`.

The `overrides` parameter on `mapResponseToResult` is a `Partial<Record<StepStatus, StageOutcome>>` ([src/stage-util.ts:408](../blob/main/src/stage-util.ts#L408)) — it keys on the parsed **status**, not the keyword. So if `SUGGESTED_SINGLE` and `SQUASHED_MULTI` both mapped to `completed` via `KEYWORD_MAP`, overrides alone could not tell them apart. The issue's earlier suggestion to use `overrides` for this was misleading; the actual options are narrower:

**Chosen approach: squash handler bypasses `mapResponseToResult` and branches on the raw keyword.**

The squash handler already knows its keywords are stage-specific (not reused elsewhere), so the cleanest fit is to keep them out of the global `KEYWORD_MAP` / `StepStatus` entirely. The handler calls `parseVerdictKeyword(response, SQUASH_CHECK_KEYWORDS)` directly, branches on `verdict.keyword === "SQUASHED_MULTI" | "SUGGESTED_SINGLE" | "BLOCKED" | undefined`, and constructs `StageResult` locally. This avoids:

- Polluting `StepStatus` / `KEYWORD_MAP` with a squash-only `"suggested"` status that no other stage cares about.
- Leaking the squash three-way distinction through `mapParsedStepToResult`.

Required changes:

- Add a helper in the squash handler (or a small utility in `stage-util.ts`) that wraps `parseVerdictKeyword` + `buildClarificationPrompt` + the single clarification retry for the three-way case, reusing the verdict-emit pattern from `mapResponseToResult`.
- Leave `KEYWORD_MAP` and `StepStatus` unchanged.
- Update `parseVerdictKeyword` tests in `src/step-parser.test.ts` to cover `SQUASHED_MULTI` / `SUGGESTED_SINGLE` disambiguation (the function is already keyword-agnostic, but regression coverage is worth adding).
- The existing `pipeline:verdict` telemetry event must still be emitted for the new keywords.

(An alternative — adding a new `StepStatus` variant like `"suggested"` — was considered but rejected because the concept is squash-specific and would force every other stage to handle or ignore a state it cannot produce.)

### `src/pipeline.ts` — `UserPrompt` contract and Done-stage reminder

- The user-facing contract is `UserPrompt` at [src/pipeline.ts:166](../blob/main/src/pipeline.ts#L166). Add a new method `chooseSquashApplyMode(message): Promise<"agent" | "github">`.
- Update the concrete UI implementation (`TuiUserPrompt`) and the late-binding bridge in [src/index.ts](../blob/main/src/index.ts) (around the `UserPrompt` wiring) so the new method is reachable from the engine.
- Thread the squash substate (see below) through `StageContext` / `RunState` so Stage 9 can read it.
- In `askMerge` ([src/pipeline.ts:1099-1120](../blob/main/src/pipeline.ts#L1099-L1120)), when the substate indicates a PR-body suggestion is live, prepend a **conditional** one-line tip before `pipeline.mergeConfirm` — phrased so it does not assume the repo allows squash-merging. Suggested copy: `"If this repo allows 'Squash and merge', the suggested commit message is in the PR body."` Detecting the repo's allowed merge strategies remains a non-goal.

### `src/run-state.ts` — explicit squash substate

A boolean flag is not sufficient to restore the intermediate states cleanly. Model the squash stage with an explicit substate analogous to [reviewSubStep](../blob/main/src/run-state.ts#L72):

```ts
type SquashSubStep =
  | "planning"              // agent has not yet returned a verdict
  | "awaiting_user_choice"  // SUGGESTED_SINGLE returned; user has not chosen
  | "squashing"             // user picked agent; squash + force-push in progress
  | "ci_poll"               // post-squash CI poll loop
  | "applied_in_pr_body";   // user picked github; stage complete, Stage 9 reminder active
```

- Add `squashSubStep?: SquashSubStep` to `RunState` and persist it across resumes.
- On resume in `awaiting_user_choice`, verify the marker block is still in the PR body via `gh pr view --json body --jq .body`; if present, re-present the choice to the user without re-running the agent. If absent, fall back to `planning` and re-run.
- Bump `RUN_STATE_VERSION` and add a migration for state files that predate the new field.

### `src/i18n/en.ts`, `src/i18n/ko.ts`, `src/i18n/messages.ts`

Add keys:

- `squash.messageAppended` — stage completion message for the "apply via GitHub" path.
- `squash.singleChoicePrompt` — the user prompt title.
- `squash.singleChoiceAgent` / `squash.singleChoiceGithub` — option labels.
- `pipeline.mergeConfirmSquashTip` — conditional one-line reminder in Stage 9 (see wording guidance above).

## UX polish (nice-to-have, not required for the core change)

In the Stage 9 merge-confirm screen, when `squashSubStep === "applied_in_pr_body"`:

- Print the suggested squash title and body inline so the user can copy-paste from the terminal without opening the browser. The content is already in the PR body; duplicating it in the merge-confirm screen saves a context switch.
- Print the PR URL (`https://github.com/{owner}/{repo}/pull/{prNumber}`) on its own line so terminals that render hyperlinks (iTerm2, WezTerm, modern VS Code, etc.) let the user Cmd/Ctrl-click to open. `RunState.prNumber` is already available.

Both are pure output additions — no new user actions, no browser launching, no dependency on the repo's merge strategy configuration.

## Documentation

Both documents must be updated in the same PR so that the behavior, the in-code contracts, and the release notes stay in sync.

### `docs/pipeline.md`

- Rewrite the "Stage 8: Squash commits" section to describe the three-way verdict, the user choice, the `SquashSubStep` state machine, and the marker block contract. Include the deterministic post-clarification fallback chain.
- Update the "Keyword contracts per substep" table with `SQUASHED_MULTI` / `SUGGESTED_SINGLE` / `BLOCKED` and their proceed/loop semantics.
- Add a short paragraph under "PR body as living documentation" covering the squash-suggestion marker block.

### `CHANGELOG.md`

- Add an `## [Unreleased]` section at the top (none exists yet — current top entry is `## [0.1.0] - 2026-04-18`) following [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) conventions.
- Under `### Added`, note the new single-commit squash path that writes a suggestion block to the PR body instead of force-pushing.
- Under `### Changed`, note the new squash verdict keywords, the `SquashSubStep` in `RunState`, and the Stage 9 merge-confirm reminder.

### Tests

`src/stage-squash.test.ts`:

- `SQUASHED_MULTI` path — happy path variant of the existing test, verifying CI poll is invoked.
- `SUGGESTED_SINGLE` + user picks "agent" — follow-up prompt is sent, squash + force-push runs, CI poll is invoked.
- `SUGGESTED_SINGLE` + user picks "github" — stage completes with `squash.messageAppended`, no CI poll, `squashSubStep === "applied_in_pr_body"`.
- `SUGGESTED_SINGLE` with missing marker block in PR body — falls back to `BLOCKED`.
- **Post-clarification fallback ordering** — given an ambiguous verdict:
  - Commit count decreased AND marker block present → classified as `SQUASHED_MULTI` (commit-count check wins).
  - Commit count unchanged AND marker block present → classified as `SUGGESTED_SINGLE`.
  - Neither → `BLOCKED`.
- Resume in each substate — `awaiting_user_choice` re-presents the choice; `ci_poll` resumes the poll loop; `applied_in_pr_body` advances straight to Stage 9.
- `BLOCKED` path — unchanged.

`src/step-parser.test.ts`:

- `parseVerdictKeyword` correctly distinguishes `SQUASHED_MULTI` from `SUGGESTED_SINGLE` when both appear in a free-form response (should resolve to ambiguous) and when only one is present.

`src/stage-e2e.test.ts` (or equivalent):

- Stage 9 surfaces the merge-confirm reminder when `squashSubStep === "applied_in_pr_body"`.
- Stage 9 includes the inline title/body preview and the PR URL in the merge-confirm screen for the same substate.

## Non-goals

- Changing the multi-commit squash flow (it stays as-is).
- Auto-detecting the repository's allowed merge strategies. The Stage 9 tip is phrased conditionally so the user is not misled if "Squash and merge" is disabled; a preflight check can be revisited later.
- Auto-merging from AgentCoop. Stage 9's existing merge confirmation stays manual.
- Launching a browser from AgentCoop (e.g. `gh pr view --web`). The PR URL is printed as text so terminals that support OSC 8 hyperlinks can open it, but AgentCoop itself does not spawn a browser.
- Adding a new `StepStatus` variant for the squash-suggestion state. The squash handler branches on the raw keyword from `parseVerdictKeyword` instead, keeping squash-specific concepts out of the shared status enum.

sehkone:
## Summary

When stage 8 (Squash) judges that the branch should be consolidated into a single commit, it no longer rewrites history and force-pushes. Instead, the agent writes the suggested title and body into a marker-delimited block in the PR body and the user is asked whether to:

- let the agent squash now (runs CI again), or
- apply the suggestion via GitHub's "Squash and merge" at merge time (no CI rerun).

The multi-commit path (`SQUASHED_MULTI`) is unchanged. Stage 8's verdict keywords change from `COMPLETED` / `BLOCKED` to `SQUASHED_MULTI` / `SUGGESTED_SINGLE` / `BLOCKED`. The squash handler parses the keyword directly rather than going through `mapResponseToResult`, keeping the three-way distinction out of the shared `StepStatus`. When the clarification retry still returns an ambiguous keyword, a deterministic fallback chain runs — commit-count decrease (detects a completed force-push), then PR-body suggestion-block presence, then `BLOCKED` — so a completed squash is never misclassified as a suggestion. The derived verdict is also emitted as a `pipeline:verdict` event so telemetry consumers see every verdict, not just parsed ones.

`RunState` gains an explicit `squashSubStep` (`planning` / `awaiting_user_choice` / `squashing` / `ci_poll` / `applied_in_pr_body`) so resume re-enters the stage at the correct sub-state. `RUN_STATE_VERSION` is bumped from 2 to 3 with a non-destructive default for older state files. Stage 9's merge-confirm screen prints a conditional tip, the suggested title and body, and the PR URL inline when the suggestion path is live.

Resuming from `squashing` no longer falls through to a fresh planning run: if the commit count has already collapsed to 1 the handler jumps straight to the CI poll; otherwise it re-sends only the follow-up squash prompt on the saved session. This prevents the extra force-push / CI cycle the feature was designed to avoid.

In-process retries read the persisted squash sub-step and agent-A session id through live getters, not the one-shot snapshots the pipeline passes through `StageContext`. Without that, a transient `ci_poll` failure followed by a retry could see `squashSubStep === undefined` and, with the branch already collapsed to a single commit, fall into the single-commit skip path — turning a recoverable CI failure into a false successful completion.

The SUGGESTED_SINGLE / `applied_in_pr_body` paths now validate the PR-body block strictly via `parseSquashSuggestionBlock` (markers present **and** a `**Title:**` line). A bare start marker or a block missing `**Title:**` / the end marker fails closed with `blocked` (or re-runs planning on resume) instead of completing with `squash.messageAppended` and leaving Stage 9 unable to render the inline preview from the same parser.

Stage 8 also persists the verdict turn's session id via `ctx.onSessionId` before transitioning to `awaiting_user_choice`. Adapters may surface a new session id on follow-up turns, so the verdict session is not guaranteed to match the planning session persisted earlier. Without persisting it, a resume where the user picks "agent squashes now" could re-send the follow-up on the older planning session rather than the exact conversation that drafted the PR-body suggestion.

New i18n keys: `squash.messageAppended`, `squash.singleChoicePrompt`, `squash.singleChoiceAgent`, `squash.singleChoiceGithub`, `squash.agentChoiceMissingSession`, `pipeline.mergeConfirmSquashTip`, `pipeline.suggestedSquashTitle`, `pipeline.suggestedSquashBody`, `pipeline.prUrl` (English + Korean). `docs/pipeline.md` is rewritten for Stage 8, including the new keyword contracts, the `SquashSubStep` state machine, the marker block contract, the post-clarification fallback chain, and the live-getter resume story. `CHANGELOG.md` gains an `[Unreleased]` section.

Closes #252

## Test plan

- [x] `pnpm run check` passes (Biome)
- [x] `pnpm run test` passes (1790 tests)
- [x] `pnpm run build` succeeds (tsc)
- [x] Stage 8 `SQUASHED_MULTI` path: agent force-pushes, CI poll runs, stage completes with `squash.completed`
- [x] Stage 8 `SUGGESTED_SINGLE` + user picks "agent": follow-up prompt is sent, force-push + CI poll run, stage completes with `squash.completed`
- [x] Stage 8 `SUGGESTED_SINGLE` + user picks "github": stage completes with `squash.messageAppended`, no CI poll, `squashSubStep === "applied_in_pr_body"`
- [x] Stage 8 `SUGGESTED_SINGLE` with missing marker block falls back to `BLOCKED`
- [x] Stage 8 `SUGGESTED_SINGLE` with malformed marker block (start marker only / missing `**Title:**`) fails closed without prompting the user
- [x] Stage 8 fallback chain treats malformed block as missing → `BLOCKED` (not `SUGGESTED_SINGLE`)
- [x] Resume from `awaiting_user_choice` with malformed block re-runs planning instead of re-presenting the user choice
- [x] Post-clarification fallback ordering: commit-count decrease wins over marker presence
- [x] Post-clarification fallback emits `pipeline:verdict` for the derived keyword (SQUASHED_MULTI / SUGGESTED_SINGLE / BLOCKED)
- [x] Resume in `awaiting_user_choice` re-presents the user choice without re-running the agent
- [x] Resume in `ci_poll` resumes the poll loop
- [x] Resume in `applied_in_pr_body` advances straight to Stage 9
- [x] Resume in `squashing` with commit count collapsed to 1 jumps straight to CI poll (no agent invocation)
- [x] Resume in `squashing` with count still > 1 and a saved session re-sends only the follow-up prompt (no planning prompt)
- [x] Resume in `squashing` with count still > 1 and no saved session falls back to a fresh planning run
- [x] In-process retry: live `savedSquashSubStep` getter re-enters `ci_poll` instead of falsely completing via single-commit skip
- [x] In-process retry: live `getSavedAgentSessionId` getter re-uses the persisted session even when `ctx.savedAgentASessionId` was cleared
- [x] Stage 8 `BLOCKED` path unchanged
- [x] \`parseVerdictKeyword\` disambiguates `SQUASHED_MULTI` vs `SUGGESTED_SINGLE` (single-keyword matches resolve; both-present resolves to ambiguous)
- [x] Stage 9 merge-confirm screen shows the conditional tip, inline title/body, and PR URL when `squashSubStep === "applied_in_pr_body"`
- [x] `RunState` migration: state files without `squashSubStep` load cleanly as `undefined`
- [x] Stage 8 persists the verdict turn's session id before `awaiting_user_choice`; resume + user "agent" choice uses the verdict session (not the planning session) for the follow-up

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,80 @@ This file documents recent notable changes to this project. The format of this
 file is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and
 this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Stage 8 (Squash) now offers a single-commit suggestion path: when the agent
+  judges that one commit is appropriate, it writes the suggested title and
+  body into a marker-delimited block in the PR body instead of force-pushing.
+  The user can then either let the agent perform the squash (which reruns CI)
+  or apply the suggestion via GitHub's "Squash and merge" at merge time
+  (which avoids the extra CI cycle).
+- Stage 9 (Done) merge-confirm screen now prints the suggested squash title,
+  body, and PR URL inline when the squash stage finished via the
+  PR-body suggestion path, so the user can copy-paste without opening the
+  browser.
+
+### Changed
+
+- Stage 8 verdict keywords are now `SQUASHED_MULTI` / `SUGGESTED_SINGLE` /
+  `BLOCKED` (previously `COMPLETED` / `BLOCKED`).  When the verdict is
+  ambiguous after a clarification retry, the handler runs a deterministic
+  fallback chain — commit-count decrease, then PR-body marker presence,
+  then BLOCKED.
+- `RunState` now persists a `squashSubStep` field tracking progress through
+  the squash stage's substates so resume can re-enter at the correct point.
+  `RUN_STATE_VERSION` bumped from 2 to 3 (the new field defaults to
+  `undefined` for older state files; no destructive migration).
+- Stage 9 merge-confirm screen now includes a conditional one-line tip
+  (`pipeline.mergeConfirmSquashTip`) when a squash suggestion is live in
+  the PR body.
+
+### Fixed
+
+- Resuming Stage 8 from the `squashing` substate no longer re-sends the
+  full planning prompt.  The handler now checks whether the squash
+  already landed (commit count collapsed to 1) and jumps straight to
+  the CI poll, or re-sends only the follow-up squash prompt on the
+  saved session — avoiding a spurious second force-push / CI cycle
+  that defeated the purpose of this feature.
+- `RunState` migration from v2 to v3 no longer re-applies the v1 → v2
+  stage 7 / 8 swap, which would have remapped existing v2 runs
+  persisted at stage 7 or 8 to the wrong handler on upgrade.  The
+  swap is now guarded on the source version actually being v1 /
+  unversioned.
+- Stage 8 no longer silently routes a user's "agent squashes now"
+  choice to the "apply via GitHub" completion when no agent session
+  is available.  It fails closed with a clear message so the user is
+  not misled about what happened.
+- Stage 8 in-process retries now read the live persisted squash
+  sub-step and agent session id on each handler invocation instead
+  of the startup snapshot.  Previously, a transient `ci_poll`
+  failure followed by a retry could observe `squashSubStep ===
+  undefined` and, with the branch already collapsed to a single
+  commit, fall into the single-commit skip path — turning a
+  recoverable CI failure into a false successful completion.
+- Stage 8's deterministic fallback chain (commit-count /
+  marker-presence) now emits the `pipeline:verdict` event for the
+  derived verdict.  Previously the event was only emitted when the
+  agent response parsed into a concrete keyword, so the fallback
+  branch silently skipped telemetry.
+- Stage 8 now validates the squash suggestion block strictly before
+  accepting `SUGGESTED_SINGLE` / `applied_in_pr_body`.  A bare start
+  marker or a block missing `**Title:**` / the end marker is treated
+  as malformed and fails closed (or re-runs planning on resume)
+  instead of completing with `squash.messageAppended` and leaving
+  Stage 9 unable to render the inline preview.
+- Stage 8 now persists the verdict turn's session id before
+  transitioning to `awaiting_user_choice`.  Adapters can surface a
+  new session id on follow-up turns, so the verdict session is not
+  guaranteed to match the planning session that was persisted
+  earlier.  Without this, a resume where the user picks "agent
+  squashes now" could re-send the follow-up on the older planning
+  session rather than the exact conversation that drafted the
+  PR-body suggestion.
+
 ## [0.1.0] - 2026-04-18
 
 ### Added
--- a/docs/pipeline.md
+++ b/docs/pipeline.md
@@ -92,7 +92,7 @@ Do not include any other commentary — just the keyword.
 | Test plan verdict | `FIXED` / `DONE` | `DONE` | `FIXED` → repeat |
 | PR creation check | `COMPLETED` / `BLOCKED` | `COMPLETED` | `BLOCKED` → user chooses |
 | CI findings review | _(no verdict keyword)_ | SHA unchanged | SHA changed → re-poll CI |
-| Squash check | `COMPLETED` / `BLOCKED` | `COMPLETED` | `BLOCKED` → user chooses |
+| Squash check | `SQUASHED_MULTI` / `SUGGESTED_SINGLE` / `BLOCKED` | `SQUASHED_MULTI` (CI poll), `SUGGESTED_SINGLE` (user choice) | `BLOCKED` → user chooses |
 | Reviewer verdict | `APPROVED` / `NOT_APPROVED` | `APPROVED` | `NOT_APPROVED` → repeat |
 | Author completion | `COMPLETED` / `BLOCKED` | `COMPLETED` | `BLOCKED` → user chooses |
 | Unresolved summary | `NONE` / `COMPLETED` | either | — |
@@ -135,6 +135,18 @@ The standard PR sync instructions appear in stages 5 through 8:
 > does. If the description is outdated or inaccurate, update it
 > using `gh pr edit --body "..."`. Keep the issue reference
 > (Closes #N or Part of #N) in the body.
+
+The squash stage extends this contract with a marker-delimited
+**squash suggestion block**.  When a single commit is the right
+shape for the branch, the agent writes the proposed title and body
+into the PR description between the markers
+`<!-- agentcoop:squash-suggestion:start -->` and
+`<!-- agentcoop:squash-suggestion:end -->` instead of force-pushing.
+The block is idempotent — re-runs replace any prior block between
+the same markers rather than stacking duplicates.  The Stage 9
+merge-confirm screen reads this block back so the user can apply
+the message via GitHub's "Squash and merge" without opening the
+browser.
 
 ### Issue-implementation reconciliation
 
@@ -1212,69 +1224,165 @@ avoiding re-execution of already-completed work within a round.
 ### Stage 8: Squash commits
 
 **Agent:** A\
-**Purpose:** Consolidate branch commits into one or a few
-meaningful commits. Skipped automatically if the branch has only
-one commit.
+**Purpose:** Decide whether the branch is best presented as a
+single squash commit (in which case the agent writes the suggested
+message into the PR body and lets GitHub's "Squash and merge"
+apply it at merge time) or several meaningful commits (rewrite
+history and force-push).  Skipped automatically if the branch has
+only one commit.
+
+The single-commit suggestion path was added to avoid wasting an
+extra CI cycle on a force-push that GitHub's "Squash and merge"
+button would perform anyway.
 
 **Prompt:**
 
-```text
-You are squashing commits for the following GitHub issue.
+The work prompt asks the agent to:
 
-## Repository
-- Owner: {owner}
-- Repo: {repo}
-- Branch: {branch}
-- Worktree: {worktree_path}
-
-## Issue #{number}: {title}
-
-{issue_body}
-
-## Instructions
-
-1. Before pushing, check whether the PR description still accurately
-   reflects the current code changes.  Run
-   `gh pr view --json body --jq .body` to read the current
-   description, then compare it against what the branch actually does.
-   If the description is outdated or inaccurate, update it using
-   `gh pr edit --body "..."`.  Keep the issue reference
-   (Closes #{number} or Part of #{number}) in the body.
-2. Review the commits after the base commit `{baseSha}` and
-   consolidate them into one or a few meaningful commits.  Only
-   commits introduced on this branch should be touched — do not
-   include commits from the base branch.  Use
-   `git reset --soft {baseSha}` followed by `git commit`, or an
-   interactive rebase — whichever is simpler.
-3. Write clear, concise commit messages that summarise the changes.
-   Do not include issue or PR numbers in the commit title.
-   Instead, reference the issue in the commit body using
-   `Closes #N` or `Part of #N`.
-4. Force-push the branch (`git push --force-with-lease`).
-```
+1. Sync the PR description (standard PR-sync instructions).
+2. Decide whether the work belongs in **one** commit or **several**.
+3. Branch on that decision:
+   - **Single commit appropriate:** do not rewrite history.  Draft
+     the squash title and body, then write them into the PR body
+     between the markers
+     `<!-- agentcoop:squash-suggestion:start -->` and
+     `<!-- agentcoop:squash-suggestion:end -->` (replacing any prior
+     block between the same markers).  Do not force-push.
+   - **Multiple commits appropriate:** consolidate the branch
+     commits — using `git reset --soft {baseSha}` + `git commit` or
+     interactive rebase — write clear messages, and force-push
+     (`git push --force-with-lease`).
 
 **Completion check:**
 
 ```text
-You have finished your squash attempt.  Please evaluate the result
-and respond with exactly one of the following keywords:
+You have finished your squash decision.  Respond with exactly one
+of the following keywords:
 
-- COMPLETED — if the commits were squashed and force-pushed
-- BLOCKED — if you could not squash and need user intervention
+- SQUASHED_MULTI — if you rewrote history into multiple meaningful
+  commits and force-pushed
+- SUGGESTED_SINGLE — if a single commit is appropriate and you
+  wrote the suggested title/body into the PR body marker block
+  (no force-push)
+- BLOCKED — if you could not complete either path and need user
+  intervention
 
 Do not include any other commentary — just the keyword.
 ```
 
-**Ambiguous response handling:** Same internal clarification
-retry pattern as stage 4 (Create PR).  If clarification also fails,
-the handler re-checks the branch commit count to verify whether the
-squash actually happened.  If the count decreased, the stage
-completes; otherwise it reports `BLOCKED`.
+The handler calls `parseVerdictKeyword` directly with these three
+keywords (rather than feeding them through the shared `KEYWORD_MAP`)
+because the three-way distinction is squash-specific.
 
-**Post-squash CI:** After a successful squash and force-push, the
-orchestrator polls CI and invokes Agent A to fix failures if
-needed (up to 3 internal fix attempts). The stage only completes
-when CI passes.
+**Verdict handling:**
+
+- `SQUASHED_MULTI` → poll CI after the force-push (existing
+  behaviour, up to 3 internal fix attempts) and finish with
+  `squash.completed` when CI passes.
+- `SUGGESTED_SINGLE` → verify a fully parseable suggestion block is
+  present in the PR body before asking the user.  "Parseable" means
+  both markers are present **and** `parseSquashSuggestionBlock`
+  succeeds (the block contains a `**Title:**` line).  A bare start
+  marker or a block missing `**Title:**`/the end marker fails
+  closed with `blocked` rather than completing as if the suggestion
+  were valid — Stage 9 reads the same block via
+  `parseSquashSuggestionBlock` to render the inline preview, so a
+  malformed block would leave the merge-confirm screen blank.  Once
+  validated, ask the user via `chooseSquashApplyMode`:
+  - **"Agent squashes now"** → send a follow-up prompt on the
+    same session telling the agent to perform the squash using
+    the message it already drafted, then run the post-squash CI
+    poll loop.
+  - **"Apply via GitHub"** → finish the stage with
+    `squash.messageAppended` and persist
+    `squashSubStep === "applied_in_pr_body"` so Stage 9 surfaces
+    the suggestion in the merge-confirm screen.  No CI rerun.
+- `BLOCKED` → existing blocked flow.
+
+**Ambiguous response handling:** Same internal clarification
+retry pattern as stage 4 (Create PR).  If clarification also
+fails, the handler runs a **deterministic fallback chain**:
+
+1. If the branch commit count **decreased** relative to the
+   snapshot taken at stage entry, treat as `SQUASHED_MULTI` (the
+   force-push has already happened — that hard-to-undo side
+   effect must be detected first).
+2. Else, fetch the PR body and check for a fully parseable
+   suggestion block (markers present **and**
+   `parseSquashSuggestionBlock` returns a value).  If present,
+   treat as `SUGGESTED_SINGLE`; a malformed block (e.g. only the
+   start marker) is treated as missing.
+3. Else, treat as `BLOCKED`.
+
+The order matters: checking the commit count first prevents a
+completed squash from being misclassified as a SINGLE suggestion
+when an earlier run left a stale marker block in the PR body.
+
+The derived verdict is emitted as a `pipeline:verdict` event just
+like a parsed-keyword verdict, so telemetry consumers see every
+verdict regardless of whether it came from the agent response or
+the fallback chain.
+
+**`SquashSubStep` state machine:**
+
+```text
+planning → verdict
+        ├── (SQUASHED_MULTI)   → ci_poll → done
+        ├── (SUGGESTED_SINGLE) → awaiting_user_choice
+        │       ├── (agent)  → squashing → ci_poll → done
+        │       └── (github) → applied_in_pr_body (stage done)
+        └── (BLOCKED)         → user chooses
+```
+
+`RunState.squashSubStep` persists the current substate so resume
+re-enters at the correct point:
+
+- `applied_in_pr_body` → stage already done; advance to Stage 9.
+- `awaiting_user_choice` → re-verify a parseable suggestion block is
+  still in the PR body (same strict check as the verdict path); if
+  present, re-present the user choice without re-invoking the agent.
+  If absent or malformed, fall back to `planning` rather than
+  re-presenting a choice the user could not act on.  If
+  the user then picks "agent squashes now" but no agent session is
+  available (neither the saved run state nor the current verdict
+  produced a session ID), the stage fails closed with `blocked`
+  rather than silently completing as if the user had picked
+  "github" — the user's choice must not be misrepresented.
+- `ci_poll` → resume the post-squash CI poll loop directly.
+- `squashing` → the user picked "agent squashes now" and the
+  follow-up prompt was sent, but we were interrupted before the
+  transition to `ci_poll`.  Resume runs a deterministic check
+  before doing anything expensive:
+  1. Re-count the branch commits.  If it collapsed to 1, the
+     squash already landed — jump straight to `ci_poll`.  This is
+     the common case and avoids a second force-push / CI cycle.
+  2. Otherwise, if a saved session is available, re-send **only**
+     the follow-up squash prompt on that session so the agent
+     continues the same conversation rather than restarting
+     planning.
+  3. As a last resort (no session persisted), fall back to a
+     fresh planning run.
+- `planning` (or absent) → run the agent prompt fresh (the work
+  prompt is idempotent — the agent will redo or confirm the same
+  decision).
+
+The stage reads the sub-step and the persisted agent-A session id
+via live getters on each handler invocation, so an in-process
+retry (e.g. after a transient `ci_poll` error) observes the
+values the previous iteration persisted rather than the startup
+snapshot.  Without this, a retry from `ci_poll` after the branch
+had already collapsed could fall into the single-commit skip path
+and falsely complete the stage.
+
+The stage also persists the **verdict turn's** session id (the one
+returned by `resolveVerdict`) before transitioning to
+`awaiting_user_choice`.  Adapters can surface a new session id on
+a follow-up turn, so the verdict session is not guaranteed to
+equal the planning session.  Persisting it ensures that a resume
+from `awaiting_user_choice` — where the user picks "agent
+squashes now" — re-sends the follow-up on the exact conversation
+that drafted the PR-body suggestion, rather than the earlier
+planning session.
 
 **Outcome handling:** `requiresArtifact: true` — if `BLOCKED`,
 only **Instruct** and **Halt** are offered.
--- a/src/i18n/en.ts
+++ b/src/i18n/en.ts
@@ -141,6 +141,12 @@ export const en: Messages = {
     `Pipeline for ${owner}/${repo}#${issue} completed.`,
   "pipeline.mergeConfirm":
     "Has the PR been merged? Confirm to clean up the worktree.",
+  "pipeline.mergeConfirmSquashTip":
+    "If this repo allows 'Squash and merge', the suggested commit " +
+    "message is in the PR body.",
+  "pipeline.suggestedSquashTitle": (title) => `Suggested title: ${title}`,
+  "pipeline.suggestedSquashBody": "Suggested body:",
+  "pipeline.prUrl": (url) => `PR: ${url}`,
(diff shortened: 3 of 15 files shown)


sehkone:
**[Reviewer Round 1]**

I think there’s still a gap in the `SquashSubStep` resume story. `src/run-state.ts:42` introduces `"squashing"` specifically so stage 8 can resume after the user chose “agent squashes now”, but `src/stage-squash.ts:426` explicitly treats `"squashing"` as “fall through to a fresh agent run.” If the process is interrupted after the follow-up squash prompt is sent but before `onSquashSubStep("ci_poll")` runs, a resume will re-send the full planning prompt instead of continuing the in-flight squash. That can re-trigger the extra force-push / CI cycle this issue is trying to avoid, and it also means the persisted `"squashing"` state never actually restores “the intermediate state cleanly” promised by the new state machine.

I think `"squashing"` needs its own resume path (for example, re-send the follow-up prompt on the saved session or deterministically detect that the squash already happened and jump to CI polling), or else the substep should be removed from the persisted enum/docs. I’d also add a regression test alongside `src/stage-squash.test.ts:906` that resumes from `"squashing"` and proves we don’t go back through the planning prompt.

sehkone:

sehkone:
**[Author Round 1]**

- **Fixed — dedicated resume path for `squashing`.** You're right: the previous fall-through would have re-sent the full planning prompt and could trigger another force-push / CI cycle, defeating the point of the feature. The handler now owns a real resume branch for `squashing` that runs before the single-commit skip check:
  1. Re-count the branch commits. If the count has collapsed to 1, the squash already landed — jump straight to the CI poll.
  2. Otherwise, if `ctx.savedAgentASessionId` is available, re-send **only** the follow-up squash prompt (`buildAgentSquashFollowupPrompt`) on that session so the agent continues the same conversation instead of restarting planning.
  3. If no session was persisted (rare — session IDs are persisted alongside `squashSubStep`), fall back to a fresh planning run as a last resort.
  See `src/stage-squash.ts:413-460`.
- **Fixed — regression tests added.** `src/stage-squash.test.ts` now covers the three branches explicitly:
  - `squashing` + commit count collapsed to 1 → no `invoke`, no `resume`, CI poll runs, `states` never contains `"planning"`.
  - `squashing` + count still > 1 + session → `resume` is called exactly once with the follow-up prompt; the planning prompt is asserted absent from the prompt sink; `invoke` is not called.
  - `squashing` + count still > 1 + no session → fall-through to `invoke` (the planning prompt).
- **Fixed — docs and CHANGELOG updated.** `docs/pipeline.md` (Stage 8 resume section) now documents the three-step resume contract for `squashing` instead of lumping it in with `planning`. `CHANGELOG.md` gains a `### Fixed` entry under `[Unreleased]`.

Verified locally: `pnpm run check`, `pnpm run test` (1773 tests), `pnpm run build` all pass.

sehkone:
**[Reviewer Round 2]**

Thanks — the dedicated `squashing` resume path looks fixed.

I found two remaining issues:

- `src/run-state.ts:189` still applies the stage-7/stage-8 swap to every state with `version < 3`. That means existing persisted runs from the current release (`version: 2`) will be remapped to the wrong handler on upgrade if they were saved at stage 7 or 8. The v1→v2 swap needs to stay conditional on the source version actually being v1/unversioned; the v2→v3 migration should just backfill `squashSubStep`. Please add a regression test in `src/run-state.test.ts` that loads a raw v2 state at stage 7/8 and proves the stage number is preserved.

- `src/stage-squash.ts:601` silently converts the user's `"agent"` choice into the GitHub path when no session ID is available. In that branch the stage completes with `squash.messageAppended` / `applied_in_pr_body` even though the agent never performed the squash, which violates the new choice contract and can mislead the user about what happened. This should fail closed (for example, return `blocked`) or explicitly recover another way, rather than auto-completing as if the user had picked GitHub. A focused test for the session-less `"agent"` path would help here too.

sehkone:

sehkone:
**[Author Round 2]**

- **Fixed — v2 → v3 migration no longer re-swaps stages 7/8.** You're right: `migrateRunState` applied the v1 → v2 swap for any `version < RUN_STATE_VERSION`, so bumping to v3 would have remapped existing v2 runs at stage 7 or 8 to the wrong handler. The swap is now guarded on the source version actually being v1 / unversioned (`migrated.version < 2`); the v2 → v3 step just reuses the `squashSubStep` backfill that `loadRunState` already performs. See `src/run-state.ts:189-219`. Added a regression test group `loadRunState — migration from v2` in `src/run-state.test.ts` that writes raw `version: 2` states at stage 7 and stage 8 and proves the stage number is preserved, plus a check that `squashSubStep` backfills to `undefined`.

- **Fixed — session-less "agent" choice now fails closed.** The previous branch in `askUserAndApply` silently routed a user's "agent squashes now" choice to the GitHub completion message when no session was available, which misrepresented what happened. The handler now returns `{ outcome: "blocked", message: t()["squash.agentChoiceMissingSession"] }` instead, clears `squashSubStep`, and skips the `applied_in_pr_body` transition entirely. See `src/stage-squash.ts:601-613`. New i18n key `squash.agentChoiceMissingSession` (English + Korean). Focused test added under `resume on saved substate` that drives the `awaiting_user_choice` resume path with no `savedAgentASessionId` and verifies the outcome is `blocked` with a session-related message and that `applied_in_pr_body` is never emitted.

- **Docs / changelog updated.** `docs/pipeline.md` notes the session-less fail-closed behavior under the `awaiting_user_choice` resume description. `CHANGELOG.md` gains two `### Fixed` entries for these two regressions.

Verified locally: `pnpm run check`, `pnpm run test` (1777 tests), `pnpm run build` all pass.

sehkone:
**[Reviewer Round 3]**

Thanks — the two Round 2 items look fixed to me (`migrateRunState()` no longer re-swaps v2 states, and the session-less `"agent"` choice now fails closed).

I found two remaining issues:

- `src/index.ts:581-585` captures `savedSquashSubStep` only once from the startup snapshot, and `src/pipeline.ts:701-719` still treats saved session IDs as one-shot inputs. That means an in-process Retry does not reuse the substep/session that Stage 8 just persisted. A concrete bad case is `SUGGESTED_SINGLE -> agent -> ci_poll`: `runCiPollAndFinish()` leaves `squashSubStep` at `"ci_poll"` on failure (`src/stage-squash.ts:646-665`), but a Retry re-enters the handler with no saved substep and, after the branch has already collapsed to one commit, falls into the single-commit skip path (`src/stage-squash.ts:470-477`) instead of resuming CI polling. That can let a transient CI-poll failure turn into a false successful `singleCommitSkip` completion. I think the squash stage needs to read the live persisted squash state/session on each retry (or have the pipeline refresh those fields between handler invocations), and this needs a regression test.

- The new deterministic fallback chain never emits `pipeline:verdict` for the derived keyword. Telemetry is only emitted inside `resolveVerdict()` when `parseVerdictKeyword()` returns a concrete keyword on the direct response or clarification (`src/stage-squash.ts:329-338`, `src/stage-squash.ts:366-377`). If both responses stay ambiguous and Stage 8 later classifies the result via commit-count / marker fallback (`src/stage-squash.ts:525-542`), the stage proceeds with `SQUASHED_MULTI`, `SUGGESTED_SINGLE`, or `BLOCKED` but no `pipeline:verdict` event is sent. The issue explicitly called out keeping that telemetry, so I think the fallback branch should emit the final keyword too, with a test around the ambiguous -> fallback path.

sehkone:

sehkone:
**[Author Round 3]**

- **Fixed — in-process retries now read live persisted squash state and session.** You're right: `src/index.ts:585` captured `savedSquashSubStep` once from `savedState` at startup, and `src/pipeline.ts:707-719` clears `savedAgentASessionId` after the first iteration of `runStage`'s inner loop.  After `SUGGESTED_SINGLE -> agent -> ci_poll`, a transient CI-poll failure followed by an in-process retry saw `squashSubStep === undefined` + no session, fell into the single-commit skip path at `src/stage-squash.ts:470-477`, and returned a false successful `squash.singleCommitSkip`.
  - `SquashStageOptions` now accepts a getter form for both: `savedSquashSubStep: SquashSubStep | (() => SquashSubStep | undefined)` and a new `getSavedAgentSessionId?: () => string | undefined`.  The handler resolves both on every invocation (`src/stage-squash.ts:107-123` for the option types, `:393-400` for the sub-step, `:393-398` + the new `resolveSavedSessionId()` helper for session reads).  Legacy plain-value callers still work.
  - `src/index.ts:581-598` switches to the getter form for both fields, reading `runState.squashSubStep` and `runState.agentA.sessionId` live.
  - Regression tests added under a new `live sub-step / session on in-process retry` describe block in `src/stage-squash.test.ts`:
    - `ci_poll resume re-enters via live getter after branch collapsed to 1` — asserts the getter is called, CI poll runs, and the single-commit skip message is NOT returned even though `countBranchCommits` returns 1.
    - `resume from squashing reuses session via getSavedAgentSessionId when ctx session is absent` — asserts `agent.resume` is called exactly once with the persisted session id while `ctx.savedAgentASessionId` is undefined.

- **Fixed — deterministic fallback chain now emits `pipeline:verdict`.** Previously, emission only happened inside `resolveVerdict()` when `parseVerdictKeyword()` returned a concrete keyword.  When both attempts stayed ambiguous and the fallback chain derived the verdict via commit-count / marker-presence, the stage proceeded with `SQUASHED_MULTI` / `SUGGESTED_SINGLE` / `BLOCKED` but no telemetry event went out.  The fallback branch in `src/stage-squash.ts:559-581` now emits the derived keyword with the last clarification response as `raw`.  Three new tests under `pipeline:verdict emission from deterministic fallback` (`src/stage-squash.test.ts`) cover each derived verdict.

- **Docs / changelog updated.** `docs/pipeline.md` notes the live-getter resume story and documents that the fallback chain emits `pipeline:verdict` like a parsed-keyword verdict.  `CHANGELOG.md` gains two new `### Fixed` entries.

Verified locally: `pnpm run check`, `pnpm run test` (1782 tests), `pnpm run build` all pass.

sehkone:
**[Reviewer Round 4]**

Thanks — the Round 3 retry-state and fallback-telemetry fixes look good to me.

I found one remaining issue:

- `src/stage-squash.ts:299-301`, `src/stage-squash.ts:564-566`, and `src/stage-squash.ts:593-604` currently treat the presence of the start marker alone as proof that the SINGLE path succeeded. That means a malformed PR-body block (for example, only the start marker, or a block missing `**Title:**` / the end marker) is still classified as `SUGGESTED_SINGLE`, can re-present the choice on resume, and can complete via `squash.messageAppended`. But Stage 9 parses the block more strictly in `src/index.ts:761-770`, so the preview can end up blank because there is no usable title/body to show. The issue contract here is stronger than “marker exists”: the agent is supposed to write the suggested title/body into the PR body. I think Stage 8 should validate `parseSquashSuggestionBlock(prBody)` before accepting `SUGGESTED_SINGLE` / `applied_in_pr_body`, and fail closed (or re-run planning on resume) when the block is malformed. A focused regression test for “marker present but suggestion unparsable” would help.

sehkone:

sehkone:

sehkone:
**[Author Round 4]**

- **Fixed — Stage 8 now validates the suggestion block via `parseSquashSuggestionBlock` instead of marker presence.** You're right: `hasSuggestionMarker` only checked for the start marker, so a malformed block (start marker only, missing `**Title:**`, or missing the end marker) was still classified as `SUGGESTED_SINGLE` even though Stage 9 — which uses the strict `parseSquashSuggestionBlock` to render the inline preview at `src/index.ts:761-770` — would have nothing to show. The check is now centralised in `hasValidSuggestionBlock(prBody)` (`src/stage-squash.ts:299-315`), which wraps `parseSquashSuggestionBlock(prBody) !== undefined`. It is applied in all three gates:
  - `SUGGESTED_SINGLE` verdict branch (`src/stage-squash.ts:600-614`) — fails closed with `blocked` rather than calling `askUserAndApply` when the block is malformed.
  - Deterministic post-clarification fallback chain (`src/stage-squash.ts:564-576`) — a malformed block is treated as missing, so the verdict resolves to `BLOCKED` instead of `SUGGESTED_SINGLE`.
  - Resume from `awaiting_user_choice` (`src/stage-squash.ts:503-510`) — falls back to a fresh planning run instead of re-presenting a choice the user could not act on.

- **Tests added — `src/stage-squash.test.ts`.** A new `malformed suggestion block` describe block uses `test.each` over two malformed bodies (`start marker only` and `both markers but no **Title:**`) to cover all three gates:
  - `verdict SUGGESTED_SINGLE + malformed → blocked, no user choice` — asserts the outcome is `blocked`, `chooseSquashApplyMode` is never called, and neither `applied_in_pr_body` nor `awaiting_user_choice` is emitted.
  - `fallback chain + malformed → BLOCKED (not SUGGESTED_SINGLE)` — asserts the derived verdict is `BLOCKED` and `pipeline:verdict` carries that keyword.
  - `resume awaiting_user_choice + malformed → re-runs planning, no user choice` — asserts `agent.invoke` is called and `chooseSquashApplyMode` is not.

- **Docs / changelog updated.** `docs/pipeline.md` now spells out the strict-validation contract under both the `SUGGESTED_SINGLE` verdict description and the fallback chain step (and under `awaiting_user_choice` in the resume section). `CHANGELOG.md` gains a new `### Fixed` entry under `[Unreleased]` covering this behaviour change.

Verified locally: `pnpm run check`, `pnpm run test` (1788 tests), `pnpm run build` all pass.

sehkone:
**[Reviewer Round 5]**

Thanks — the malformed-block validation change from Round 4 looks good to me.

I found one remaining issue:

- `src/stage-squash.ts:538`, `src/stage-squash.ts:551`, `src/stage-squash.ts:503`, and `src/stage-squash.ts:661` still leave the resume path a bit shaky for `awaiting_user_choice`. The handler persists the planning turn’s session id (`squashResult.sessionId`) before calling `resolveVerdict()`, but it never persists the later verdict/clarification session id that `resolveVerdict()` returns. On the same invocation that’s fine because `askUserAndApply()` uses the local `verdictSessionId`, but after an interruption/resume the stage falls back to `getSavedAgentSessionId()` / `ctx.savedAgentASessionId`, which still points at the older planning session. That means a resumed “agent squashes now” choice is not guaranteed to continue the exact conversation that produced `SUGGESTED_SINGLE` and drafted the PR-body message. This matters because the adapters do allow resume turns to surface a different session id (for example `src/claude-adapter.ts:107-109` updates from the result event). I think Stage 8 should persist the latest verdict session id before entering `awaiting_user_choice`, and there should be a regression test for “planning session id != verdict session id, resume from `awaiting_user_choice`, user picks agent → follow-up resumes the verdict session”. Right now the resume tests in `src/stage-squash.test.ts:926` and nearby only cover the stable-session case.

sehkone:

sehkone:
**[Author Round 5]**

- **Fixed — Stage 8 now persists the verdict session id before entering `awaiting_user_choice`.** You're right: `squashResult.sessionId` (the planning session) was persisted via `ctx.onSessionId` at `src/stage-squash.ts:538-540`, but the verdict session returned by `resolveVerdict()` was never persisted. On the same invocation that was masked because `askUserAndApply` uses the local `verdictSessionId`, but after an interruption/resume from `awaiting_user_choice` the handler falls back to `getSavedAgentSessionId()` / `ctx.savedAgentASessionId` — still pointing at the older planning session. A resumed "agent squashes now" choice would then re-send the follow-up on that earlier conversation rather than the one that drafted the PR-body suggestion. The handler now calls `ctx.onSessionId?.("a", handle.sessionId)` immediately after `resolveVerdict` returns successfully (see `src/stage-squash.ts:568-581`), so the persisted agent-A session id is the verdict session before any transition to `awaiting_user_choice`.

- **Tests added — `src/stage-squash.test.ts` (`verdict session id persistence` describe block):**
  - `SUGGESTED_SINGLE persists verdict session (distinct from planning) before awaiting choice` — drives a single in-process run where `invoke` returns `sess-planning` and the verdict resume surfaces `sess-verdict`. Asserts `onSessionId` was called with `sess-verdict` before the user-choice prompt, and that the subsequent `resume` follow-up is sent on `sess-verdict` (not `sess-planning`).
  - `resume from awaiting_user_choice + user picks agent resumes the verdict session id via getSavedAgentSessionId` — simulates the persisted state from the first case: `savedSquashSubStep === "awaiting_user_choice"`, `getSavedAgentSessionId()` returns `sess-verdict`, user picks "agent". Asserts `resume` is called exactly once with `sess-verdict` and `invoke` is never called.

- **Docs / changelog updated.** `docs/pipeline.md` now notes, under the Stage 8 resume description, that the stage persists the verdict turn's session id before transitioning to `awaiting_user_choice`, and why. `CHANGELOG.md` gains a new `### Fixed` entry under `[Unreleased]` covering this behavior change.

Verified locally: `pnpm run check`, `pnpm run test` (1790 tests), `pnpm run build` all pass.

sehkone:
**[Reviewer Round 6]**

I re-checked the Round 5 item and the updated diff against the issue.

- The verdict-session persistence gap looks fixed. `createSquashStageHandler` now persists `handle.sessionId` immediately after `resolveVerdict()` returns, before any transition into `awaiting_user_choice`, so a resumed "agent squashes now" path continues the verdict conversation rather than the earlier planning session (`src/stage-squash.ts:568`).
- The regression coverage is in place for both the in-process case and the resumed `awaiting_user_choice` case (`src/stage-squash.test.ts:1138`, `src/stage-squash.test.ts:1197`).
- I also spot-checked the surrounding wiring and it matches the issue: the v1→v2 stage swap migration is now correctly guarded during the v2→v3 upgrade (`src/run-state.ts:201`), Stage 9 renders the squash hint/preview through the new hook (`src/pipeline.ts:1125`, `src/index.ts:759`), and the docs/changelog are in sync.

No remaining concerns from my side.

sehkone:

PR #257, opened 2026-04-18
Fix TokenBar text collapse in vertical layout (#256)

### LINKED ISSUE #256: TokenBar: Agent A/B rows show empty boxes in vertical layout
## Problem

When the TUI is switched to vertical layout (Ctrl+L), the TokenBar at the bottom renders only the bordered rectangles — the per-agent text (`A (Claude): ... in / ... out` and `B (Codex): ... in / ... out`) disappears. In horizontal (row) layout the bar renders correctly.

## Repro

1. Start agentcoop and run any pipeline so token usage accumulates.
2. Press Ctrl+L to switch to vertical layout.
3. Observe the TokenBar area above the StatusBar.

Expected: two stacked boxes, each showing the agent's token counts.
Actual: two stacked boxes with only top/bottom borders; no text inside.

## Root cause

`src/ui/TokenBar.tsx` applies `flexGrow={1}` and `flexBasis={0}` to each inner box unconditionally. In row layout the main axis is horizontal, so this split the width evenly between the two agents — the intended effect. In column layout the main axis is vertical, so `flexBasis={0}` sets each box's initial height to 0; when the TokenBar sits inside a bounded-height parent (as it does in `App.tsx`, which gives the root a fixed `height={terminalHeight}` and distributes rows via flex), `flexGrow={1}` then stretches each box to fill available space and the `height={3}` prop that was meant to reserve one content row between the borders is effectively ignored. The content row collapses to 0, leaving only the two border rows.

The bug is specific to the bounded-parent case. A standalone `TokenBar` rendered without a bounded-height ancestor still shows its text correctly — which is why the existing column test does not catch it.

Confirmed by a render with `layout="column"`, `contentWidth=76`, inside a `<Box height={20} flexDirection="column">` parent: the frame contains only `┌─…─┐` / `└─…─┘` pairs, no text rows.

## Suggested fix

Apply the `flexGrow` / `flexBasis` equal-sizing trick only when `layout === "row"`. In column layout, let each inner box take its intrinsic (or explicit `height`) size and stretch on the cross axis (width) naturally.

A regression test should render `TokenBar` with `layout="column"`, `contentWidth` set, **and** a bounded-height parent, and assert that both agents' text strings appear in the frame. The existing column test at `src/ui/components.test.tsx:2220` only checks that A and B end up on different lines, which passes even when both boxes are empty borders.

sehkone:
## Summary

In column (vertical) layout, `TokenBar` applied `flexGrow={1}` and `flexBasis={0}` to each agent box unconditionally. When the bar sits inside a bounded-height parent (as it does in `App.tsx`), the main axis is vertical, so `flexBasis={0}` zeros each box's initial height and `flexGrow={1}` stretches the borders, overriding the explicit `height={3}`. The content row collapses and only the borders render.

The fix restricts the flex equal-sizing trick to `layout === "row"`. In column layout each inner box keeps its explicit height and stretches only on the cross axis (width).

A regression test renders `TokenBar` with `layout="column"`, `contentWidth=76`, inside a `<Box height={20} flexDirection="column">` parent and asserts both agents' text strings appear in the frame. The previous column test only checked that A and B landed on different lines, which passes even when both boxes are empty borders.

Closes #256

## Test plan

- [x] `pnpm vitest run` — full suite passes (1747 tests)
- [x] `pnpm biome check` — no issues
- [x] New test: TokenBar in column layout inside bounded-height parent shows both agents' text (`components.test.tsx:2246`)
- [x] Column layout (Ctrl+L): verified programmatically via the new regression test, which reproduces the App.tsx bounded-height parent scenario and asserts both `Agent A (author)` and `Agent B (reviewer)` strings render in the frame
- [x] Row layout: verified programmatically via the existing `renders two boxes side by side in row layout` test (`components.test.tsx:2195`), which continues to pass — row rendering is unchanged

### DIFF
--- a/src/ui/TokenBar.tsx
+++ b/src/ui/TokenBar.tsx
@@ -139,12 +139,14 @@ export function TokenBar({
       : textB;
 
   const boxHeight = contentWidth !== undefined ? 3 : undefined;
+  // In column layout, flexBasis={0}+flexGrow={1} would collapse the explicit
+  // `height` inside a bounded-height parent, leaving only borders.
+  const mainAxisSizing = layout === "row" ? { flexGrow: 1, flexBasis: 0 } : {};
 
   return (
     <Box flexDirection={layout} flexShrink={0}>
       <Box
-        flexGrow={1}
-        flexBasis={0}
+        {...mainAxisSizing}
         borderStyle="single"
         borderColor="gray"
         paddingX={1}
@@ -154,8 +156,7 @@ export function TokenBar({
         <Text>{displayA}</Text>
       </Box>
       <Box
-        flexGrow={1}
-        flexBasis={0}
+        {...mainAxisSizing}
         borderStyle="single"
         borderColor="gray"
         paddingX={1}
--- a/src/ui/components.test.tsx
+++ b/src/ui/components.test.tsx
@@ -2242,6 +2242,29 @@ describe("TokenBar layout prop", () => {
     expect(agentBIdx).toBeGreaterThanOrEqual(0);
     expect(agentAIdx).not.toBe(agentBIdx);
   });
+
+  test("shows agent text in column layout inside a bounded-height parent", async () => {
+    const emitter = new PipelineEventEmitter();
+    const { lastFrame } = render(
+      <Box height={20} flexDirection="column">
+        <TokenBar emitter={emitter} layout="column" contentWidth={76} />
+      </Box>,
+    );
+
+    emitter.emit("agent:usage", {
+      agent: "a",
+      usage: { inputTokens: 1000, outputTokens: 500, cachedInputTokens: 0 },
+    });
+    emitter.emit("agent:usage", {
+      agent: "b",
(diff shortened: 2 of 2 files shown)


sehkone:
**[Reviewer Round 1]** No blocking issues found.

The fix in `src/ui/TokenBar.tsx:141` looks targeted and correct: the equal-width flex sizing only makes sense when the main axis is horizontal, and removing `flexGrow`/`flexBasis={0}` in column layout avoids the bounded-height collapse that `App` creates at `src/ui/App.tsx:463` and `src/ui/App.tsx:492`.

The new regression test in `src/ui/components.test.tsx:2246` is the right test for this bug. It reproduces the missing ingredient from the previous column-layout test — a bounded-height parent plus `contentWidth` — and asserts on the actual user-visible failure mode by checking that both agent labels render. That would have failed before this change, so it meaningfully protects the fix.

I did not spot any surprising side effects or unrelated churn in the patch.

sehkone:

PR #258, opened 2026-04-18
Make status bar issue/PR references clickable (#253)

### LINKED ISSUE #253: Make issue/PR references in the status bar clickable
## Problem

The status bar shows the issue reference (`aicers/{repo}#{issue}: {title}`) on line 1 and can display other metadata on line 2, but nothing in the status bar is clickable. Users who want to jump to the issue or PR in GitHub must context-switch to another tool (browser history, `gh browse`, etc.) even though AgentCoop already knows the exact URLs.

## Proposal

A status bar UX change centered on [src/ui/StatusBar.tsx](../blob/main/src/ui/StatusBar.tsx), with supporting wiring elsewhere (see "Implementation notes" for the `prNumber` propagation path):

### 1. Wrap the existing issue reference with an OSC 8 terminal hyperlink

Apply OSC 8 escape sequences around the `issueLine` rendered at [src/ui/StatusBar.tsx:285-308](../blob/main/src/ui/StatusBar.tsx#L285-L308) so that Cmd/Ctrl-click opens `https://github.com/{owner}/{repo}/issues/{issueNumber}` in the default browser. Terminals that support OSC 8 (iTerm2, WezTerm, Ghostty, Kitty, modern VS Code, Windows Terminal, Alacritty with newer versions, etc.) render it as a clickable link; older terminals render the text unchanged, so the change is a safe upgrade with no regressions.

The escape sequence format is:

```
\x1b]8;;<URL>\x07<text>\x1b]8;;\x07
```

**Important:** wrap *after* truncation in [src/ui/StatusBar.tsx:305-308](../blob/main/src/ui/StatusBar.tsx#L305-L308). Wrapping before truncation risks cutting the escape sequence in half. `string-width` strips ANSI escapes before measuring, so width calculations for Line 1's right-aligned elapsed-time layout remain correct.

### 2. Add a `PR: #{prNumber}` segment on Line 2

Add a new info segment alongside `Base:`, `Stage:`, `Last:`, `Completed:`, `Layout:` at [src/ui/StatusBar.tsx:313-326](../blob/main/src/ui/StatusBar.tsx#L313-L326):

- Shown only when `prNumber` is defined (i.e., once `prNumber` is known — see the propagation path in "Implementation notes" below). The component already receives `owner`, `repo`, `issueNumber` via props; extend the interface with `prNumber?: number`.
- Position: between the `Base:` segment and the `Stage:` segment so both identifiers (base commit + PR) visually group together.
- Drop priority: `1`, matching `Base:`. The required `Stage:` segment (priority `0`) is never dropped, and `PR:` falls out before it on narrow terminals.
- Wrap the rendered text with the same OSC 8 escape pointing at `https://github.com/{owner}/{repo}/pull/{prNumber}`.
- New i18n key: `statusBar.pr: (prNumber) => "PR: #${prNumber}"`.

Example output:

```
aicers/bootroot#517: Auto-read admin DSN from ca.json in `rotate db`
Base: 06c69e9  |  PR: #523  |  Stage 9: Done  |  Last: completed  |  ...
```

Both references become clickable destinations so the user can jump between the issue and the PR without leaving the terminal.

### A small helper

A `wrapTerminalHyperlink(url, text)` helper (local to `StatusBar.tsx` or in a new `src/ui/terminal-hyperlink.ts`) keeps the escape sequence in one place and makes the wrapping call sites readable:

```ts
export function wrapTerminalHyperlink(url: string, text: string): string {
  const OSC = "\x1b]8;;";
  const BEL = "\x07";
  return `${OSC}${url}${BEL}${text}${OSC}${BEL}`;
}
```

### Sourcing `prNumber` — explicit propagation path

`RunState.prNumber` is populated inside the orchestrator's `onStageTransition` callback the first time the pipeline enters stage 5 or later. Currently this value is written to disk but is **not** exposed to the Ink UI tree — neither `App.tsx` nor `StatusBar` receive it today. The change needs a concrete propagation path:

1. Emit a new pipeline event, e.g. `pr:resolved` with payload `{ prNumber: number }`, from the orchestrator right after `findPrNumber` succeeds in `onStageTransition`. Add the event type to `PipelineEventEmitter` alongside the existing `stage:enter` / `stage:exit` / `review:posted` events. (The resume-path `findPrNumber` calls earlier in bootstrap run before the `PipelineEventEmitter` is constructed, so they cannot emit; the resume case is covered by step 2's initial-prop seeding.)
2. In [src/ui/App.tsx](../blob/main/src/ui/App.tsx), subscribe to `pr:resolved` on mount (mirroring how `StatusBar` already subscribes to stage events), store the number in App state, and pass it as a new `prNumber?: number` prop to `StatusBar`. On resume, the saved `RunState.prNumber` that `App` receives via props (e.g. as `initialPrNumber`) seeds the initial state so the segment renders without waiting for a stage transition.
3. `StatusBar` renders the segment conditionally on the prop. No direct coupling between `StatusBar` and `RunState`.

This keeps the UI's data sources consistent (everything flows through `PipelineEventEmitter` or initial props) and avoids the alternative of polling `RunState` each render.

### i18n

- `statusBar.pr` — new key for the PR segment label.
- No copy change for the issue line; the text stays the same, only the wrapping changes.

## Non-goals

- Launching the browser via `gh pr view --web` or similar. OSC 8 is passive — the terminal does the work, so AgentCoop does not need to spawn processes.
- Detecting whether the user's terminal supports OSC 8. Non-supporting terminals render the text as-is with a small amount of invisible garbage text only on very old emulators; the trade-off is worth it.
- Applying OSC 8 to other non-URL content (commit SHAs, stage names). Only stable canonical GitHub URLs are in scope.

## Tests

Add to the existing `describe("StatusBar", ...)` block at [src/ui/components.test.tsx:869](../blob/main/src/ui/components.test.tsx#L869) (there is no separate `StatusBar.test.tsx` in this repo):

- Line 1 output contains the OSC 8 escape sequence wrapping `https://github.com/{owner}/{repo}/issues/{issueNumber}`.
- `PR: #{n}` segment appears when `prNumber` is set, and the rendered text contains the OSC 8 escape sequence wrapping `https://github.com/{owner}/{repo}/pull/{prNumber}`.
- `PR: #{n}` segment is absent when `prNumber` is undefined.
- Truncation still fits inside the width budget (escape sequences are zero-width for display).

Also add coverage for the new event wiring in `App.tsx` — emit `pr:resolved` on a test emitter and assert that `StatusBar` receives an updated `prNumber` prop (or that the rendered output gains the PR segment). The right home for this is likely [src/ui/App.test.tsx](../blob/main/src/ui/App.test.tsx).

### `CHANGELOG.md`

The file does not yet have an `## [Unreleased]` section (current top entry is `## [0.1.0] - 2026-04-18`). Create the section following [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) conventions, then under `### Added` note the clickable issue/PR references and the new `PR: #{n}` status bar segment.

## Difficulty

Closer to `S~M` than `good first issue`. A first-time contributor would need to be comfortable with Ink rendering, OSC 8 / ANSI escape interaction with `string-width`, and the pipeline event flow to wire `prNumber` through cleanly.

sehkone:
## Summary

- Wrap the status bar's issue reference on line 1 with an OSC 8 terminal hyperlink so supporting terminals (iTerm2, WezTerm, Ghostty, Kitty, modern VS Code, Windows Terminal, Alacritty, etc.) render it as a clickable link to the GitHub issue page. Older terminals render the text unchanged.
- Add a `PR: #{n}` segment on line 2 (between `Base:` and `Stage:`, drop priority `1`) once the pull-request number is known, also wrapped with an OSC 8 hyperlink pointing at the PR page.
- Propagate the PR number from the orchestrator to the UI via a new `pr:resolved` pipeline event, and seed it on resume from the persisted `RunState.prNumber` via a new `initialPrNumber` prop on `App`.
- Add a `wrapTerminalHyperlink(url, text)` helper to keep the OSC 8 escape sequence in one place.
- Add `statusBar.pr` i18n key (en/ko).
- `CHANGELOG.md`: add `## [Unreleased]` / `### Added` notes.

Closes #253

## Test plan

- [x] `pnpm biome check` passes.
- [x] `pnpm vitest run` passes (includes new coverage in `src/ui/components.test.tsx` and `src/ui/App.test.tsx`).
- [x] `pnpm build` succeeds.
- [x] Line 1 output contains the OSC 8 escape sequence wrapping `https://github.com/{owner}/{repo}/issues/{issueNumber}`.
- [x] `PR: #{n}` segment appears when `prNumber` is set and wraps `https://github.com/{owner}/{repo}/pull/{prNumber}`.
- [x] `PR: #{n}` segment is absent when `prNumber` is undefined.
- [x] Truncation still fits inside the width budget (escape sequences are zero-width for display).
- [x] `pr:resolved` event updates `StatusBar` and `initialPrNumber` seeds it on resume.
- [x] Manual smoke test in an OSC 8-capable terminal: Cmd/Ctrl-click on the issue reference and `PR: #{n}` both open the correct GitHub URLs.
- [x] Smoke test in an older terminal without OSC 8 support: text renders as plain text without visible artifacts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- Status bar now wraps the issue reference on line 1 with an OSC 8 terminal
+  hyperlink so supporting terminals (iTerm2, WezTerm, Ghostty, Kitty, modern
+  VS Code, Windows Terminal, Alacritty, etc.) render it as a clickable link
+  pointing at the GitHub issue page. Older terminals render the text
+  unchanged.
+- Status bar line 2 now includes a `PR: #{n}` segment (between `Base:` and
+  `Stage:`) once the pull-request number is known, also wrapped with an
+  OSC 8 hyperlink pointing at the PR page.
 - Stage 8 (Squash) now offers a single-commit suggestion path: when the agent
   judges that one commit is appropriate, it writes the suggested title and
   body into a marker-delimited block in the PR body instead of force-pushing.
--- a/src/i18n/en.ts
+++ b/src/i18n/en.ts
@@ -196,6 +196,7 @@ export const en: Messages = {
     `Stage ${number}: ${name} (round ${round})`,
   "statusBar.last": (outcome) => `Last: ${outcome}`,
   "statusBar.base": (sha) => `Base: ${sha}`,
+  "statusBar.pr": (prNumber) => `PR: #${prNumber}`,
   "statusBar.completed": (selfCheckCount, reviewCount) =>
     `Completed: self-check \u00d7${selfCheckCount}, review \u00d7${reviewCount}`,
   "statusBar.layout": (mode) => `Layout: ${mode}`,
--- a/src/i18n/index.test.ts
+++ b/src/i18n/index.test.ts
@@ -78,6 +78,7 @@ describe("catalogs are complete", () => {
       "statusBar.stage": [1, "Implement"],
       "statusBar.stageRound": [3, "Self-check", 1],
       "statusBar.last": ["ok"],
+      "statusBar.pr": [42],
       "statusBar.completed": [2, 3],
       "stageError.maxTurns": [" (stage 1)"],
       "stageError.inactivityTimeout": [" (stage 2)"],
--- a/src/i18n/ko.ts
+++ b/src/i18n/ko.ts
@@ -231,6 +231,7 @@ export const ko: Messages = {
     `${number}\uB2E8\uACC4: ${name} (\uB77C\uC6B4\uB4DC ${round})`,
   "statusBar.last": (outcome) => `\uC774\uC804: ${outcome}`,
   "statusBar.base": (sha) => `\uAE30\uC900: ${sha}`,
+  "statusBar.pr": (prNumber) => `PR: #${prNumber}`,
   "statusBar.completed": (selfCheckCount, reviewCount) =>
     `\uC644\uB8CC: \uC140\uD504 \uCCB4\uD06C \u00d7${selfCheckCount}, \uB9AC\uBDF0 \u00d7${reviewCount}`,
   "statusBar.layout": (mode) => `\uB808\uC774\uC544\uC6C3: ${mode}`,
--- a/src/i18n/messages.ts
+++ b/src/i18n/messages.ts
@@ -187,6 +187,7 @@ export interface Messages {
   ) => string;
   "statusBar.last": (outcome: string) => string;
   "statusBar.base": (sha: string) => string;
+  "statusBar.pr": (prNumber: number) => string;
   "statusBar.completed": (
     selfCheckCount: number,
     reviewCount: number,
--- a/src/index.ts
+++ b/src/index.ts
@@ -868,6 +868,9 @@ try {
       // Update PR number when entering stage 5+ (PR should exist by then).
       if (stageNumber >= 5 && runState.prNumber === undefined) {
         runState.prNumber = findPrNumber(owner, repo, wt.branch);
+        if (runState.prNumber !== undefined) {
+          emitter.emit("pr:resolved", { prNumber: runState.prNumber });
+        }
       }
       // Track review round.
       if (stageNumber === 7) {
@@ -923,6 +926,7 @@ try {
       notifications,
       initialSelfCheckCount: runState.selfCheckCount,
       initialReviewCount: runState.reviewCount,
+      initialPrNumber: runState.prNumber,
     });
   });
 
--- a/src/pipeline-events.ts
+++ b/src/pipeline-events.ts
@@ -25,6 +25,10 @@ export interface ReviewPostedEvent {
   round: number;
 }
 
+export interface PrResolvedEvent {
+  prNumber: number;
+}
+
 export interface AgentInvokeEvent {
   agent: "a" | "b";
   type: "invoke" | "resume";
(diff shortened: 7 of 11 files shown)


sehkone:
**[Reviewer Round 1]** I reviewed the implementation against #253 and I do not see any blocking issues.

- `src/ui/StatusBar.tsx:320` truncates the issue text before applying the OSC 8 wrapper, which is the key correctness detail for line 1. The new PR segment at `src/ui/StatusBar.tsx:334` is also wrapped with the canonical PR URL and placed in the requested spot between `Base:` and `Stage:`.
- `src/ui/App.tsx:301` and `src/index.ts:869` wire `prNumber` into the Ink tree cleanly: the event path updates the live UI, and `initialPrNumber` covers the resume case without making `StatusBar` reach into `RunState`.
- The new coverage in `src/ui/components.test.tsx:1274` and `src/ui/App.test.tsx:81` exercises the behavior that matters here (issue hyperlink, conditional PR segment, truncation behavior, and `pr:resolved` propagation), and these are the sorts of assertions that would have caught regressions in this feature.

One small nit: the drop-order comment at `src/ui/StatusBar.tsx:329` still says `base (1)` even though `PR` now shares that priority too. Not blocking, but it would be worth updating so the comment stays aligned with the code.

sehkone:

sehkone:
**[Reviewer Unresolved Round 1]**
- `src/ui/StatusBar.tsx:329` — the drop-order comment still mentions only `base (1)`, but `PR` now shares the same drop priority. This is non-blocking, but the comment is still out of sync with the implementation.

Issue #262, opened 2026-04-18
Make suggested squash title and body easy to copy


sehkone:
Umbrella issue — tracked via sub-issues.

## Context

When stage 8 takes the single-commit suggestion path (SUGGESTED_SINGLE verdict + user picks "apply via GitHub at merge time"), the suggested commit title and body surface in **two places**, and in both the user's next action is the same: copy the title into GitHub's "Squash and merge" title input, copy the body into the body input.

1. The **PR body on GitHub** — written by the agent into a marker-delimited block. Currently `**Title:** <title>` / `**Body:**` plain Markdown — no one-click copy button, and the body is re-interpreted as Markdown so characters like `*`, `` ` ``, `_`, leading `#` can be mangled.
2. The **TUI merge-confirm screen** (stage 9) — rendered inline as a tip. Currently `Suggested title: <title>` inline label + plain multi-line body — triple-click picks up the label prefix, and body blends into surrounding prose. No keypress copy.

## Goal

Both surfaces should let the user copy the title and body with a single action (click on GitHub, keypress in the TUI), with the body preserved verbatim.

## Sub-issues

- #263 — Part A: render the suggestion block in the PR body as fenced code blocks (GitHub's auto copy-button + literal body preservation).
- #264 — Part B: split label from value in the TUI merge-confirm preview so triple-click selects just the value, and the layout has space for Part C's hotkey hints.
- #265 — Part C: add OSC 52 hotkey copy (`t`/`b`) in the TUI so the keypress path mirrors GitHub's click-to-copy.

Close when all three sub-issues are closed.

## Out of scope

- Any change to how the agent decides SQUASHED_MULTI vs SUGGESTED_SINGLE.
- Mouse-click handling in the TUI (breaks terminal's own select-to-copy across most setups). Hotkey + OSC 52 is the right primitive.

sehkone:
Done

PR #266, opened 2026-04-19
Split label from value in squash merge preview (#264)

### LINKED ISSUE #264: Split label from value in TUI squash merge-confirm preview
Sub-issue of #262 (Part B).

## Problem

The stage 9 merge-confirm screen composes a tip with the suggested title and body inline in [src/pipeline.ts:1134-1145](https://github.com/aicers/agentcoop/blob/main/src/pipeline.ts#L1134-L1145):

```
If this repo allows 'Squash and merge', the suggested commit ...
Suggested title: <title>
Suggested body:
<body line 1>
<body line 2>
...
https://github.com/.../pull/N
```

- **Title label is inline with the value.** Triple-clicking the title line selects `Suggested title: ` along with the value, so the user has to manually trim before pasting.
- **Body blends into surrounding prose.** Body lines sit directly between the `Suggested body:` label and the PR URL / merge prompt, with no clean selection boundary.

## Proposal

Split the label from the value so each value sits on its own line with no prefix, separated by blank lines from surrounding content:

```
If this repo allows 'Squash and merge', the suggested commit ...

Suggested title:
<title>

Suggested body:
<body>

https://github.com/.../pull/N
```

- Triple-clicking the title line selects exactly the title — no prefix to trim.
- Blank lines between sections give a clean visual boundary for multi-line body selection.
- Pure string-composition change — zero risk to existing terminal behavior.

This also creates the visual space that #262's Part C (hotkey copy hints) needs to sit in cleanly.

## Scope

- Update the `askMerge` rendering in `src/pipeline.ts` so the title and body each sit on their own line without a prefix, separated by blank lines.
- Update `pipeline.suggestedSquashTitle` in `src/i18n/messages.ts`, `src/i18n/en.ts`, `src/i18n/ko.ts` so it is a plain label string (no `(title) => ...` arg), mirroring `pipeline.suggestedSquashBody`.
- Extend the existing `MERGEABLE: includes squash suggestion hint in confirmMerge message` test in `src/pipeline.test.ts` (currently at line 1122) to assert the new line structure — the value appears on its own line with no prefix.

sehkone:
## Summary

- Split the suggested squash title and body off their label lines so each value sits on its own line with no prefix, and separate every labelled section with a blank line. Triple-clicking the title now selects exactly the value, and the body has a clean selection boundary.
- Simplify `pipeline.suggestedSquashTitle` from `(title) => string` to a plain label string, mirroring `pipeline.suggestedSquashBody`, in `messages.ts`, `en.ts`, and `ko.ts`.
- Extend the existing stage 9 squash suggestion test to assert the new line structure — values on their own lines, blank lines between sections, and no inline prefix.

Part of #262
Closes #264

## Test plan

- [x] `pnpm vitest run` passes (full suite).
- [x] Stage 9 merge-confirm preview shows `Suggested title:` and the title on separate lines, with a blank line before and after.
- [x] Stage 9 merge-confirm preview shows `Suggested body:` and the body on separate lines, with a blank line before the PR URL.
- [x] Triple-clicking the suggested-title line selects only the title (no `Suggested title: ` prefix).
- [x] Korean locale renders the same structure with the `제안 제목:` / `제안 본문:` labels on their own lines.

### DIFF
--- a/src/i18n/en.ts
+++ b/src/i18n/en.ts
@@ -145,7 +145,7 @@ export const en: Messages = {
   "pipeline.mergeConfirmSquashTip":
     "If this repo allows 'Squash and merge', the suggested commit " +
     "message is in the PR body.",
-  "pipeline.suggestedSquashTitle": (title) => `Suggested title: ${title}`,
+  "pipeline.suggestedSquashTitle": "Suggested title:",
   "pipeline.suggestedSquashBody": "Suggested body:",
   "pipeline.prUrl": (url) => `PR: ${url}`,
   "pipeline.worktreeCleanedUp": "Worktree cleaned up.",
--- a/src/i18n/ko.ts
+++ b/src/i18n/ko.ts
@@ -179,7 +179,7 @@ export const ko: Messages = {
     "PR이 병합되었습니까? 확인하면 워크트리를 정리합니다.",
   "pipeline.mergeConfirmSquashTip":
     "이 저장소가 'Squash and merge'를 허용한다면, 제안된 커밋 메시지가 PR 본문에 있습니다.",
-  "pipeline.suggestedSquashTitle": (title) => `제안 제목: ${title}`,
+  "pipeline.suggestedSquashTitle": "제안 제목:",
   "pipeline.suggestedSquashBody": "제안 본문:",
   "pipeline.prUrl": (url) => `PR: ${url}`,
   "pipeline.worktreeCleanedUp": "워크트리가 정리되었습니다.",
--- a/src/i18n/messages.ts
+++ b/src/i18n/messages.ts
@@ -139,7 +139,7 @@ export interface Messages {
   ) => string;
   "pipeline.mergeConfirm": string;
   "pipeline.mergeConfirmSquashTip": string;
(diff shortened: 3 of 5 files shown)


sehkone:
**[Reviewer Round 1]** I don’t see any blockers here. `src/pipeline.ts:1183` and `src/pipeline.ts:1188` move the suggested title/body values onto their own lines, and `src/pipeline.ts:1194` joins the hint sections with blank lines, which directly addresses both UX problems from #264: triple-clicking the title no longer grabs the label, and the body now has a clean boundary before the PR URL and merge prompt. The i18n change in `src/i18n/messages.ts:142`, `src/i18n/en.ts:148`, and `src/i18n/ko.ts:182` is also the right simplification, since the title text is now a plain label just like `pipeline.suggestedSquashBody` instead of a formatter that forces the value inline. The test updates in `src/pipeline.test.ts:1148`, `src/pipeline.test.ts:1150`, and `src/pipeline.test.ts:1154` are meaningful too: they assert the new structure in a way that would have failed before this patch, rather than only checking that the strings still appear somewhere in the prompt.

sehkone:

Issue #267, opened 2026-04-19
Remove legacy parseSquashSuggestionBlock format


sehkone:
Follow-up to #263.

Remove the legacy `**Title:** …` / `**Body:** …` branch from `parseSquashSuggestionBlock` (`src/stage-squash.ts`) and its associated test case in `src/stage-squash.test.ts`.

## Rationale

#263 switched the squash suggestion block emitted by Stage 8 to use fenced code blocks so GitHub renders a one-click copy icon and does not reinterpret Markdown characters. The parser was updated to accept *both* formats so PRs already in the `applied_in_pr_body` state (written by the previous release) continue to render in the Stage 9 inline preview after upgrade.

That compatibility window is no longer needed — there are no in-flight `applied_in_pr_body` runs pointing at PR bodies written in the legacy format that need to be preserved. The legacy branch is dead weight and should be removed now.

## Scope

- Delete `parseLegacySuggestion` and its call site in `parseSquashSuggestionBlock`.
- Delete the legacy-format test case in the `parseSquashSuggestionBlock` describe block.
- Update the docstring on `parseSquashSuggestionBlock` to drop the legacy paragraph.
- Update `CHANGELOG.md` — move the deprecation entry from `### Deprecated` on the previous release to `### Removed` on this one (or consolidate into the `[Unreleased]` section if #263 has not yet been released).

Not needed:
- No `RUN_STATE_VERSION` bump. `squashSubStep` is not format-coupled.
- No migration shim. Legacy PR bodies that somehow persist past the removal will simply fail to render the Stage 9 preview — a cosmetic degradation, not a pipeline failure.

## Timing

Ready to land immediately. The original one-release-cycle delay was a precaution against PRs in `applied_in_pr_body` state written by a prior release; maintainer confirmed no such in-flight runs exist, so the delay no longer applies.

## Parallelism

Independent of #265. Only `CHANGELOG.md` can collide with concurrent work, and the `[Unreleased]` subsections rarely conflict; if they do the resolution is a trivial text merge. Both issues can be worked on in parallel sessions.

PR #268, opened 2026-04-19
Wrap squash suggestion title/body in fenced code blocks (#263)

### LINKED ISSUE #263: Render squash suggestion as fenced code blocks in PR body
Sub-issue of #262 (Part A).

## Problem

The stage 8 prompt template in [src/stage-squash.ts:170-179](https://github.com/aicers/agentcoop/blob/main/src/stage-squash.ts#L170-L179) instructs the agent to emit the suggested commit title and body in the PR body as bold-labeled plain Markdown:

```
<!-- agentcoop:squash-suggestion:start -->
## Suggested squash commit

**Title:** <your title>

**Body:**
<your body — may include `Closes #N` / `Part of #N`>
<!-- agentcoop:squash-suggestion:end -->
```

Two consequences when the user later runs GitHub's "Squash and merge":

- **No one-click copy.** GitHub renders a copy icon automatically on fenced code blocks; bold-labeled plain text does not get one. The user must drag-select past the `**Title:**` label for the title, and drag a multi-line plain-Markdown region for the body.
- **Markdown re-interpretation of the commit body.** Characters that are legal in a commit message (`*`, `` ` ``, `_`, leading `#`, `>`, backslashes, etc.) may be re-formatted or swallowed. What the user sees is not always what they should paste.

## Proposal

Wrap the title and body in separate fenced code blocks inside the marker region:

```
<!-- agentcoop:squash-suggestion:start -->
## Suggested squash commit

**Title**

` ` `text
<your title>
` ` `

**Body**

` ` `text
<your body>
` ` `
<!-- agentcoop:squash-suggestion:end -->
```

(Backticks spaced above only so this issue body renders. Real output uses standard triple backticks.)

### Fence length must be chosen dynamically

Commit bodies can contain code examples (README excerpts, reproducer snippets, fix examples) with their own triple-backtick fences. A hardcoded three-backtick fence would close early at the first `` ``` `` line inside the body, corrupting the block.

Per CommonMark, a fenced block closes only on a line containing a fence of **the same or greater length** made of the same character. The agent prompt and the parser must therefore both select a fence length equal to `max(run of backticks anywhere inside the content, 2) + 1`:

- Title fence length = `max_backtick_run(title) + 1`, minimum 3.
- Body fence length = `max_backtick_run(body) + 1`, minimum 3.
- Use the same character (`` ` ``) for both the opening and closing line of a given block.

This rule keeps "body preserved verbatim" true for any commit message — including ones that reference code blocks.

## Scope

- Update the prompt template in `src/stage-squash.ts` so the agent is instructed to:
  - Emit the title and body inside fenced blocks (with the `text` info string) instead of `**Title:** <title>` / `**Body:**` plain text.
  - Choose the fence length per the dynamic rule above. Include a short worked example in the prompt so the agent handles the backtick-in-body case correctly.
- Rewrite `parseSquashSuggestionBlock` (currently at [src/stage-squash.ts:271](https://github.com/aicers/agentcoop/blob/main/src/stage-squash.ts#L271)) to extract the title and body from the fenced blocks. The parser must:
  - Still key off the existing start/end markers so the surrounding contract is unchanged.
  - Find the opening fence on its own line (allow info-string suffix like `text`), record its length + character, and scan for the first closing line that is a fence of the same character with length `>=` opening. This matches CommonMark fenced block semantics.
  - Tolerate optional surrounding whitespace and the trailing newline inside each fenced block.
  - Also accept the **legacy format** (`**Title:** <title>` / `**Body:**` plain text) for one release cycle — see *Migration / compatibility* below.
  - Return `undefined` when **neither** format is present, or when the block is syntactically malformed. The strict-validation gates in the stage 8 handler rely on that to block rather than silently accept a bad block.
- Update `parseSquashSuggestionBlock` fixtures and tests in `src/stage-squash.test.ts`. Add explicit test cases for:
  - A body that contains a `` ``` `` fenced code block (outer fence length = 4).
  - A body that contains a line of 5 backticks (outer fence length = 6).
  - A title that contains a single backtick (fence length stays at 3).
  - A legacy-format block (parser still returns the same structure for one release).
  - A block that is neither fenced nor legacy-format (parser returns `undefined`).
- `hasValidSuggestionBlock` stays as-is (already defined in terms of `parseSquashSuggestionBlock`).

## Migration / compatibility

**Dual-format parsing is required**, not optional. Stage 8 short-circuits on `squashSubStep === "applied_in_pr_body"` (see [src/stage-squash.ts:442-448](https://github.com/aicers/agentcoop/blob/main/src/stage-squash.ts#L442-L448)) — the agent is **not** re-invoked, so any old-format marker block already written to an existing PR is never rewritten after the upgrade. Stage 9's `getSquashMergeHint` ([src/index.ts:766-780](https://github.com/aicers/agentcoop/blob/main/src/index.ts#L766-L780)) reads that same PR body through the parser. A parser that only accepts the new fenced format would silently drop the stage 9 inline preview for every PR currently in the `applied_in_pr_body` state.

Requirements:

- `parseSquashSuggestionBlock` must accept **either** the new fenced format **or** the legacy `**Title:** <title>` / `**Body:**` plain-text format, returning the same `{ title, body }` shape.
- The new agent prompt only produces the fenced format — legacy parsing is for *reading* existing PR bodies, not writing new ones.
- Mark the legacy branch as deprecated in `CHANGELOG.md` and schedule removal one release after this lands. Open a follow-up issue for the removal in the same PR so it does not get lost.

`RUN_STATE_VERSION` does not need a bump — `squashSubStep` is not format-coupled.

sehkone:
## Summary

- Stage 8's squash suggestion block in the PR body now emits the title and body inside separate fenced code blocks (info string `text`) instead of `**Title:** …` / `**Body:** …` plain Markdown. GitHub renders a one-click copy icon on fenced blocks and does not reinterpret Markdown characters inside them, so the user can paste the suggestion verbatim into the "Squash and merge" dialog.
- The agent is instructed to choose each fence length dynamically per the CommonMark rule (`max(longest backtick run in content, 2) + 1`, minimum 3), with a worked example covering the backtick-in-body case. Title and body fence lengths are computed independently.
- `parseSquashSuggestionBlock` is rewritten to extract the title and body from the fenced blocks using CommonMark-style fence matching (close length `>=` open length, same character). It still accepts the legacy `**Title:** …` / `**Body:** …` plain-text format for one release cycle so that PRs already in the `applied_in_pr_body` state — which stage 8 does not re-invoke the agent for — continue to render in stage 9's inline preview after upgrade.
- The legacy branch is marked deprecated in `CHANGELOG.md`; removal is tracked by #267.

Closes #263

## Suggested squash commit

**Title:** Render squash suggestion as fenced code blocks

**Body:**
Stage 8 previously emitted the suggested squash title and body in
the PR body as `**Title:** <title>` / `**Body:** <body>` plain
Markdown.  GitHub's "Squash and merge" dialog does not render a
one-click copy icon for that shape, and Markdown characters that
are legal in a commit message (`*`, `` ` ``, `_`, leading `#`,
`>`, backslashes) can be reinterpreted or swallowed when the user
drag-selects the suggestion to paste.

Wrap each field in its own CommonMark fenced code block with the
`text` info string so GitHub renders a one-click copy icon and
preserves the content verbatim.  Fence length is chosen per
`max(longest backtick run in content, 2) + 1` (minimum 3),
independently for title and body, so commit bodies that contain
code samples with their own triple-backtick fences do not close
the outer block early.

Rewrite `parseSquashSuggestionBlock` to match CommonMark fence
semantics (close length `>=` open length, same character).  For
one release cycle it also accepts the legacy bold-labeled format
so PRs already in the `applied_in_pr_body` state — which stage 8
does not re-invoke the agent for — continue to render stage 9's
inline preview after upgrade.  Legacy-branch removal is tracked
by #267.

Closes #263

## Test plan

- [x] `pnpm vitest run` — full suite green (1827 tests).
- [x] `parseSquashSuggestionBlock` parses a well-formed fenced block.
- [x] Parser handles a body containing a nested triple-backtick block (outer fence length 4).
- [x] Parser handles a body containing a run of five backticks (outer fence length 6).
- [x] Parser handles a title that contains a single backtick (fence length stays 3).
- [x] Parser tolerates extra blank lines around fences.
- [x] Parser still accepts the legacy `**Title:** …` / `**Body:** …` format.
- [x] Parser returns `undefined` for a block that is neither fenced nor legacy.
- [x] Parser returns `undefined` for unterminated fenced blocks.
- [x] `pnpm biome check` clean for changed files.
- [x] `pnpm build` succeeds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,21 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Changed
 
+- Stage 8's squash suggestion block in the PR body now emits the title and
+  body inside separate fenced code blocks (info string `text`) instead of
+  bold-labeled plain Markdown (`**Title:** …` / `**Body:** …`).  GitHub
+  renders a one-click copy icon on fenced blocks and does not reinterpret
+  Markdown characters inside, so the user can paste the suggestion verbatim
+  into the "Squash and merge" dialog.  The agent chooses each fence length
+  dynamically per the CommonMark rule
+  (`max(longest backtick run in content, 2) + 1`, minimum 3) so commit
+  bodies containing their own triple-backtick samples survive unchanged.
+- `parseSquashSuggestionBlock` now accepts either the new fenced format or
+  the legacy `**Title:** …` / `**Body:** …` plain-text format, returning
+  the same `{ title, body }` shape.  This is required because Stage 8
+  short-circuits on `squashSubStep === "applied_in_pr_body"` without
+  re-invoking the agent, so existing PRs in that state still render
+  correctly in the Stage 9 inline preview after upgrade.
 - Stage 8 verdict keywords are now `SQUASHED_MULTI` / `SUGGESTED_SINGLE` /
   `BLOCKED` (previously `COMPLETED` / `BLOCKED`).  When the verdict is
   ambiguous after a clarification retry, the handler runs a deterministic
@@ -51,6 +66,15 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Stage 9 merge-confirm screen now includes a conditional one-line tip
   (`pipeline.mergeConfirmSquashTip`) when a squash suggestion is live in
   the PR body.
+
+### Deprecated
+
+- The legacy `**Title:** …` / `**Body:** …` plain-text form of the squash
+  suggestion block (written by older versions of Stage 8) is deprecated.
+  `parseSquashSuggestionBlock` continues to accept it for one release cycle
+  so PRs already in the `applied_in_pr_body` state still render in the
+  Stage 9 inline preview after upgrade.  The legacy branch will be removed
+  in the following release — tracked by #267.
 
 ### Fixed
 
--- a/src/stage-squash.test.ts
+++ b/src/stage-squash.test.ts
@@ -954,7 +954,7 @@ describe("createSquashStageHandler", () => {
     }
 
     test("count decreased AND marker present → SQUASHED_MULTI (count wins)", async () => {
-      const prBody = `${SQUASH_SUGGESTION_START_MARKER}\n**Title:** stale\n${SQUASH_SUGGESTION_END_MARKER}`;
+      const prBody = `${SQUASH_SUGGESTION_START_MARKER}\n**Title:** stale\n\n**Body:**\nstale body\n${SQUASH_SUGGESTION_END_MARKER}`;
       const opts = makeOpts({
         agent: makeAmbiguousAgent(),
         countBranchCommits: vi
@@ -970,7 +970,7 @@ describe("createSquashStageHandler", () => {
     });
 
     test("count unchanged AND marker present → SUGGESTED_SINGLE", async () => {
-      const prBody = `${SQUASH_SUGGESTION_START_MARKER}\n**Title:** suggested\n${SQUASH_SUGGESTION_END_MARKER}`;
+      const prBody = `${SQUASH_SUGGESTION_START_MARKER}\n**Title:** suggested\n\n**Body:**\nsuggested body\n${SQUASH_SUGGESTION_END_MARKER}`;
       const chooseSquashApplyMode = vi.fn().mockResolvedValue("github");
       const opts = makeOpts({
         agent: makeAmbiguousAgent(),
@@ -1018,7 +1018,7 @@ describe("createSquashStageHandler", () => {
     });
 
     test("awaiting_user_choice with marker present → re-presents user choice", async () => {
-      const prBody = `${SQUASH_SUGGESTION_START_MARKER}\n**Title:** T\n${SQUASH_SUGGESTION_END_MARKER}`;
+      const prBody = `${SQUASH_SUGGESTION_START_MARKER}\n**Title:** T\n\n**Body:**\nB\n${SQUASH_SUGGESTION_END_MARKER}`;
       const chooseSquashApplyMode = vi.fn().mockResolvedValue("github");
       const opts = makeOpts({
         savedSquashSubStep: "awaiting_user_choice",
@@ -1432,7 +1432,7 @@ describe("createSquashStageHandler", () => {
     });
 
     test("emits SUGGESTED_SINGLE when marker block is present", async () => {
-      const prBody = `${SQUASH_SUGGESTION_START_MARKER}\n**Title:** s\n${SQUASH_SUGGESTION_END_MARKER}`;
+      const prBody = `${SQUASH_SUGGESTION_START_MARKER}\n**Title:** s\n\n**Body:**\ns body\n${SQUASH_SUGGESTION_END_MARKER}`;
       const events = new PipelineEventEmitter();
       const handler = vi.fn();
       events.on("pipeline:verdict", handler);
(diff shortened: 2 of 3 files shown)


sehkone:
**[Reviewer Round 1]**

The overall direction here looks right, but I do see one blocking correctness issue in the new dual-format parser.

At `src/stage-squash.ts:351`, `parseSquashSuggestionBlock()` falls back to `parseLegacySuggestion(inner)` whenever `parseFencedSuggestion(inner)` returns `undefined`. The problem is that `parseLegacySuggestion()` at `src/stage-squash.ts:443` matches `**Title:**` / `**Body:**` anywhere inside the marker block, not just an actual top-level legacy block. That means a syntactically malformed fenced block can still be accepted if the fenced content happens to contain those strings.

For example, this now parses successfully even though the fenced title block is unterminated:

````text
<!-- agentcoop:squash-suggestion:start -->
## Suggested squash commit

**Title**

```text
Broken suggestion

**Title:** this is just prose inside the broken fence

**Body:**
still just prose inside the same broken fence
<!-- agentcoop:squash-suggestion:end -->
````

That contradicts the issue requirement that malformed blocks return `undefined`, because stage 8 relies on that strict failure to block instead of silently accepting a bad suggestion.

I think the fix needs to make the legacy branch opt-in only when the block actually looks like legacy format at top level, or otherwise avoid trying legacy parsing after the fenced labels have already been seen. I’d also add a regression test near the new malformed-block coverage in `src/stage-squash.test.ts:1730` proving that an unterminated fenced block containing legacy-looking `**Title:**` / `**Body:**` text still returns `undefined`.

**Comment by sehkone** [comment]

**Comment by sehkone** [comment]
**[Author Round 1]**

- **Fixed — strict fenced-intent detection.** Good catch. `parseSquashSuggestionBlock` no longer falls through from `parseFencedSuggestion` to `parseLegacySuggestion` on failure. Format is now chosen upfront by label shape: if the inner content contains `**Title**` on its own line (the new-format label, no colon), the block is treated as fenced-intent and a parse failure returns `undefined` instead of silently accepting legacy-looking prose that happened to live inside an unterminated fence. The legacy branch (`**Title:** <title>` / `**Body:**`) is only reached when no `**Title**` label is present at all. See `src/stage-squash.ts:346-360`.

- **Fixed — regression test.** Added `"returns undefined when fenced-intent block has legacy-looking prose inside an unterminated fence"` in `src/stage-squash.test.ts` next to the existing malformed-block coverage. It uses the exact shape from the review — an unterminated title fence whose contents include `**Title:** …` and `**Body:** …` prose — and asserts `undefined`.

Verified: `pnpm vitest run` (1828/1828 green), `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm build` all clean.

**Comment by sehkone** [comment]
**[Reviewer Round 2]**

The Round 1 fenced-intent fix in `src/stage-squash.ts:351-362` looks good; I’m not carrying that item forward.

I do still see one remaining correctness issue in the legacy branch at `src/stage-squash.ts:454-468`. `parseLegacySuggestion()` still accepts malformed legacy blocks as valid because it only requires a `**Title:**` match and treats a missing `**Body:**` marker as `body = ""`. For example, this now parses successfully even though it is not a syntactically valid legacy block:

```text
<!-- agentcoop:squash-suggestion:start -->
## Suggested squash commit

**Title:** Fix widget rendering

This forgot the body label entirely.
<!-- agentcoop:squash-suggestion:end -->
```

That contradicts the issue requirement that malformed blocks return `undefined`, and it still weakens the Stage 8 strict gate because `hasValidSuggestionBlock()` at `src/stage-squash.ts:484-485` will accept the malformed block. More generally, the legacy parser is still using a whole-block `match()` / `indexOf()` approach, so stray `**Title:**` / `**Body:**` text in prose can be mistaken for a real top-level legacy block.

I think the legacy path needs the same sort of top-level shape check you added for fenced intent: require actual legacy labels on their own lines in order, then parse relative to those labels. I’d also add regression coverage in `src/stage-squash.test.ts` for a marker block with `**Title:** ...` but no `**Body:**` label (and ideally one where those strings appear only inside prose) to prove it now returns `undefined`.

sehkone:

sehkone:
**[Author Round 2]**

- **Fixed — strict top-level shape for legacy parse.** `parseLegacySuggestion` no longer uses a whole-block `match()` / `indexOf()` scan. It now splits into lines and requires both labels on their own top-level lines, in order: a line matching `^\s*\*\*Title:\*\*\s+(.+)\s*$` (content required on the same line) followed by a later line matching `^\s*\*\*Body:\*\*\s*$`. A missing `**Body:**` label now returns `undefined` rather than silently yielding `body = ""`, and stray `**Title:**` / `**Body:**` strings inside prose can no longer be mistaken for real labels because both regexes are line-anchored. See `src/stage-squash.ts:454-494`.

- **Fixed — regression tests.** Added two cases next to the existing legacy coverage in `src/stage-squash.test.ts`:
  - `"legacy format returns undefined when the `**Body:**` label is missing"` — the exact shape from the review (title label present, no body label, just trailing prose).
  - `"legacy format returns undefined when labels appear only inside prose"` — both `**Title:**` and `**Body:**` appear only mid-line inside a paragraph, must still return `undefined`.

- **Adjusted — existing legacy fixtures in unrelated tests.** Four pre-existing test fixtures in `src/stage-squash.test.ts` used truncated legacy blocks (`**Title:** T` with no `**Body:**` label) purely as marker-presence signals. Under the stricter parser those blocks are no longer valid, so I added a `**Body:**` section to each so they continue to exercise their intended code path (marker-present → SUGGESTED_SINGLE, SQUASHED_MULTI when count wins, resume on `awaiting_user_choice`, pipeline verdict emission).

Verified: `pnpm vitest run` (1830/1830 green), `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm build` all clean.

sehkone:
**[Reviewer Round 3]**

The Round 2 legacy-shape fix at `src/stage-squash.ts:462` is present, so I’m not carrying that item forward.

I do still see one remaining compatibility bug in the format-selection logic at `src/stage-squash.ts:359`. `parseSquashSuggestionBlock()` currently treats any line that is exactly `**Title**` anywhere inside the marker block as proof that the block is fenced-format, and then skips the legacy parser entirely. That breaks otherwise valid legacy blocks whose body happens to contain a standalone `**Title**` line, even though the issue explicitly requires one-release dual-format parsing for existing PR bodies.

For example, this is a valid legacy block under the current documented contract, but it now parses as `undefined`:

```text
<!-- agentcoop:squash-suggestion:start -->
## Suggested squash commit

**Title:** Fix widget rendering

**Body:**
Intro paragraph.

**Title**

Closes #42
<!-- agentcoop:squash-suggestion:end -->
```

The detector at `src/stage-squash.ts:359` routes that block to `parseFencedSuggestion()`, which then fails because the standalone `**Title**` line in the legacy body is not followed by a fenced code block. So the new strictness fix regresses the required legacy compatibility on a body shape that commit messages can legitimately contain.

I think the format choice needs to key off an actual top-level fenced shape rather than the mere presence of `**Title**` somewhere in the block, and I’d add a regression test near the legacy coverage in `src/stage-squash.test.ts:1774` proving that a valid legacy block still parses when its body contains a standalone `**Title**` line.

sehkone:

sehkone:
**[Author Round 3]**

- **Fixed — format selector requires actual fenced shape.** Good catch. `parseSquashSuggestionBlock` no longer keys off the bare presence of a `**Title**` line. The new `looksLikeFencedIntent` helper requires an actual `**Title**` label line *followed (after optional blank lines) by an opening code fence* (`` ``` `` or longer) before routing to `parseFencedSuggestion`. A legacy block whose body contains a standalone `**Title**` line as prose no longer gets misrouted — it falls through to the legacy parser and parses correctly. The Round 1 strictness guarantee is preserved: once a block is classified as fenced-intent (label + fence shape both present), a malformed fence still fails to `undefined` rather than silently falling through to legacy. See `src/stage-squash.ts:351-381`.

- **Fixed — regression test.** Added `"legacy format parses when body contains a standalone `**Title**` line"` next to the existing legacy coverage in `src/stage-squash.test.ts`. It uses the exact shape from the review — a valid legacy block whose body includes a standalone `**Title**` line — and asserts it parses with the full body (including the `**Title**` line) preserved.

Verified: `pnpm vitest run` (1831/1831 green), `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm build` all clean.

sehkone:
**[Reviewer Round 4]**

The Round 3 fix for the plain `**Title**` prose case is present, so I’m not carrying that exact item forward.

I do still see a narrower form of the same legacy-compatibility bug in the format selector at `src/stage-squash.ts:374`. `looksLikeFencedIntent()` now requires `**Title**` followed by an opening fence, but it still scans the entire marker block for that pattern. That means a valid legacy block is still misclassified if its legacy body legitimately contains a standalone `**Title**` line followed by a fenced code sample.

For example, this is still a valid legacy block under the one-release compatibility requirement, but the current code routes it to `parseFencedSuggestion()` and returns `undefined` instead of parsing it as legacy:

```text
<!-- agentcoop:squash-suggestion:start -->
## Suggested squash commit

**Title:** Fix widget rendering

**Body:**
Intro paragraph.

**Title**

```text
example
```

Closes #42

```

That happens because the body's later `**Title**` + fence sequence satisfies `looksLikeFencedIntent()` at `src/stage-squash.ts:374`, even though the actual top-level format is legacy. The issue explicitly requires dual-format parsing for existing PR bodies, so the detector still needs to key off the actual top-level start of the suggestion block rather than any later prose inside the legacy body.

I’d fix that by making the format decision based on the first top-level recognizable label sequence near the start of the block (or otherwise preferring an earlier legacy `**Title:**` over a later fenced-looking pattern), and I’d add a regression test near `src/stage-squash.test.ts:1782` covering a legacy body that contains `**Title**` immediately followed by a fenced block.

sehkone:

sehkone:
**[Author Round 4]**

- **Fixed — format chosen by first top-level title label.** Good catch. `parseSquashSuggestionBlock` no longer asks "does a fenced shape exist anywhere in the block?". The new `detectFormat` helper walks the inner content line by line and returns whichever top-level title label it encounters *first*: a `**Title:** <content>` line routes to the legacy parser, while a `**Title**` label line followed (after optional blank lines) by an opening code fence routes to the fenced parser. A legacy block whose top-level `**Title:**` sits above a later `**Title**` + fence sequence inside its body now wins the legacy route, as required by the one-release dual-format contract. The Round 1 strictness guarantee still holds: once classified as fenced, a malformed fenced block fails to `undefined` rather than silently falling through to legacy. See `src/stage-squash.ts:351-394`.

- **Fixed — regression test.** Added `"legacy format parses when body contains `**Title**` immediately followed by a fenced block"` next to the existing legacy coverage in `src/stage-squash.test.ts`. It uses the exact shape from the review — a valid legacy block whose body contains a standalone `**Title**` line directly followed by a `` ```text `` fenced code sample — and asserts the full body (including the `**Title**` line and the fenced sample) is preserved.

Verified: `pnpm vitest run` (1832/1832 green), `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm build` all clean.

sehkone:
**[Reviewer Round 5]**

I verified the Round 4 fix in the diff.

- `parseSquashSuggestionBlock()` now classifies by the first top-level title label instead of treating any later `**Title**` + fence sequence as fenced intent, which resolves the legacy-body misclassification at `src/stage-squash.ts:351`.
- The regression coverage for a legacy body containing `**Title**` immediately followed by a fenced block is present at `src/stage-squash.test.ts:1808`.

I do not have any remaining blocking issues from this round.

One non-blocking nit: a few internal comments still describe a valid suggestion block as requiring `**Title:**`, which is now stale with the fenced format at `src/stage-squash.ts:523`, `src/stage-squash.ts:843`, and `src/stage-squash.ts:883`.

sehkone:

sehkone:
**[Reviewer Unresolved Round 5]**

- A few internal comments still describe a valid suggestion block as requiring `**Title:**`, which is stale now that the current format is fenced. Please update the wording at `src/stage-squash.ts:523`, `src/stage-squash.ts:843`, and `src/stage-squash.ts:883` so the comments match the new dual-format parsing behavior.

PR #270, opened 2026-04-19
Add [t]/[b] hotkey copy for squash suggestion (#265)

### LINKED ISSUE #265: Add OSC 52 hotkey copy for squash suggestion in TUI
Sub-issue of #262 (Part C).

## Problem

Even after the label/value split from the sibling sub-issue, the TUI user still has to drag-select and invoke the terminal's own copy shortcut to paste the suggested title and body into GitHub's "Squash and merge" dialog. The GitHub side (sibling sub-issue for the PR body) gets one-click copy for free via fenced code blocks; the TUI side has no equivalent.

## Proposal

On the merge-confirm screen, add two hotkeys that copy the corresponding value to the system clipboard:

- `t` → copies the suggested **title** (single line, no trailing newline).
- `b` → copies the suggested **body** (multi-line, exact contents — not title + body combined).

The two values are copied **separately** because GitHub's "Squash and merge" dialog has two input fields (title + body); a combined copy would force the user to split again on paste.

## Clipboard mechanism — candidate list, not single mechanism

OSC 52 support cannot be detected synchronously. The escape sequence is a fire-and-forget write: the terminal either honors it or silently drops it, with no response. `stdout.isTTY` only proves "stdout is a terminal" — it does not prove the terminal honors OSC 52. Conversely, a native clipboard tool (`pbcopy` / `wl-copy` / `xclip`) on PATH returns a deterministic exit code so its success is verifiable.

A "pick one mechanism at render time" policy therefore has a real failure mode: a terminal that is a TTY but rejects OSC 52 (tmux without passthrough, old VTE builds, locked-down corporate terminals), on a machine that also has a native tool available — the render commits to OSC 52, the write silently does nothing, and the user pastes an unchanged clipboard without knowing why.

The utility must therefore return an **ordered candidate list**, and `writeToClipboard` must try each candidate in order until one succeeds (for candidates whose success is verifiable) or falls through.

### Ordering policy

Detection picks the order based on environment:

- **SSH session** (`process.env.SSH_TTY` or `SSH_CONNECTION` set) → `[osc52, ...native-if-any]`. A native tool on the remote machine writes to the *remote* clipboard, which is not what the user wants; OSC 52 rides the terminal escape stream back to the user's local terminal. Native tools stay in the list only as a last resort for remote-side convenience (very rare).
- **Local session** → `[<platform-native tool>, osc52]`:
  - macOS: `pbcopy` (always available).
  - Linux + `$WAYLAND_DISPLAY` set and `wl-copy` on PATH: `wl-copy`.
  - Linux + `$DISPLAY` set and `xclip` on PATH: `xclip -selection clipboard`.
  - Fall through to OSC 52 if none of the above match.
- **Neither OSC 52 nor any native tool reachable** → empty list. UI treats this as "unavailable" and does not render hints at all (see below).

### Write semantics

- **Native tool candidate** (`pbcopy` / `wl-copy` / `xclip`): success = `spawnSync` returns exit code 0 within a short timeout (e.g. 1s). Any non-zero exit or timeout is a failure — the writer moves on to the next candidate.
- **OSC 52 candidate**: `process.stdout.write` of the encoded sequence. The write itself is synchronous and cannot verify terminal behavior, so success = write returned without error. When OSC 52 is the **last** candidate in the list, its result is reported to the UI as `"ok"` (best-effort). When OSC 52 is **not last** — i.e. a native candidate follows it on the rare SSH-plus-native-tool path — a later candidate may still be tried only if the OSC 52 write throws; a silent-ignore terminal is indistinguishable from success in that slot.
- Overall result reported to UI: `"ok"` as soon as any candidate reports `"ok"`; `"error"` only if the list is exhausted without an `"ok"`.

### Stdout caveat

OSC 52 writes must go to the controlling terminal's stdout. When stdout is redirected (non-TTY), OSC 52 ends up in the file — the fallback path handles this case. The app currently only enforces `stdin.isTTY` ([src/index.ts:312](https://github.com/aicers/agentcoop/blob/main/src/index.ts#L312)), so detection must not assume a stdout TTY. OSC 52 enters the candidate list only when `stdout.isTTY` is true.

## Availability and feedback — no silent states

Clipboard support is classified once per merge-confirm render and drives UI rendering:

| State | Condition | UI behavior |
| --- | --- | --- |
| `available` | Candidate list is non-empty | Render `[t] copy` / `[b] copy` hints next to the title / body. On keypress, attempt copy through the candidate list and show "Copied" on `"ok"`, "Copy failed" on `"error"`. |
| `unavailable` | Candidate list is empty | **Do not render the hints at all.** The user is not told about a feature that cannot work here. Falls back to the existing drag-select path from sibling sub-issue Part B. |

Explicit non-goals for this state machine:

- **No silent no-op.** Pressing `t` or `b` must always produce visible feedback when hints are rendered. Never the combination of "hint is shown" + "keypress does nothing visible" — that reads as broken.
- **No late availability surprises.** Determining availability at render time avoids "hint promises copy, but copy silently fails" flows. (OSC 52's unverifiability is the one residual gap; the candidate-list policy above shrinks it to "OSC 52 is the only option and the terminal silently ignores it," which is rare and indistinguishable from success at the protocol level.)

## Hint rendering contract — placeholder substitution

The `InputArea` currently renders `message: string` line-by-line via `message.split("\n").map(line => <Text>{line}</Text>)` ([src/ui/InputArea.tsx:50-53](https://github.com/aicers/agentcoop/blob/main/src/ui/InputArea.tsx#L50-L53)). A plain-text message gives `InputArea` no way to know *where* a hotkey hint sits or *which* substring to update when the per-hotkey status changes. The rendering contract must therefore be pinned down.

**Chosen contract: sentinel substitution inside `message`.** The caller (`askMerge`) embeds a unique sentinel token per hotkey at the position where the hint should appear; `InputArea` walks each line, finds the sentinel, and replaces it with a `<Text>` child whose content reflects the current per-hotkey status. The `message` remains a `string` and the API surface of `confirmMerge(message: string, hotkeys?: ...)` stays minimal.

Why sentinel substitution and not a structured block/line model:

- Preserves the existing `InputRequest { message: string, choices? }` shape — no ripple across all other `confirmMerge`-style surfaces.
- The only screen that needs inline hotkey hints today is merge-confirm; a full structured model would be over-engineering for one use case. The structured variant can be added later alongside the string form if a second use case appears.
- i18n already has function-valued messages, so callers are comfortable composing strings with typed inputs — sentinel embedding fits that style.

### Sentinel syntax

- Token form: `{{hk:<id>}}` where `<id>` matches a `hotkeys[].id` field.
- `<id>` is constrained to `[a-zA-Z0-9_-]+` so matching is trivial and cannot collide with regex metacharacters.
- The token is treated as literal text by `message.split("\n")` (it contains no newlines). On each line, `InputArea` splits by the `{{hk:...}}` regex to get text / sentinel / text fragments and renders them in order.
- A sentinel with an unknown id (not in `hotkeys`) is rendered as literal text — silent drop would hide caller bugs.

### Per-hotkey status → rendered label

The label rendered in place of `{{hk:<id>}}` is derived from the status InputArea holds for that hotkey:

| Status | Label | Color / weight |
| --- | --- | --- |
| `idle` | `[<key>] copy` | dim (cyan bracketed key, default text) |
| `copied` | `[<key>] copied` | dim, auto-clears to `idle` after ~1s |
| `failed` | `[<key>] copy failed` | warning color, persists until re-render |

`<key>` is the character the user presses (e.g. `t`, `b`). The `copy` / `copied` / `copy failed` words come from i18n.

### Example

The caller composes:

```
Suggested title:   {{hk:title}}
<title>

Suggested body:   {{hk:body}}
<body>
```

with `hotkeys: [{ id: "title", key: "t", onPress: ... }, { id: "body", key: "b", onPress: ... }]`.

`InputArea` renders (idle state):

```
Suggested title:   [t] copy
<title>

Suggested body:   [b] copy
<body>
```

After the user presses `t` and the writer returns `"ok"`:

```
Suggested title:   [t] copied
<title>

Suggested body:   [b] copy
<body>
```

(and auto-reverts after ~1s).

If the `hotkeys` array is empty (clipboard unavailable per the detection policy above), the caller composes the message **without** sentinels so no `[t] copy` text appears anywhere.

## Scope

- **Input model**: extend `InputRequest` at [src/ui/InputArea.tsx:8](https://github.com/aicers/agentcoop/blob/main/src/ui/InputArea.tsx#L8) with an optional `hotkeys?: { id: string; key: string; onPress: () => Promise<"ok" | "error"> }[]` field. `id` is the sentinel key; `key` is the keyboard character. Extend `useInput` at [src/ui/InputArea.tsx:28](https://github.com/aicers/agentcoop/blob/main/src/ui/InputArea.tsx#L28) to dispatch on `key` without closing the prompt. Track per-hotkey status in local state.
- **Line rendering**: in the choices branch (and the text-input branch if needed), replace the current `message.split("\n").map(line => <Text>{line}</Text>)` with a helper that, for each line, splits on the `{{hk:<id>}}` regex and renders fragments — text fragments as plain `<Text>` children, sentinel fragments as the state-derived label with the appropriate color/weight.
- **Prompt surface**: `confirmMerge(message: string)` at [src/pipeline.ts:213](https://github.com/aicers/agentcoop/blob/main/src/pipeline.ts#L213) takes only a string — add an optional `hotkeys` parameter so `askMerge` can pass title/body copy handlers in. The parameter is forwarded verbatim to `InputRequest`.
- **Clipboard utility**: isolated module with two exported pieces:
  - `detectClipboardSupport(): ClipboardCandidate[]` — runs at render time, returns the ordered candidate list per the policy above. Pure function of `process.env` / `process.platform` / `stdout.isTTY` / PATH presence — no side effects, easy to test with mocks.
  - `writeToClipboard(value: string, candidates: ClipboardCandidate[]): Promise<"ok" | "error">` — iterates the candidate list, returns `"ok"` on first success, `"error"` only if all candidates fail. Does not throw.
- **Caller composition**: `askMerge` calls `detectClipboardSupport` once per render. If empty, the message is composed without sentinels and `hotkeys` is not passed. If non-empty, the message embeds `{{hk:title}}` / `{{hk:body}}` and `hotkeys` is populated with handlers that call `writeToClipboard`.
- **i18n**: en + ko strings for the label variants: `copy`, `copied`, `copy failed`. The bracketed key and bracket characters come from the renderer, not i18n, so translators only translate the verb.
- **Key selection**: `t` and `b` are currently unused. Merge-confirm choices are numeric-only ([src/ui/InputArea.tsx:28-37](https://github.com/aicers/agentcoop/blob/main/src/ui/InputArea.tsx#L28-L37)), so there is no conflict today. The input model extension must still guard against `hotkeys[].key` colliding with any `choices` digit.

## Tests

- Unit test the OSC 52 emitter — capture bytes written to stdout, assert header + base64 payload.
- Unit test `detectClipboardSupport` with mocked env / platform / PATH / stdout. Cover every branch of the ordering policy: local macOS, local Wayland, local X11, SSH, SSH-plus-native-tool, none-available.
- Unit test `writeToClipboard` with a mocked candidate list. Cover: first-candidate success, first fails + second succeeds, all fail.
- UI-layer test in the Ink input testing pattern used in [src/ui/components.test.tsx](https://github.com/aicers/agentcoop/blob/main/src/ui/components.test.tsx):
  - Message containing `{{hk:title}}` with an `idle` status renders `[t] copy` at that position; the sentinel is not visible in the output.
  - After pressing `t` and the writer returning `"ok"`, the same position renders `[t] copied`; after the ~1s auto-clear, it reverts to `[t] copy`.
  - After the writer returns `"error"`, the position renders `[t] copy failed` and does not auto-revert.
  - Empty `hotkeys` → no sentinels in the message → no `[t] copy` text rendered; pressing `t` is a no-op; prompt does not close.
  - Sentinel with an unknown id renders as literal text (parser does not silently drop).

## Dependencies

- Sibling sub-issue for the TUI label split (#264) should land first, because it creates the visual layout that the `[t] copy` / `[b] copy` hints sit next to. Not a hard dependency, but landing Part B first keeps this PR focused on the input model + clipboard work.

## Out of scope

- Mouse-click handling. Enabling terminal mouse mode breaks the terminal's own select-to-copy behavior across most setups and introduces cross-terminal flakiness. Hotkey-based copy is the right primitive for a TUI.
- Exposing the clipboard utility to other surfaces in this PR. Building it as a reusable module is in scope, but wiring other surfaces (PR URL, issue title, etc.) to it is a follow-up.
- A fully structured `InputRequest.blocks` model. The sentinel contract covers the current need; upgrading to a structured variant is a separate change if a second surface needs inline hotkey hints.

sehkone:
## Summary

- On the Stage 9 merge-confirm screen, render `[t] copy` / `[b] copy` hotkey hints next to the suggested squash title and body. Pressing `t` or `b` writes the corresponding value to the system clipboard independently, so it lines up with GitHub's separate title / body fields in the "Squash and merge" dialog.
- New `src/ui/clipboard.ts` utility: `detectClipboardSupport()` returns an ordered candidate list (`pbcopy` / `wl-copy` / `xclip` on local sessions, OSC 52 first on SSH, OSC 52 fallback when stdout is a TTY). `writeToClipboard()` tries candidates in order, returns `"ok"` on first success, `"error"` only when every candidate fails. The macOS `pbcopy` candidate is enqueued unconditionally and invoked via `/usr/bin/pbcopy` so a stripped `PATH` never hides the deterministic native path.
- `InputArea` gains an optional `hotkeys` field on `InputRequest` and a `{{hk:<id>}}` sentinel-substitution contract inside `message`. Sentinels render as `[<key>] copy` / `[<key>] copied` / `[<key>] copy failed` based on per-hotkey status; `copied` auto-reverts after ~1s, `copy failed` persists until re-render. Unknown sentinel ids render as literal text so caller bugs surface. A per-hotkey sequence counter protects against out-of-order completions when the user presses the same hotkey twice while the first attempt is still pending. Hotkeys whose `key` collides with an active numeric choice digit are dropped so prompt semantics are never silently shadowed.
- `askMerge` generates per-render unique sentinel ids (`title-<random>` / `body-<random>`) so that agent-authored title/body text containing a literal `{{hk:title}}` or `{{hk:body}}` substring cannot spoof the real hotkey hints; user-authored sentinel-looking text falls through to the unknown-id path and renders as literal text.
- When the candidate list is empty, `askMerge` composes the message without sentinels and does not pass `hotkeys` — the user falls back to drag-select rather than being shown a hint that silently does nothing.
- i18n: `input.copy` / `input.copied` / `input.copyFailed` in en + ko.

Closes #265

## Test plan

- [x] `pnpm biome check` is clean for the changed files.
- [x] `pnpm tsc --noEmit` is clean.
- [x] `pnpm vitest run` — all suites pass, including new coverage in `src/ui/clipboard.test.ts`, `src/ui/components.test.tsx`, and `src/pipeline.test.ts`.
- [x] Unit: `encodeOsc52` emits `ESC]52;c;<base64>BEL` and round-trips multibyte UTF-8.
- [x] Unit: `detectClipboardSupport` branches cover local macOS, local Wayland, local X11, SSH, SSH-plus-native-tool, non-TTY, and linux without DISPLAY/WAYLAND.
- [x] Unit: macOS `pbcopy` stays enqueued on a stripped `PATH` and without stdout TTY; the writer invokes `/usr/bin/pbcopy`, never bare `pbcopy`.
- [x] Unit: `writeToClipboard` covers first-candidate success, first fails + second succeeds, OSC 52 write that throws falls through, native spawn error skips, all-fail, and empty-list.
- [x] UI: sentinel substitution renders `[t] copy` inline; the raw sentinel is not visible in the frame.
- [x] UI: pressing `t` swaps the label to `[t] copied` on `"ok"` and reverts after ~1s; `"error"` swaps to `[t] copy failed` and does not revert.
- [x] UI: out-of-order completions cannot overwrite a newer press's result.
- [x] UI: hotkey colliding with an active numeric choice digit is filtered out; the numeric dispatch still wins.
- [x] UI: empty `hotkeys` means no `[t] copy` text and pressing `t` is a no-op that does not close the prompt.
- [x] UI: sentinel with an unknown id renders as literal text.
- [x] UI: numeric choice keys (`1`, `2`) still work alongside hotkeys.
- [x] Pipeline: `MERGEABLE` with clipboard available embeds per-render unique `{{hk:title-<hex>}}` / `{{hk:body-<hex>}}` sentinels and forwards hotkey handlers that call `writeToClipboard` with the raw title / body.
- [x] Pipeline: sentinel-looking substrings (literal `{{hk:title}}` / `{{hk:body}}`) inside the suggested title or body are rendered literally and do not spoof the real hotkey hints.
- [x] Pipeline: `MERGEABLE` with clipboard unavailable omits sentinels and forwards no `hotkeys` argument.
- [x] Manual: local macOS — `[t]` on a live Stage 9 pastes the suggested title into GitHub's "Squash and merge" title field; `[b]` pastes the body. (Verified end-to-end against the built `dist/ui/clipboard.js`: `detectClipboardSupport` → `["pbcopy"]`, `writeToClipboard` → `"ok"`, `pbpaste` returned the written value.)
- [ ] Manual: SSH into a Linux box — `[t]` / `[b]` write via OSC 52 to the user's local clipboard (verify in the host OS). *(Requires a real remote host; not reachable from CI. OSC 52 byte emission is covered by the `encodeOsc52` unit test and the SSH ordering policy by `detectClipboardSupport` tests.)*
- [ ] Manual: terminal with no OSC 52 and no native tool on PATH — no `[t] copy` hints are rendered. *(Requires a specially-locked terminal session; covered by the `no TTY and no native tool: empty list` detection test plus the `empty hotkeys` UI test that verifies hints are absent.)*

## Suggested squash commit

**Title**

```text
Add [t]/[b] hotkey copy for squash suggestion
```

**Body**

```text
The Stage 9 merge-confirm screen already prints a suggested squash
title and body, but the user still had to drag-select and trigger
the terminal's own copy shortcut to paste each field into GitHub's
"Squash and merge" dialog. Add per-field hotkeys that copy the two
values independently so the flow matches GitHub's two input fields.

Clipboard writes go through an ordered candidate list rather than a
single mechanism. OSC 52 cannot be detected synchronously — a
terminal may silently drop the sequence — so the writer prefers
native tools on local sessions, OSC 52 first on SSH, and reports
"ok" only when a candidate verifiably succeeds. When no candidate
is reachable the UI renders no hint, so the user is never promised
a copy action that would silently no-op. pbcopy is invoked via the
absolute /usr/bin/pbcopy path so a stripped PATH cannot hide the
deterministic native writer on macOS.

InputArea picks up a {{hk:<id>}} sentinel-substitution contract
inside message so per-hotkey status (idle / copied / failed) can
be reflected inline without reshaping the InputRequest API. The
sentinel ids are randomized per render so agent-authored title or
body text cannot spoof the real hotkey hints, and hotkeys whose
key collides with an active numeric choice digit are dropped so
the prompt's numeric dispatch is never shadowed.

Closes #265
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,20 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   body, and PR URL inline when the squash stage finished via the
   PR-body suggestion path, so the user can copy-paste without opening the
   browser.
+- Stage 9 (Done) merge-confirm screen now renders `[t] copy` / `[b] copy`
+  hotkey hints next to the suggested squash title and body when the terminal
+  can write to the system clipboard.  Pressing `t` copies the title and `b`
+  copies the body (each independently, so the values line up with GitHub's
+  separate "Squash and merge" title / body fields).  A small clipboard
+  utility detects the environment and returns an ordered candidate list
+  (`pbcopy` / `wl-copy` / `xclip` on local sessions, OSC 52 first on SSH
+  sessions, OSC 52 as fallback everywhere stdout is a TTY), and the writer
+  tries candidates in order until one succeeds.  When no candidate is
+  reachable, the hints are not rendered at all — the user falls back to
+  drag-select as before, rather than seeing a hint that silently does
+  nothing.  Per-hotkey status (`copy` / `copied` / `copy failed`) is
+  reflected in the label; `copied` auto-reverts after ~1s, `copy failed`
+  persists until the next re-render.
 
 ### Changed
 
--- a/docs/pipeline.md
+++ b/docs/pipeline.md
@@ -1524,6 +1524,18 @@ Cleanup
      delete the worktree, delete the remote branch, and close
      the PR. Each action is individually selectable.
 
+   When a squash suggestion is live in the PR body, the
+   merge-confirm screen also renders `[t] copy` / `[b] copy`
+   hotkey hints next to the suggested title and body.  Pressing
+   `t` or `b` writes the corresponding value to the system
+   clipboard (via `pbcopy` / `wl-copy` / `xclip` on local
+   sessions, or OSC 52 on SSH) so the user can paste it straight
+   into GitHub's "Squash and merge" dialog.  If the environment
+   can reach neither a native clipboard tool nor an OSC 52–capable
+   stdout, the hints are not rendered — the user falls back to
+   drag-select without being told about a feature that cannot
+   work here.
+
 **Agent rebase prompt:**
 
 ```text
--- a/src/i18n/en.ts
+++ b/src/i18n/en.ts
@@ -232,6 +232,9 @@ export const en: Messages = {
   // ---- input area --------------------------------------------------------
 
   "input.pipelineRunning": "Pipeline running...",
+  "input.copy": "copy",
+  "input.copied": "copied",
+  "input.copyFailed": "copy failed",
 
   // ---- agent pane / labels ------------------------------------------------
 
--- a/src/i18n/ko.ts
+++ b/src/i18n/ko.ts
@@ -270,6 +270,9 @@ export const ko: Messages = {
 
   "input.pipelineRunning":
     "\uD30C\uC774\uD504\uB77C\uC778 \uC2E4\uD589 \uC911...",
+  "input.copy": "복사",
+  "input.copied": "복사됨",
+  "input.copyFailed": "복사 실패",
 
   // ---- agent pane / labels ------------------------------------------------
 
--- a/src/i18n/messages.ts
+++ b/src/i18n/messages.ts
@@ -230,6 +230,9 @@ export interface Messages {
   // ---- input area (InputArea.tsx) ----------------------------------------
 
   "input.pipelineRunning": string;
+  "input.copy": string;
+  "input.copied": string;
+  "input.copyFailed": string;
 
   // ---- agent pane / labels ------------------------------------------------
 
--- a/src/pipeline.test.ts
+++ b/src/pipeline.test.ts
@@ -1108,6 +1108,10 @@ function makeDoneOpts(
     stopServices: vi.fn(),
     hasRunningServices: vi.fn().mockReturnValue(false),
     onNotMerged: vi.fn().mockResolvedValue(undefined),
+    // Default: clipboard unavailable so the confirmMerge message keeps
+    // the bare "Suggested title:\n<title>" shape tests historically
+    // assert on.  Individual tests override to exercise hotkey paths.
+    detectClipboardSupport: () => [],
     ...rest,
   };
 }
@@ -1152,6 +1156,127 @@ describe("createDoneStageHandler", () => {
     expect(message).toContain("\n\nSuggested title:");
     expect(message).toContain("\n\nSuggested body:");
     expect(message).toMatch(/Closes #5\n\nPR: https:\/\//);
+  });
+
+  test("MERGEABLE: clipboard available embeds sentinels and forwards hotkeys", async () => {
+    const confirmMerge = vi.fn().mockResolvedValue("merged");
+    const writeToClipboard = vi.fn().mockResolvedValue("ok");
+    const opts = makeDoneOpts({
+      prompt: { confirmMerge },
+      getSquashMergeHint: () => ({
+        title: "Fix widget rendering",
+        body: "Body line\n\nCloses #5",
+        prUrl: "https://github.com/org/repo/pull/123",
+      }),
+      detectClipboardSupport: () => ["pbcopy"],
+      writeToClipboard,
+    });
+    const stage = createDoneStageHandler(opts);
+    const ctx: StageContext = {
+      ...BASE_CTX,
+      iteration: 0,
+      lastAutoIteration: false,
+      userInstruction: undefined,
+    };
+    await stage.handler(ctx);
+    expect(confirmMerge).toHaveBeenCalledOnce();
+    const [message, hotkeys] = confirmMerge.mock.calls[0] as [
+      string,
+      { id: string; key: string; onPress: () => Promise<"ok" | "error"> }[],
+    ];
+    // Sentinel ids carry a per-render random suffix so agent-authored
+    // title/body text cannot spoof them.  Assert the prefix shape only.
+    const titleSentinel = message.match(/\{\{hk:title-[0-9a-f]+\}\}/);
+    const bodySentinel = message.match(/\{\{hk:body-[0-9a-f]+\}\}/);
+    expect(titleSentinel).not.toBeNull();
+    expect(bodySentinel).not.toBeNull();
+    expect(message).toContain(
+      `Suggested title: ${titleSentinel?.[0]}\nFix widget`,
+    );
+    expect(message).toContain(
+      `Suggested body: ${bodySentinel?.[0]}\nBody line`,
+    );
+    expect(hotkeys).toHaveLength(2);
+    expect(hotkeys[0].id).toMatch(/^title-[0-9a-f]+$/);
+    expect(hotkeys[0].key).toBe("t");
+    expect(hotkeys[1].id).toMatch(/^body-[0-9a-f]+$/);
+    expect(hotkeys[1].key).toBe("b");
+    // Sentinel ids in the message must match the ids in the hotkeys
+    // array so InputArea can resolve them.
+    expect(titleSentinel?.[0]).toBe(`{{hk:${hotkeys[0].id}}}`);
+    expect(bodySentinel?.[0]).toBe(`{{hk:${hotkeys[1].id}}}`);
+
+    await hotkeys[0].onPress();
+    expect(writeToClipboard).toHaveBeenCalledWith("Fix widget rendering", [
+      "pbcopy",
+    ]);
+    await hotkeys[1].onPress();
+    expect(writeToClipboard).toHaveBeenCalledWith("Body line\n\nCloses #5", [
+      "pbcopy",
+    ]);
+  });
+
+  test("MERGEABLE: sentinel-looking substrings in body are rendered literally", async () => {
+    const confirmMerge = vi.fn().mockResolvedValue("merged");
+    const opts = makeDoneOpts({
+      prompt: { confirmMerge },
+      // Body contains both fixed tokens from earlier iterations of the
+      // design and a literal use of the current sentinel prefix.  Neither
+      // should collide with the real per-render sentinel ids.
+      getSquashMergeHint: () => ({
+        title: "Docs: document {{hk:title}} placeholder",
+        body: "Explain {{hk:body}} and {{hk:title}} in the guide",
+      }),
+      detectClipboardSupport: () => ["pbcopy"],
+      writeToClipboard: vi.fn().mockResolvedValue("ok"),
+    });
+    const stage = createDoneStageHandler(opts);
+    const ctx: StageContext = {
+      ...BASE_CTX,
+      iteration: 0,
+      lastAutoIteration: false,
+      userInstruction: undefined,
+    };
+    await stage.handler(ctx);
+    const [message, hotkeys] = confirmMerge.mock.calls[0] as [
+      string,
+      { id: string; key: string; onPress: () => Promise<"ok" | "error"> }[],
+    ];
+    // The literal `{{hk:title}}` / `{{hk:body}}` substrings from the
+    // user-authored content must survive untouched (they still appear in
+    // the message text), and the real sentinels are distinct per-render
+    // ids with a random suffix.
+    expect(message).toContain("Docs: document {{hk:title}} placeholder");
+    expect(message).toContain(
+      "Explain {{hk:body}} and {{hk:title}} in the guide",
+    );
+    expect(hotkeys[0].id).toMatch(/^title-[0-9a-f]+$/);
+    expect(hotkeys[0].id).not.toBe("title");
+    expect(hotkeys[1].id).toMatch(/^body-[0-9a-f]+$/);
+    expect(hotkeys[1].id).not.toBe("body");
+  });
+
+  test("MERGEABLE: clipboard unavailable omits sentinels and forwards no hotkeys", async () => {
+    const confirmMerge = vi.fn().mockResolvedValue("merged");
+    const opts = makeDoneOpts({
+      prompt: { confirmMerge },
+      getSquashMergeHint: () => ({
+        title: "Fix widget",
+        body: "body",
+      }),
+      detectClipboardSupport: () => [],
+    });
+    const stage = createDoneStageHandler(opts);
+    const ctx: StageContext = {
+      ...BASE_CTX,
+      iteration: 0,
+      lastAutoIteration: false,
+      userInstruction: undefined,
+    };
+    await stage.handler(ctx);
+    const [message, hotkeys] = confirmMerge.mock.calls[0] as [string, unknown];
+    expect(message).not.toContain("{{hk:");
+    expect(hotkeys).toBeUndefined();
   });
 
   test("MERGEABLE: omits hint when getSquashMergeHint returns undefined", async () => {
--- a/src/pipeline.ts
+++ b/src/pipeline.ts
@@ -9,10 +9,17 @@
  * modules (issues #6, #7, #8).
  */
 
+import { randomBytes } from "node:crypto";
 import { t } from "./i18n/index.js";
 import type { PipelineEventEmitter } from "./pipeline-events.js";
 import type { PromptSink, StreamSink, UsageSink } from "./stage-util.js";
 import { buildClarificationPrompt } from "./step-parser.js";
+import {
+  type ClipboardCandidate,
+  detectClipboardSupport as defaultDetectClipboardSupport,
+  writeToClipboard as defaultWriteToClipboard,
+} from "./ui/clipboard.js";
+import type { InputHotkey } from "./ui/InputArea.js";
 
 // ---- public types --------------------------------------------------------
 
@@ -210,7 +217,10 @@ export interface UserPrompt {
    * - `"check_conflicts"` — check for conflicts and rebase if needed.
    * - `"exit"` — stop asking and proceed to cleanup options.
    */
-  confirmMerge(message: string): Promise<"merged" | "check_conflicts" | "exit">;
+  confirmMerge(
+    message: string,
+    hotkeys?: InputHotkey[],
+  ): Promise<"merged" | "check_conflicts" | "exit">;
 
   /**
    * Present a conflict notification and let the user choose between
@@ -883,6 +893,7 @@ export interface DoneStageOptions {
   prompt: {
     confirmMerge: (
       message: string,
+      hotkeys?: InputHotkey[],
     ) => Promise<"merged" | "check_conflicts" | "exit">;
     handleConflict: (message: string) => Promise<"agent_rebase" | "manual">;
     handleUnknownMergeable: (message: string) => Promise<"recheck" | "exit">;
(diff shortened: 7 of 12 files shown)


sehkone:
**[Reviewer Round 1]**

Overall this looks close to the issue: the candidate-list clipboard design matches the failure model well, `askMerge()` only renders hints when clipboard support is actually available, and the new tests cover most of the important paths.

I do see one requirement gap that I think should be fixed before merge:

- `src/ui/InputArea.tsx:106` checks hotkeys before numeric choices, but there is no guard that prevents a hotkey from reusing one of the active choice digits. The issue explicitly called out that `hotkeys[].key` must be checked for collisions with `choices`. As written, a request with `hotkeys: [{ key: "1", ... }]` would silently shadow choice `1` and change the prompt semantics. The current merge-confirm caller uses `t`/`b`, so this does not bite this screen today, but the new generic `InputRequest.hotkeys` API makes the footgun reusable elsewhere. I would either reject colliding hotkeys when the request is built or ignore them in `InputArea`, and add a regression test for that collision case.

sehkone:

sehkone:
**[Author Round 1]**

- **Fixed**: hotkey/choice-digit collision guard. `InputArea` now filters out any hotkey whose `key` is a digit in `[1, choices.length]` before wiring `useInput` or the sentinel lookup, so the numeric dispatch is no longer shadowed. The sentinel for a dropped hotkey falls through to the unknown-id path (literal text) so a buggy caller is visible in the frame rather than silently swallowed. Added a regression test in `components.test.tsx` (`hotkey colliding with an active choice digit is ignored…`) that asserts pressing `1` with `hotkeys: [{ key: "1", … }]` and `choices: [Yes, No]` calls `onSubmit("yes")` and does not fire `onPress`, and that `{{hk:one}}` renders as literal text.

Commit: 8a82156.

sehkone:
**[Reviewer Round 2]**

Thanks — the Round 1 collision issue looks fixed. `src/ui/InputArea.tsx:62` now drops hotkeys that would shadow active numeric choices, and `src/ui/components.test.tsx:1684` covers the regression.

I do still see two things I think should be addressed before merge:

- `src/ui/clipboard.ts:88` still treats macOS native clipboard support as conditional on `hasCommand("pbcopy")`. The issue text made the darwin path unconditional (`pbcopy` is the local-native choice on macOS), specifically so we do not hide or downgrade clipboard support just because PATH inspection is imperfect. With the current code, a stripped PATH that omits `/usr/bin` makes `detectClipboardSupport()` return only `osc52` — or `[]` if stdout is not a TTY — even though `/usr/bin/pbcopy` would still work. That means the merge-confirm hints can disappear or fall back to best-effort OSC 52 on a machine where the deterministic native path is actually available. I think darwin should always enqueue `pbcopy` (or at least check the absolute path), and there should be a test for the PATH-missing macOS case.

- `src/ui/InputArea.tsx:87` has a race on repeated hotkey presses. `runHotkey()` fires `hk.onPress()` and unconditionally applies whichever promise settles last. If the user presses `t` twice while the first clipboard attempt is still pending, an older result can overwrite the newer one: e.g. second press returns `"ok"`, label flips to `[t] copied`, then the first press resolves `"error"` and the UI regresses to `[t] copy failed`. Because native candidates can sit on the 1s timeout, this is reachable in practice. I’d track a per-hotkey sequence/token and only let the latest completion update status/timers, plus add a test that resolves two presses out of order.

sehkone:

sehkone:
**[Author Round 2]**

- **Fixed**: darwin `pbcopy` is now unconditional. `detectClipboardSupport` no longer gates `pbcopy` on `hasCommand("pbcopy")` — `pbcopy` ships with macOS at `/usr/bin/pbcopy`, so a stripped PATH (missing `/usr/bin`) must not hide or downgrade the deterministic native path. Added two regression tests in `clipboard.test.ts`: `macOS: pbcopy is unconditional even when PATH does not list it` and `macOS without stdout TTY: pbcopy still enqueued`. The SSH-plus-darwin case continues to put OSC 52 first and `pbcopy` last, as before.

- **Fixed**: repeated-hotkey-press race in `runHotkey`. Added a per-hotkey sequence counter (`hotkeySeqRef`): each press increments the token and captures it; a completion only applies status / timer changes if its captured token still matches the latest dispatched press. Stale older completions (e.g. the first press's late `"error"` arriving after the second press's `"ok"` has already flipped the label to `[t] copied`) are dropped. The ref is reset on request change and on unmount alongside the existing timer cleanup. Added a UI test `out-of-order completions: older press must not overwrite newer press result` that dispatches two `t` presses, resolves the second first with `"ok"`, resolves the first later with `"error"`, and asserts the frame still shows `[t] copied` and never `[t] copy failed`.

Commit: 41b303b.

sehkone:
**[Reviewer Round 3]**

Thanks — the repeated-hotkey race in `src/ui/InputArea.tsx` looks fixed.

I do still see one remaining issue from Round 2:

- `src/ui/clipboard.ts:82` now advertises `pbcopy` unconditionally on macOS, but the actual invocation path still shells out to bare `"pbcopy"` at `src/ui/clipboard.ts:180`. In the stripped-`PATH` case called out in Round 2, `detectClipboardSupport()` now returns `['pbcopy', ...]`, so the merge-confirm screen shows copy hints, but `writeToClipboard()` still hits `ENOENT` because `spawnSync()` resolves commands through `PATH`. That means the deterministic darwin fallback is still not actually reachable in the environment this change was meant to fix. I think the writer needs to execute `/usr/bin/pbcopy` (or otherwise bypass `PATH` for the darwin candidate), and there should be a regression test that asserts the darwin candidate invokes the absolute path rather than bare `pbcopy`.

sehkone:

sehkone:
**[Author Round 3]**

- **Fixed**: darwin `pbcopy` now shells out to `/usr/bin/pbcopy` instead of bare `"pbcopy"`. `detectClipboardSupport()` enqueues `pbcopy` unconditionally on macOS so a stripped `PATH` (e.g. missing `/usr/bin`) does not hide the deterministic native candidate, but the writer still went through `spawnSync("pbcopy", …)`, which resolves commands through `PATH` — so the exact environment the Round 2 detection change was meant to fix would still ENOENT at write time and the hint would advertise a copy that never lands. Using the absolute path of the stock macOS binary bypasses `PATH` entirely. Added a regression test `pbcopy invocation uses absolute /usr/bin path (bypasses PATH)` asserting the darwin candidate invokes `/usr/bin/pbcopy` and never bare `pbcopy`, and updated the existing `native write receives value via stdin` test to match the new argv.

Commit: e27d20a.

sehkone:
**[Reviewer Round 4]**

I re-checked the Round 3 item around macOS invocation, and that part looks resolved: `writeToClipboard()` now calls `/usr/bin/pbcopy`, so the stripped-`PATH` case is actually reachable.

I do still see one new correctness issue in the sentinel wiring:

- `src/pipeline.ts:1218` and `src/pipeline.ts:1232` embed fixed `{{hk:title}}` / `{{hk:body}}` tokens into a single `message` string that also contains the raw suggested title/body text, and `src/ui/InputArea.tsx:206` then substitutes **every** matching token it finds anywhere in that message. That means if the suggested title or body itself literally contains `{{hk:title}}` or `{{hk:body}}`, the displayed suggestion is corrupted because the value text gets rewritten into a hotkey label too. Since the squash text is free-form / agent-generated, this is reachable in practice (for example when the suggestion is documenting template syntax). The issue description called these “unique sentinel” tokens inserted by the caller; right now they are only unique by convention. I think `askMerge()` needs to generate per-render unique ids/tokens (or otherwise escape the interpolated title/body content before handing the string to `InputArea`), and there should be a regression test where the suggested body contains a sentinel-looking substring and is still rendered literally.

sehkone:

sehkone:
**[Author Round 4]**

- **Fixed**: sentinel ids are now per-render unique so user-authored title/body content cannot spoof the real hotkey hints. `askMerge` generates a 48-bit random hex suffix and embeds the full id (`title-<hex>` / `body-<hex>`) in both the `{{hk:<id>}}` tokens inside `message` and in the `hotkeys` array handed to `InputArea`. A short re-roll loop guarantees the generated suffix never appears verbatim in the interpolated title or body, so the invariant holds even under adversarial input. Literal `{{hk:title}}` / `{{hk:body}}` substrings inside the suggestion now fall through the unknown-id path and render as plain text — the user-visible suggestion is preserved exactly. Added a regression test `MERGEABLE: sentinel-looking substrings in body are rendered literally` that feeds a title of `"Docs: document {{hk:title}} placeholder"` and a body of `"Explain {{hk:body}} and {{hk:title}} in the guide"` through `askMerge` and asserts (a) both literal substrings still appear in the composed `message`, and (b) the real hotkey ids match `/^title-[0-9a-f]+$/` / `/^body-[0-9a-f]+$/` and are distinct from the bare `title` / `body` strings. Updated the pre-existing `clipboard available embeds sentinels and forwards hotkeys` test to assert the per-render shape (`{{hk:title-<hex>}}`) and that the sentinel id in `message` matches the id on each hotkey.

Commit: cf086fe.

sehkone:
**[Reviewer Round 5]**

I re-checked the Round 4 sentinel item, and that fix is now actually present: `askMerge()` generates per-render sentinel ids via `makeSentinelSuffix()` and threads those ids consistently into both the rendered `message` and the `hotkeys` array (`src/pipeline.ts:1224`, `src/pipeline.ts:1357`). The added regression coverage in `src/pipeline.test.ts:1219` also exercises literal `{{hk:title}}` / `{{hk:body}}` substrings in the suggested squash text, which was the missing guardrail last round.

I also did another pass over the updated diff against #265 and I don’t have any further issues to raise in this round.

sehkone:

PR #271, opened 2026-04-19
Remove legacy squash suggestion format parser (#267)

### LINKED ISSUE #267: Remove legacy parseSquashSuggestionBlock format

sehkone:
## Summary

- Delete `parseLegacySuggestion` and its call site in `parseSquashSuggestionBlock`, plus the companion `detectFormat` classifier that was only needed to route between the two formats.
- Drop the legacy-format tests in the `parseSquashSuggestionBlock` describe block and update the remaining test fixtures to use the fenced format.
- Update the docstring on `parseSquashSuggestionBlock` and related inline comments to drop the legacy paragraph / references.
- Consolidate `CHANGELOG.md`: remove the legacy acceptance bullet from `### Changed` and the `### Deprecated` block since #263 has not shipped yet, so there is no deprecation entry to migrate to `### Removed`.

The compatibility branch was kept for one release cycle as a safety net for PRs already in the `applied_in_pr_body` state written by the previous release. The maintainer has confirmed no such in-flight runs exist, so the branch is dead weight.

No `RUN_STATE_VERSION` bump is needed (`squashSubStep` is not format-coupled), and no migration shim is needed (a legacy PR body that somehow persists past removal only causes a cosmetic Stage 9 preview degradation, not a pipeline failure).

Closes #267

## Test plan

- [x] `pnpm vitest run src/stage-squash.test.ts` passes (70/70)
- [x] `pnpm vitest run` full suite passes (1858/1858)
- [x] Confirm `parseSquashSuggestionBlock` no longer parses a legacy `**Title:** …` / `**Body:** …` block and returns `undefined` for it
- [x] Confirm fenced-format parsing still works end-to-end through the Stage 8 → Stage 9 preview path

## Suggested squash commit

**Title**

```text
Remove legacy squash suggestion format parser
```

**Body**

```text
The legacy `**Title:** …` / `**Body:** …` plain-text branch in
`parseSquashSuggestionBlock` was kept for one release cycle as a
safety net so PRs already in the `applied_in_pr_body` state could
still render in the Stage 9 inline preview after upgrade. The
maintainer has confirmed no such in-flight runs exist, so the
branch is dead weight.

Delete `parseLegacySuggestion`, its `detectFormat` classifier, the
legacy-format tests, and the matching docstring / inline comment
paragraphs. Add a regression test that verifies the parser now
returns `undefined` for a well-formed legacy block. Drop the
companion CHANGELOG entries: the legacy-acceptance bullet under
`### Changed` and the whole `### Deprecated` block, since #263 has
not shipped yet and there is no deprecation entry to migrate to
`### Removed`.

No `RUN_STATE_VERSION` bump is required — `squashSubStep` is not
format-coupled — and no migration shim is needed, since a stray
legacy PR body past the removal only causes a cosmetic Stage 9
preview degradation, not a pipeline failure.

Closes #267
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,12 +62,6 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   dynamically per the CommonMark rule
   (`max(longest backtick run in content, 2) + 1`, minimum 3) so commit
   bodies containing their own triple-backtick samples survive unchanged.
-- `parseSquashSuggestionBlock` now accepts either the new fenced format or
-  the legacy `**Title:** …` / `**Body:** …` plain-text format, returning
-  the same `{ title, body }` shape.  This is required because Stage 8
-  short-circuits on `squashSubStep === "applied_in_pr_body"` without
-  re-invoking the agent, so existing PRs in that state still render
-  correctly in the Stage 9 inline preview after upgrade.
 - Stage 8 verdict keywords are now `SQUASHED_MULTI` / `SUGGESTED_SINGLE` /
   `BLOCKED` (previously `COMPLETED` / `BLOCKED`).  When the verdict is
   ambiguous after a clarification retry, the handler runs a deterministic
@@ -80,15 +74,6 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Stage 9 merge-confirm screen now includes a conditional one-line tip
   (`pipeline.mergeConfirmSquashTip`) when a squash suggestion is live in
   the PR body.
-
-### Deprecated
-
-- The legacy `**Title:** …` / `**Body:** …` plain-text form of the squash
-  suggestion block (written by older versions of Stage 8) is deprecated.
-  `parseSquashSuggestionBlock` continues to accept it for one release cycle
-  so PRs already in the `applied_in_pr_body` state still render in the
-  Stage 9 inline preview after upgrade.  The legacy branch will be removed
-  in the following release — tracked by #267.
 
 ### Fixed
 
@@ -121,10 +106,10 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   branch silently skipped telemetry.
 - Stage 8 now validates the squash suggestion block strictly before
   accepting `SUGGESTED_SINGLE` / `applied_in_pr_body`.  A bare start
-  marker or a block missing `**Title:**` / the end marker is treated
-  as malformed and fails closed (or re-runs planning on resume)
-  instead of completing with `squash.messageAppended` and leaving
-  Stage 9 unable to render the inline preview.
+  marker or a block missing the `**Title**` label / the end marker
+  is treated as malformed and fails closed (or re-runs planning on
+  resume) instead of completing with `squash.messageAppended` and
+  leaving Stage 9 unable to render the inline preview.
 - Stage 8 now persists the verdict turn's session id before
   transitioning to `awaiting_user_choice`.  Adapters can surface a
   new session id on follow-up turns, so the verdict session is not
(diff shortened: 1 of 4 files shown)


sehkone:
**[Reviewer Round 1]** I think the code change itself is the right simplification, but I’m missing one regression test that proves the compatibility branch is actually gone.

`src/stage-squash.test.ts:1593` now only covers the fenced format plus generic malformed/missing cases. There is no explicit case asserting that a legacy block like `**Title:** ...` / `**Body:** ...` now returns `undefined`. That matters because rejecting the legacy shape is the core behavior change in this PR, and a negative test for it would have failed before this patch while passing now.

I’d add a focused test in this describe block that feeds `parseSquashSuggestionBlock` a legacy-format marker block and expects `undefined`. That would pin the new contract directly instead of only implying it via fixture rewrites elsewhere.

sehkone:

sehkone:
**[Author Round 1]**

- **Fixed** — Added a focused negative test in the `parseSquashSuggestionBlock` describe block (`src/stage-squash.test.ts`) that feeds a well-formed legacy `**Title:** …` / `**Body:** …` block and asserts `parseSquashSuggestionBlock` returns `undefined`. This pins the new contract directly: the test would have passed (returning the parsed legacy block) before this PR and fails now if the legacy branch were reintroduced.

sehkone:
**[Reviewer Round 2]**

I verified the Round 1 item: the focused negative regression test is now present at `src/stage-squash.test.ts:1782`, and it directly pins the new contract by asserting that a legacy `**Title:** …` / `**Body:** …` block returns `undefined`.

I also re-reviewed the updated diff against #267. The parser cleanup in `src/stage-squash.ts:330` now routes only through fenced parsing, the parser/doc comments are aligned with that single supported format in `src/stage-squash.ts:322` and `docs/pipeline.md:1318`, and the changelog consolidation in `CHANGELOG.md:56` looks consistent with #263 not having shipped yet.

No remaining issues from me.

sehkone:

PR #273, opened 2026-04-19
Fix squash hotkey sentinels leaking to TUI (#272)

### LINKED ISSUE #272: Squash hotkey sentinels leak to TUI because index.ts confirmMerge wrapper drops hotkeys arg
## Symptom

On the Stage 9 merge-confirm prompt after a SUGGESTED_SINGLE squash path, the user sees raw `{{hk:title-<suffix>}}` / `{{hk:body-<suffix>}}` sentinel tokens rendered literally next to the `Suggested title:` / `Suggested body:` labels, instead of the `[t] copy` / `[b] copy` hotkey hints that #265 specified.

Example (paraphrased from a live run of #267):

```
Pipeline for aicers/agentcoop#267 completed.

If this repo allows 'Squash and merge', the suggested commit message is in the PR body.

Suggested title: {{hk:title-270bf96f70b4}}
Remove legacy squash suggestion format parser

Suggested body: {{hk:body-270bf96f70b4}}
The legacy `**Title:** …` / `**Body:** …` plain-text branch in
...
```

The sentinels are internal substitution markers that InputArea is supposed to replace with inline `[t] copy` / `[b] copy` hints; they must never appear as literal text in user-facing output.

## Root cause

The `confirmMerge` wrapper defined in `src/index.ts` (around the `createDoneStageHandler` call site, currently near line 735) only forwards the `msg` argument and silently drops the `hotkeys` argument that the pipeline passes in:

```ts
confirmMerge: async (msg) => {
  if (tuiPrompt) return tuiPrompt.confirmMerge(msg);
  return "merged";
},
```

Both sides of the interface are otherwise correct:

- `askMerge` in `src/pipeline.ts` (around line 1260) calls `options.prompt.confirmMerge(sections.join("\n\n"), hotkeys.length > 0 ? hotkeys : undefined)` — passes hotkeys.
- `tuiPrompt.confirmMerge` in `src/ui/TuiUserPrompt.ts:115` accepts `(message, hotkeys?)` — receives hotkeys.
- `InputArea` substitutes `{{hk:<id>}}` against the received hotkeys list — works when hotkeys arrive.

The wrapper in the middle is the single point where hotkeys are discarded. Consequences:

- `tuiPrompt.confirmMerge` is invoked with `hotkeys = undefined`.
- The prompt's `InputRequest.hotkeys` is `undefined`.
- `InputArea`'s `hotkeyById` map is empty.
- Every `{{hk:<id>}}` sentinel lookup returns `undefined`.
- Per the "unknown id falls through as literal text" branch in `renderMessageLines` ([src/ui/InputArea.tsx](https://github.com/aicers/agentcoop/blob/main/src/ui/InputArea.tsx), the `if (!hk) return <Text>{part.raw}</Text>` branch) — which exists specifically to surface caller bugs — the raw sentinel string is rendered.

So the sentinel-leaking behavior is actually InputArea correctly surfacing a wiring bug. The fix is at the wrapper.

## Fix

Forward both arguments in the wrapper:

```ts
confirmMerge: async (msg, hotkeys) => {
  if (tuiPrompt) return tuiPrompt.confirmMerge(msg, hotkeys);
  return "merged";
},
```

One-line change. No other site needs updating — the pipeline, prompt interface, UI component, and clipboard utility are all correct.

## Test coverage gap

#265's existing tests in `src/pipeline.test.ts` exercise `askMerge` with a `confirmMerge` mock that accepts `(msg, hotkeys)` directly, and the UI tests in `src/ui/components.test.tsx` exercise `InputArea` with `hotkeys` passed in. Neither covers the `index.ts` wrapper that sits between them — which is exactly where the bug sits.

### Why not "extract and unit-test a helper"

The obvious-sounding fix is to extract the wrapper into a named helper (e.g. `makeConfirmMergeAdapter`) and unit-test it in isolation. This is rejected for this issue because:

- The wrapper is 1 line. Extraction is scope creep for a 1-line bug fix.
- The three sibling wrappers in the same options object (`handleConflict`, `handleUnknownMergeable`, `waitForManualResolve`) share the same inline-anonymous pattern. Extracting only `confirmMerge` creates asymmetry; extracting all four is a separate refactor that should not ride along with a bug fix.
- A unit-tested helper that correctly forwards arguments does **not** prevent the exact failure mode observed here: the bug was in how the helper is wired into the options object at the `createDoneStageHandler` call site, not in the helper's internal logic. A unit test would have passed while the real code still leaked sentinels.

### What the regression test should cover

An **integration-style test** exercising the assembled options object is what matches the failure mode. Concretely:

- Arrange the `createDoneStageHandler({ ... prompt: { confirmMerge, ... } ... })` assembly that `index.ts` performs, substituting `tuiPrompt.confirmMerge` with a spy.
- Act: invoke the assembled `prompt.confirmMerge(message, hotkeys)` with a non-empty `hotkeys` array.
- Assert: the spy was invoked with two arguments, and the second argument (deep-) equals the `hotkeys` array the caller passed.

The "deep-equal" wording is deliberate — it is strong enough to reject a wrong implementation that forwards a stripped value (e.g. `tuiPrompt.confirmMerge(msg, [])` or `tuiPrompt.confirmMerge(msg, undefined)`) without over-constraining future implementations that might immutably clone or wrap the array (which would not be a bug on its own).

Either of the following test layouts is acceptable; pick whichever causes least change to the existing test topology:

1. A focused test file (e.g. `src/index.test.ts` or `src/index-wiring.test.ts`) that imports the wiring factory (extracting just enough of the assembly code to test — *not* extracting the wrapper itself) and asserts forwarding.
2. A case added to an existing `index.ts`-adjacent test file that covers the `createDoneStageHandler` assembly path end-to-end for the Stage 9 TUI scenario.

No TUI render is required — asserting on the spy's invocation arguments is sufficient.

## Out of scope

- Changes to the sentinel format, the clipboard detection policy, or the hotkey feedback UX — all working as designed once the wrapper is fixed.
- Refactoring the three sibling wrappers (`handleConflict`, `handleUnknownMergeable`, `waitForManualResolve`) to a shared helper. If that cleanup is desired, track it as a separate issue after this bug is fixed.
- Non-TUI code paths (i.e. the `return "merged"` fallback when `tuiPrompt` is absent). Sentinels only reach the user via the TUI path.

sehkone:
## Summary

The `confirmMerge` wrapper inside `createDoneStageHandler({ prompt: { ... } })` in `src/index.ts` accepted only the `msg` argument and dropped the `hotkeys` argument the pipeline passes. As a result `tuiPrompt.confirmMerge` was invoked with `hotkeys = undefined`, `InputArea`'s `hotkeyById` map was empty, and every `{{hk:title-...}}` / `{{hk:body-...}}` sentinel token fell through to the "unknown id" branch in `renderMessageLines` and was rendered as literal text next to the `Suggested title:` / `Suggested body:` labels on the Stage 9 merge-confirm prompt.

The fix forwards both `msg` and `hotkeys` through the wrapper.

The prompt-options assembly is extracted into `src/done-prompt-options.ts` (`createDonePromptOptions(getTuiPrompt)`) so a regression test can exercise the real production wiring instead of a hand-copied replica. The factory takes a late-bound getter (`() => tuiPrompt | undefined`) and re-reads it on every wrapper invocation, preserving the `src/index.ts` lifecycle where `tuiPrompt` is `undefined` at factory-call time and only populated later via `onPromptReady`. It keeps the four sibling wrappers inline — the extraction is of the wiring site, not of the wrappers themselves.

Regression coverage consists of three tests in `src/pipeline.test.ts`:
1. A Stage 9 MERGEABLE integration test that drives the real factory through `createDoneStageHandler` end-to-end and confirms hotkeys arrive at `tuiPrompt.confirmMerge`.
2. A focused factory-level test that invokes the assembled `prompt.confirmMerge(message, hotkeys)` with a known `InputHotkey[]` and asserts **deep equality** on the spy's second argument, pinning `id` / `key` / `onPress` — not just `.key` shape.
3. A late-binding test that flips the getter's ref from populated to `undefined` between calls and asserts the wrapper re-reads it each time rather than snapshotting.

Addresses the Round 2 review that correctly pointed out the first extraction snapshotted `tuiPrompt` too early and permanently routed Stage 9 callbacks to the non-TUI fallback, and the Round 3 review that correctly pointed out the initial regression test only checked hotkey shape and would pass for wrappers that forward malformed payloads.

Closes #272

## Test plan

- [x] `pnpm vitest run` — 1862 tests pass
- [x] `pnpm check` (biome ci) — clean
- [x] `pnpm build` — clean
- [x] Stage 9 SUGGESTED_SINGLE squash path renders `[t] copy` / `[b] copy` instead of raw `{{hk:...}}` sentinels — covered by the combination of the three forwarding regression tests (`src/pipeline.test.ts`) plus the existing InputArea sentinel-substitution test (`src/ui/components.test.tsx`). Together they cover the full path from pipeline → late-bound wrapper → tuiPrompt → InputArea render. A live Stage 9 TUI run requires a real GitHub PR plus live Claude/Codex CLIs through the full pipeline and was not exercised end-to-end in this verification.
- [x] Pressing `[t]` / `[b]` invokes the clipboard write — covered by the existing InputArea hotkey `onPress` tests (`src/ui/components.test.tsx`). Live key-press confirmation in the TUI was not exercised.

## Suggested squash commit

**Title**

```text
Forward hotkeys through confirmMerge wrapper
```

**Body**

```text
The confirmMerge wrapper in the createDoneStageHandler options
object dropped the hotkeys argument passed by askMerge, so the
TUI received hotkeys=undefined and InputArea's hotkeyById map
was empty. Every {{hk:<id>}} sentinel fell through to the
"unknown id" literal-text branch in renderMessageLines, leaking
raw sentinel tokens next to the "Suggested title:" / "Suggested
body:" labels on the Stage 9 merge-confirm prompt.

Forward both arguments in the wrapper. Extract the prompt
options assembly into src/done-prompt-options.ts with a
late-bound getter so a regression test can drive the real
production wiring through createDoneStageHandler. The getter
indirection preserves the index.ts lifecycle where tuiPrompt is
assigned after onPromptReady fires.

Regression coverage in src/pipeline.test.ts: a Stage 9 MERGEABLE
end-to-end test, a factory-level test that deep-equals the
forwarded hotkeys payload, and a late-binding test that flips
the getter between calls.

Closes #272
```

### DIFF
--- a/src/done-prompt-options.ts
+++ b/src/done-prompt-options.ts
@@ -1,0 +1,57 @@
+import type { DoneStageOptions, UserPrompt } from "./pipeline.js";
+
+/**
+ * Minimal subset of {@link UserPrompt} that the Done stage
+ * prompt-options factory relies on.  Matches the structural shape of
+ * the `tuiPrompt` reference held in `src/index.ts`, which is itself
+ * populated by a subset of TuiUserPrompt's methods at mount time.
+ */
+export type DoneStageTuiPrompt = Pick<
+  UserPrompt,
+  | "confirmMerge"
+  | "handleConflict"
+  | "handleUnknownMergeable"
+  | "waitForManualResolve"
+>;
+
+/**
+ * Build the `prompt` options object for {@link createDoneStageHandler}
+ * from a late-bound `tuiPrompt` getter.  Each field is an inline
+ * wrapper that reads the current `tuiPrompt` via the getter at
+ * invocation time, delegating to it when present and otherwise
+ * returning a safe default for non-TUI environments.
+ *
+ * The getter indirection is required because `src/index.ts` assembles
+ * `createDoneStageHandler` before the ink <App> mounts and supplies
+ * the prompt via `onPromptReady`.  Passing a live value here would
+ * snapshot `undefined` and permanently route every Stage 9 callback
+ * to the non-TUI fallback, even after the prompt is assigned.
+ *
+ * Regression for #272: the `confirmMerge` wrapper previously dropped
+ * the `hotkeys` argument, leaking `{{hk:...}}` sentinels to the TUI.
+ */
+export function createDonePromptOptions(
+  getTuiPrompt: () => DoneStageTuiPrompt | undefined,
+): DoneStageOptions["prompt"] {
+  return {
+    confirmMerge: async (msg, hotkeys) => {
+      const tuiPrompt = getTuiPrompt();
+      if (tuiPrompt) return tuiPrompt.confirmMerge(msg, hotkeys);
+      return "merged";
+    },
+    handleConflict: async (msg) => {
+      const tuiPrompt = getTuiPrompt();
+      if (tuiPrompt) return tuiPrompt.handleConflict(msg);
+      return "manual";
+    },
+    handleUnknownMergeable: async (msg) => {
+      const tuiPrompt = getTuiPrompt();
+      if (tuiPrompt) return tuiPrompt.handleUnknownMergeable(msg);
+      return "exit";
+    },
+    waitForManualResolve: async (msg) => {
+      const tuiPrompt = getTuiPrompt();
+      if (tuiPrompt) return tuiPrompt.waitForManualResolve(msg);
+    },
+  };
+}
--- a/src/index.ts
+++ b/src/index.ts
@@ -25,6 +25,7 @@ import {
   assembleSquashStage,
   loadConfig,
 } from "./config.js";
+import { createDonePromptOptions } from "./done-prompt-options.js";
 import { getGitHubUsername, getIssue } from "./github.js";
 import { initI18n, t } from "./i18n/index.js";
 import {
@@ -731,23 +732,7 @@ try {
     events: emitter,
     checkMergeable: async () => checkMergeable(owner, repo, wt.branch),
     queryPrState: async () => queryPrState(owner, repo, wt.branch),
-    prompt: {
-      confirmMerge: async (msg) => {
-        if (tuiPrompt) return tuiPrompt.confirmMerge(msg);
-        return "merged";
-      },
-      handleConflict: async (msg) => {
-        if (tuiPrompt) return tuiPrompt.handleConflict(msg);
-        return "manual";
-      },
-      handleUnknownMergeable: async (msg) => {
-        if (tuiPrompt) return tuiPrompt.handleUnknownMergeable(msg);
-        return "exit";
-      },
-      waitForManualResolve: async (msg) => {
-        if (tuiPrompt) return tuiPrompt.waitForManualResolve(msg);
-      },
-    },
+    prompt: createDonePromptOptions(() => tuiPrompt),
     rebaseOntoMain: createRebaseHandler(agentA, defaultBranch),
     pollCiAndFix: async (ctx) => {
       return pollCiAndFix({
--- a/src/pipeline.test.ts
+++ b/src/pipeline.test.ts
@@ -1,4 +1,5 @@
 import { describe, expect, test, vi } from "vitest";
+import { createDonePromptOptions } from "./done-prompt-options.js";
 import type {
   PipelineOptions,
   StageContext,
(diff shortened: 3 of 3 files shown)


sehkone:
**[Reviewer Round 1]**

The one-line fix in `src/index.ts:735` looks right, but I don't think the new test actually protects the regression this issue is about.

At `src/pipeline.test.ts:1299`, the test builds a fresh local `prompt` object and re-implements the `confirmMerge` wrapper inside the test itself. The assertion at `src/pipeline.test.ts:1341` therefore never exercises the production wiring in `src/index.ts:734`. If `src/index.ts` regressed back to `confirmMerge: async (msg) => ...`, this test would still pass, because the copied wrapper in the test would continue forwarding `hotkeys` correctly.

That makes this more of a second implementation of the fix than a regression test for the real failure path. Given the bug lives specifically in the `createDoneStageHandler({ ..., prompt: { ... } })` assembly in `src/index.ts`, I think the test needs to hit that actual assembly path (or a very small extracted wiring factory that `src/index.ts` uses) instead of a hand-copied replica. Otherwise the test can drift out of sync with the call site it's meant to guard.

sehkone:

sehkone:
**[Author Round 1]**

**Fixed** — the regression test now exercises the real production wiring.

You were right: the previous test built a fresh local `prompt` object and re-implemented the `confirmMerge` wrapper inside the test body, so a regression in `src/index.ts` would not have been caught.

Changes in this push:

- Extracted the Stage 9 prompt-options assembly from `src/index.ts` into a new module `src/done-prompt-options.ts` exporting `createDonePromptOptions(tuiPrompt)`. `src/index.ts` now calls this factory at the `createDoneStageHandler({ prompt: ... })` call site. This matches the issue's guidance of "extracting just enough of the assembly code to test — not extracting the wrapper itself" (the four inline wrappers stay bundled inside the factory so this is not a per-wrapper refactor).
- Updated the regression test in `src/pipeline.test.ts` to import `createDonePromptOptions` directly and pass its result as `prompt:` to `makeDoneOpts`. If the factory ever regresses to dropping `hotkeys`, the spy assertion fails. The test no longer contains a hand-copied replica that could drift out of sync with production.

sehkone:
**[Reviewer Round 2]**

The new extracted factory fixes the hand-copied-test problem, but it introduces a production regression in the actual `index.ts` wiring.

At `src/index.ts:683`, `tuiPrompt` is intentionally a late-bound ref populated only in `onPromptReady` at `src/index.ts:902`. But `createDoneStageHandler` now snapshots it earlier via `prompt: createDonePromptOptions(tuiPrompt)` at `src/index.ts:735`. Inside `src/done-prompt-options.ts:28`, the wrappers close over that parameter value, not over the outer variable, so when the Stage 9 handlers eventually run they still see the original `undefined` and take the non-TUI fallbacks.

That means `confirmMerge` now returns `"merged"` without showing the TUI prompt, and the sibling wrappers also silently fall back to `"manual"`, `"exit"`, or `undefined`. So the Stage 9 TUI path this issue is trying to fix is no longer actually wired up.

The new regression test in `src/pipeline.test.ts:1296` does not catch this because it calls `createDonePromptOptions(tuiPrompt)` with an already-initialized prompt object. It still misses the real lifecycle from `src/index.ts`, where the factory is called before `tuiPrompt` is assigned.

I think this needs one more change: preserve the late-binding behavior when extracting the factory (for example by passing a getter/ref and reading it inside each wrapper), and add a test that exercises that `undefined`-then-assigned path rather than instantiating the factory with a live prompt up front.

sehkone:

sehkone:
**[Author Round 2]**

**Fixed** — the factory now late-binds `tuiPrompt` via a getter, and the tests exercise the real `undefined`-then-assigned lifecycle.

You were right about both the production regression and the test gap:

- `src/index.ts:683` declares `tuiPrompt` as a late-bound `let` populated in `onPromptReady` at `src/index.ts:902`. `createDonePromptOptions(tuiPrompt)` snapshotted the initial `undefined` value, so at Stage 9 every wrapper (`confirmMerge`, `handleConflict`, `handleUnknownMergeable`, `waitForManualResolve`) still saw `undefined` and routed to the non-TUI fallback — `"merged"` / `"manual"` / `"exit"` / no-op. The Stage 9 TUI prompt was never shown, which is a strict regression from the prior inline wrapper that closed over the outer variable.
- The previous regression test missed this because it built the factory with a live prompt already in hand.

Changes in this push:

- `src/done-prompt-options.ts`: signature changes from `createDonePromptOptions(tuiPrompt)` to `createDonePromptOptions(getTuiPrompt: () => DoneStageTuiPrompt | undefined)`. Each of the four wrappers now calls the getter on every invocation and branches on the current value. Updated the JSDoc to explain why the getter indirection is required (so the next reader does not "simplify" it back into a snapshot).
- `src/index.ts:735`: call site changes to `createDonePromptOptions(() => tuiPrompt)`, restoring the original late-binding behaviour of the inline wrappers it replaced.
- `src/pipeline.test.ts`: the existing regression test now builds the factory with a getter closing over `let tuiPromptRef: UserPrompt | undefined`, assigns `tuiPromptRef` only *after* `makeDoneOpts` has installed the factory's output into the options object, and then runs the stage. This mirrors the real `src/index.ts` order-of-operations. If the factory is reverted to a snapshot, the spy sees zero invocations and the test fails.
- Added a second focused test `MERGEABLE: done-prompt-options wrappers re-read the getter on each call` that:
  1. Starts with the ref populated, calls `confirmMerge`, asserts the spy was invoked.
  2. Flips the ref back to `undefined`, calls `confirmMerge` again, asserts the fallback `"merged"` is returned and the spy count has not increased. This pins per-call re-reads independent of the Stage 9 pipeline machinery.

Verification:
- `pnpm tsc --noEmit` — clean
- `pnpm check` (biome ci) — clean
- `pnpm vitest run` — 1861 pass (was 1860; the new late-binding test adds one)
- `pnpm build` — clean

sehkone:
**[Reviewer Round 3]**

The late-binding fix in `src/done-prompt-options.ts:33` and the `src/index.ts:735` call-site update look right, and they resolve the production regression I called out in Round 2.

I still see one remaining test gap, though:

- `src/pipeline.test.ts:1335` says this is a "Strong-form" check, but the assertions at `src/pipeline.test.ts:1338` through `src/pipeline.test.ts:1341` only verify that `hotkeys` is a 2-element array whose `.key`s are `"t"` / `"b"`. That still lets a broken wrapper pass while the TUI remains broken — for example, if the wrapper forwards objects with the right `key`s but drops or mangles `id` or `onPress`, this test still passes even though `InputArea` cannot substitute the sentinels because the real contract depends on the full `InputHotkey` payload, especially `id` (`src/ui/InputArea.tsx:14`).
- I think the forwarding regression test should pin the full second argument via deep equality, not just its shape. The simplest version is a focused factory-level case that calls the assembled `prompt.confirmMerge(message, hotkeys)` with a known `InputHotkey[]` and asserts the spy sees the same payload. That would preserve the useful late-binding coverage you added while also guarding the exact forwarding contract Stage 9 relies on.

sehkone:

sehkone:

sehkone:
**[Author Round 3]**

**Fixed** — the forwarding contract is now pinned by value, not shape.

You were right: the assertions at `src/pipeline.test.ts:1338`–`1341` only checked that the forwarded value was a 2-element array with `.key` values `"t"` / `"b"`. A broken wrapper that forwarded objects missing `id` or `onPress` would still pass, even though `InputArea` cannot substitute sentinels without `id` (`src/ui/InputArea.tsx:14`) and cannot run the `[t]` / `[b]` clipboard actions without `onPress`.

Changes in this push:

- Added a focused factory-level test `done-prompt-options: confirmMerge deep-forwards the full InputHotkey[] payload` in `src/pipeline.test.ts`. It builds a known `InputHotkey[]` (`id`, `key`, `onPress` on each entry), invokes the assembled `promptOpts.confirmMerge(message, hotkeys)` directly, and asserts the spy's second argument via `toEqual` — deep equality on the full payload.
- Left the existing Stage 9 integration test and the late-binding test unchanged. The integration test still provides end-to-end coverage from the pipeline through the factory, and the late-binding test still pins the `undefined`-then-assigned lifecycle. The new test complements them by guarding the forwarding contract at the factory layer.
- Dropped a duplicated JSDoc block above the integration test that had been accidentally repeated.

The comment above the new test explains the choice of `toEqual` (deep equality) over `toBe` (reference equality): it rejects the specific regressions called out in the issue (`[]`, `undefined`, stripped fields) without over-constraining future implementations that might immutably clone or wrap the array — matching the issue's own "deep-equal" wording.

Verification:
- `pnpm tsc --noEmit` — clean
- `pnpm check` (biome ci) — clean
- `pnpm vitest run` — 1862 pass (was 1861; the new deep-equality test adds one)
- `pnpm build` — clean

sehkone:
**[Reviewer Round 4]**

I re-checked the remaining Round 3 item and the updated diff. The production wiring now forwards `hotkeys` through `createDonePromptOptions` in `src/done-prompt-options.ts:37`, and `src/index.ts:735` uses that same factory for the real `createDoneStageHandler` assembly path.

The new focused test at `src/pipeline.test.ts:1344` closes the gap I called out in Round 3: it invokes the assembled `prompt.confirmMerge(message, hotkeys)` and asserts deep equality on the forwarded `InputHotkey[]`, so it would fail if `id` or `onPress` were dropped or mangled. Combined with the late-binding coverage at `src/pipeline.test.ts:1298` and `src/pipeline.test.ts:1374`, this now meaningfully guards both the original sentinel-leak regression and the lifecycle bug from Round 2.

I do not see any remaining issues for #272.

[Review Verdict Round 4: APPROVED]

sehkone:
[Review Verdict Round 4: APPROVED]

PR #275, opened 2026-04-19
Post squash suggestion as PR comment and ask apply policy at startup (#274)

### LINKED ISSUE #274: Post squash suggestion as PR comment and ask apply policy at startup
Two related changes to the single-commit squash flow.

## 1. Post the squash suggestion as a PR comment, not in the PR body

When stage 8 takes the SUGGESTED_SINGLE path, the agent currently writes the suggested commit title/body into a marker-delimited block inside the PR body (see `SQUASH_SUGGESTION_START_MARKER` / `SQUASH_SUGGESTION_END_MARKER` in [src/stage-squash.ts](https://github.com/aicers/agentcoop/blob/main/src/stage-squash.ts#L56-L59) and the prompt builder around [L180-L235](https://github.com/aicers/agentcoop/blob/main/src/stage-squash.ts#L180-L235)).

### Problem

The PR body sits at the top of the page on GitHub, while the "Squash and merge" dropdown — the button the user will actually paste into — sits at the bottom, next to the latest activity. Having to scroll up to the PR body to use the click-to-copy icon, then scroll back down to the merge dropdown, is avoidable friction.

### Proposal

Write the suggestion as a **PR comment** instead of editing the PR body. Comments render in chronological order and, for an active PR, the squash-suggestion comment will be the last thing above the "Squash and merge" box — one short scroll away, not a full page.

Notes on the mechanics:

- The same fenced-code-block layout from #263 carries over — GitHub's one-click copy icon works identically on comments.
- The marker-delimited block still needs to be idempotent: the agent must find its own prior comment (match by the start marker) and update it rather than posting a new one on each iteration. A fresh comment per iteration would clutter the timeline and push older ones up, defeating the proximity goal.
- On first use, create via `gh pr comment <pr> --body "..."`. On retries, locate the existing comment (list with `gh api repos/{owner}/{repo}/issues/{pr}/comments --jq '.[] | select(.body | contains(\"<start-marker>\"))'`) and edit it via `gh api --method PATCH /repos/{owner}/{repo}/issues/comments/{id}`.
- Reuse the existing [src/pr-comments.ts](https://github.com/aicers/agentcoop/blob/main/src/pr-comments.ts) infrastructure (`PrComment` type, `execFileSync gh api` helpers) — add a "find latest comment containing marker" helper there rather than rolling a new fetcher in stage-squash.
- The PR body is left alone entirely — no marker block there. Existing blocks on open PRs are not migrated; they just stop being updated.
- Parsing for `parseSquashSuggestion` (or whatever survives after #267) now reads comment bodies via `gh api`, not the PR body.

### Affected call sites

This is not a prompt-only change. All of the following read, write, or describe the suggestion location and must flip together:

**Agent-visible prompts** (most important — mismatched instructions here will confuse the agent or cause it to fall back to the old location):

- [src/stage-squash.ts:180-235](https://github.com/aicers/agentcoop/blob/main/src/stage-squash.ts#L180-L235) — main prompt that tells the agent where to write.
- [src/stage-squash.ts:280-295](https://github.com/aicers/agentcoop/blob/main/src/stage-squash.ts#L280-L295) — `buildSquashCompletionCheckPrompt`: currently says "SUGGESTED_SINGLE — if … you wrote the suggested title/body into the PR body marker block". Must say "into a PR comment" (or wording that matches the new location).
- [src/stage-squash.ts:303-313](https://github.com/aicers/agentcoop/blob/main/src/stage-squash.ts#L303-L313) — `buildAgentSquashFollowupPrompt`: currently references "the title and body you wrote into the PR body marker block" and "leave the marker block in the PR body". Both phrases must be updated for the comment location.

**Code that reads the suggestion:**

- [src/stage-squash.ts:574](https://github.com/aicers/agentcoop/blob/main/src/stage-squash.ts#L574) — handler validates suggestion presence via `getPrBody`. Must read comments instead.
- [src/index.ts:752](https://github.com/aicers/agentcoop/blob/main/src/index.ts#L752) — `getSquashMergeHint` in the Stage 9 merge-confirm path, which re-parses the PR body to surface the title/body in the TUI. Must read comments instead.

**User-facing i18n copy** (both `squash.*` and `pipeline.*` namespaces — do not limit the sweep to `squash.*`):

- `squash.messageAppended` in [src/i18n/en.ts:266-268](https://github.com/aicers/agentcoop/blob/main/src/i18n/en.ts#L266-L268) and [src/i18n/ko.ts:308](https://github.com/aicers/agentcoop/blob/main/src/i18n/ko.ts#L308) — both say "written to PR body" / 해당 한국어 표현. Must describe the comment location.
- `pipeline.mergeConfirmSquashTip` in [src/i18n/en.ts:145-147](https://github.com/aicers/agentcoop/blob/main/src/i18n/en.ts#L145-L147) and [src/i18n/ko.ts:180](https://github.com/aicers/agentcoop/blob/main/src/i18n/ko.ts#L180) — "the suggested commit message is in the PR body." This sits under `pipeline.*`, not `squash.*`, so a narrow-scoped sweep will miss it.
- Grep `src/i18n/*.ts` for any remaining squash-flow strings that mention "PR body" / "PR 본문" and update them together.

**Type / interface doc comments** (documenting behavior — wrong wording becomes a trap for the next reader):

- [src/pipeline.ts:250-256](https://github.com/aicers/agentcoop/blob/main/src/pipeline.ts#L250-L256) — `chooseSquashApplyMode` doc comment says the agent "has written to the PR body".
- [src/pipeline.ts:936-945](https://github.com/aicers/agentcoop/blob/main/src/pipeline.ts#L936-L945) — `getSquashMergeHint` doc comment says title/body are "parsed from the PR body" (two places).
- Also sweep the file-level doc comment at [src/stage-squash.ts:6](https://github.com/aicers/agentcoop/blob/main/src/stage-squash.ts#L6) and the marker-delimiter comment at [src/stage-squash.ts:55](https://github.com/aicers/agentcoop/blob/main/src/stage-squash.ts#L55) for the same wording.

**Sub-step label:**

- `applied_in_pr_body` (used at [src/stage-squash.ts:597](https://github.com/aicers/agentcoop/blob/main/src/stage-squash.ts#L597) and in [src/run-state.ts](https://github.com/aicers/agentcoop/blob/main/src/run-state.ts)) becomes a misnomer. Rename to something location-neutral like `applied_via_github` when flipping the source of truth.

### Scope

- Update all three agent-visible prompts (main + completion check + follow-up) consistently — mismatched wording is the main risk.
- Add a marker-lookup helper to `src/pr-comments.ts` and route Stage 8 validation + Stage 9 hint through it.
- Rename the sub-step literal from `applied_in_pr_body` and migrate any persisted run-state values (or treat old value as equivalent for one release).
- Update both `squash.*` and `pipeline.*` i18n strings in `en.ts` and `ko.ts` that describe the suggestion location.
- Sweep `pipeline.ts` and `stage-squash.ts` doc comments for "PR body" phrasing tied to the squash suggestion and update.
- Update `src/stage-squash.test.ts` / Stage 9 tests that fixture the PR body to fixture a comment list instead. The test churn here is expected to exceed the production-code churn.

### Out of scope

- Migrating suggestion blocks that already live in PR bodies on open PRs. The old location just stops being updated.

## 2. Ask the squash-apply policy once at startup

When the verdict is SUGGESTED_SINGLE, the handler currently calls `chooseSquashApplyMode` on every run (see [src/stage-squash.ts:110](https://github.com/aicers/agentcoop/blob/main/src/stage-squash.ts#L110) and the dispatcher at [src/ui/TuiUserPrompt.ts:175-185](https://github.com/aicers/agentcoop/blob/main/src/ui/TuiUserPrompt.ts#L175-L185)), which interrupts the pipeline with an `agent` vs `github` choice.

### Problem

For most runs the answer is "just let the agent squash it". Interrupting the pipeline every time for a question whose answer rarely changes is noisy. But the answer *does* change occasionally — e.g. when the user wants to inspect the individual commits in GitHub's squash dialog before pressing merge — so a persisted global setting is wrong too.

### Proposal

Add one more question to the startup flow, **after** `selectTarget` finishes and **before** bootstrap begins (conceptually between [startup.ts:144](https://github.com/aicers/agentcoop/blob/main/src/startup.ts#L144) and the bootstrap call):

> When a single-commit squash is suggested, apply it automatically via the agent, or ask each time?
> [Y] let the agent handle it (default)
> [n] ask me each time (current behavior)

Semantics:

- **Not persisted.** Asked on every run. Rationale: the right answer depends on the state of the current branch / PR — how much review the commits need, whether the user plans to review in GitHub's squash dialog, etc. — which is a per-run judgment, not a stable preference.
- **Default is "let the agent handle it"**, consistent with auto-mode expectations. A bare Enter keypress during startup picks the low-friction path.
- When set to "auto", the stage-squash handler skips `chooseSquashApplyMode` entirely and proceeds as if the user had picked `"agent"`. No prompt, no pause.
- When set to "ask each time", behavior is unchanged from today.
- The startup question itself should sit near the existing startup prompts, not get its own screen; a single confirm-style prompt is enough.
- Quick-start reuse (the "use previous config?" path at [startup.ts:194](https://github.com/aicers/agentcoop/blob/main/src/startup.ts#L194)) still asks this question — it is explicitly *not* part of the reused config.
- **Resume path also asks this question.** The resume branch at [src/index.ts:351](https://github.com/aicers/agentcoop/blob/main/src/index.ts#L351) currently bypasses `runStartup()` entirely and rehydrates `RunParams` from saved state. "Per-run" must mean every run — including resumes — so the policy prompt runs on the resume branch too, not inside `runStartup`. Place it at the shared join point after both the resume and fresh branches have produced `params`, so both paths go through the same prompt with no duplication. The value is not persisted in `RunState`; it is collected fresh each time.

### Scope

- Thread a `squashApplyPolicy: "auto" | "ask"` value through the pipeline boot code into `SquashStageOptions`. Since resume also needs to set it, this value lives alongside `params`, not only in `StartupResult`.
- In stage-squash, branch on the policy before dispatching `chooseSquashApplyMode`.
- i18n strings for the new startup question (en + ko).
- Tests for both policy values, covering both the fresh-startup path and the resume path.

### Out of scope

- Persisting the policy to config or `RunState`.
- A per-PR override mechanism beyond the startup question (if the user wants different policy for a different PR, they start a new run).

sehkone:
## Summary

Two changes to the single-commit squash flow, landed together because they share i18n, state-shape, and test fixtures.

**1. Move the squash suggestion from the PR body into a PR comment.** The suggestion now lives next to the "Squash and merge" dropdown at the bottom of the page rather than a full scroll above it. The agent looks up its own prior squash-suggestion comment by the start marker and PATCHes it in place on every iteration so the timeline stays close to the merge box instead of accumulating duplicates. A new `findLatestCommentWithMarker` helper in `src/pr-comments.ts` is the single source of truth for the lookup, used by both Stage 8's validation path and Stage 9's merge-confirm hint renderer.

**2. Ask the squash-apply policy once per run at startup.** Instead of interrupting the pipeline with an `agent`/`github` chooser every time a SUGGESTED_SINGLE verdict lands, the startup flow now asks once whether to apply it automatically via the agent or ask each time. Default is "let the agent handle it" — a bare Enter picks the low-friction path. The prompt is owned by a small `promptSquashApplyPolicy()` helper in `src/startup.ts`, called once at the shared join point in `src/index.ts` after both the fresh-start and resume branches produce `RunParams`, so both paths go through the same question with no duplication. Not persisted to config or `RunState`: the right answer depends on the state of the current branch / PR, so it is collected fresh every run.

**Other changes bundled in:**
- Sub-step literal renamed `applied_in_pr_body` → `applied_via_github`. `RunState` bumped to v4 with an in-place migration so in-flight runs resume into the same short-circuit branch.
- All three agent-visible prompts (main squash, completion-check, agent-squash follow-up) updated together to describe the comment location consistently — mismatched wording would fall back to the old PR-body location.
- The agent-facing example for locating a prior comment uses `--paginate --slurp` (with a `.[] | .[]` flatten) so the `| last` filter sees the latest match across all pages, matching the production helper in `src/pr-comments.ts`.
- The PR-merged guard now runs before the suggestion-comment lookup in all three Stage 8 paths (immediate `SUGGESTED_SINGLE`, `awaiting_user_choice` resume, and the deterministic fallback after an ambiguous verdict). `findPrNumber` uses open-only `gh pr list`, so without the guard a concurrent merge would either flip the verdict to `BLOCKED` or fall through to a fresh planning run instead of taking the existing `squash.alreadyMerged` short-circuit.
- i18n strings under both `squash.*` and `pipeline.*` updated in `en.ts` and `ko.ts`.
- Doc comments in `src/pipeline.ts` and `src/stage-squash.ts` plus the Stage 8 / Stage 9 sections of `docs/pipeline.md` refreshed to match.
- `stage-squash.test.ts` now fixtures the comment body via `findSuggestionCommentBody` and `findPrNumber` injections in place of the old `getPrBody`.
- A structural test in `src/index.test.ts` guards the join-point invariant: `promptSquashApplyPolicy` is imported from `startup.js`, called exactly once, and the call sits after both the resume branch's params construction and the fresh branch's `params ??=` block.

Closes #274

## Test plan

- [x] `pnpm biome check` passes on all changed files
- [x] `pnpm vitest run` passes (1878 tests including new `squashApplyPolicy`, `findLatestCommentWithMarker`, `promptSquashApplyPolicy`, and join-point structural cases)
- [x] Fresh run: startup prompt appears after issue selection, accepts Enter (auto), reaches Stage 8, verifies no mid-pipeline chooser fires on SUGGESTED_SINGLE
- [x] Fresh run with "ask" policy: mid-pipeline chooser still fires as before
- [x] Resume run: startup prompt also fires on the resume branch (not only fresh start)
- [x] Squash suggestion is posted as a PR comment (not the PR body) and re-running the stage edits the same comment instead of creating a second one
- [x] Stage 9 merge-confirm screen reads the suggestion from the PR comment; `[t]` and `[b]` hotkeys copy title / body
- [x] Load a v3 `RunState` with `squashSubStep === "applied_in_pr_body"` and confirm it migrates to `"applied_via_github"` on load

## Suggested squash commit

**Title**

```text
Post squash suggestion as PR comment, ask apply policy
```

**Body**

```text
Move the single-commit squash suggestion from the PR body to a PR
comment so it sits next to the "Squash and merge" dropdown instead
of a full scroll above it. The agent looks up its own prior comment
by start marker and PATCHes it in place each iteration, so the
timeline stays close to the merge box without accumulating
duplicates.

Ask the squash-apply policy once per run at startup instead of
interrupting the pipeline with an agent/github chooser every time a
SUGGESTED_SINGLE verdict lands. Default is "let the agent handle
it"; a bare Enter picks the low-friction path. Not persisted — the
right answer depends on the state of the current branch / PR, so it
is collected fresh every run, including on the resume path.

Rename the sub-step literal `applied_in_pr_body` to
`applied_via_github` and bump RunState to v4 with an in-place
migration so in-flight runs resume into the same short-circuit
branch.

Closes #274
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,16 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- Stage 8 (Squash) now posts the single-commit suggestion as a PR
+  comment instead of editing the PR body. The agent looks up any
+  prior squash-suggestion comment by its start marker and PATCHes it
+  in place so the timeline stays near the "Squash and merge" dropdown
+  rather than accumulating duplicates.
+- Startup now asks once per run whether a SUGGESTED_SINGLE squash
+  verdict should be applied automatically by the agent or interrupt
+  the pipeline with the per-run chooser. Asked on both the fresh
+  start and resume branches; not persisted to config or RunState.
+  Default is "let the agent handle it".
 - Stage 1 (Bootstrap) is now surfaced retrospectively in the TUI. At first
   render, the status bar shows `Stage 1: Bootstrap → Stage N: <name>` briefly
   (where `N` is 2 on a fresh run or `startFromStage` on resume) before
@@ -27,15 +37,15 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   `Stage:`) once the pull-request number is known, also wrapped with an
   OSC 8 hyperlink pointing at the PR page.
 - Stage 8 (Squash) now offers a single-commit suggestion path: when the agent
-  judges that one commit is appropriate, it writes the suggested title and
-  body into a marker-delimited block in the PR body instead of force-pushing.
-  The user can then either let the agent perform the squash (which reruns CI)
+  judges that one commit is appropriate, it posts the suggested title and
+  body inside a marker-delimited PR comment instead of force-pushing. The
+  user can then either let the agent perform the squash (which reruns CI)
   or apply the suggestion via GitHub's "Squash and merge" at merge time
   (which avoids the extra CI cycle).
 - Stage 9 (Done) merge-confirm screen now prints the suggested squash title,
   body, and PR URL inline when the squash stage finished via the
-  PR-body suggestion path, so the user can copy-paste without opening the
-  browser.
+  squash-suggestion comment path, so the user can copy-paste without opening
+  the browser.
 - Stage 9 (Done) merge-confirm screen now renders `[t] copy` / `[b] copy`
   hotkey hints next to the suggested squash title and body when the terminal
   can write to the system clipboard.  Pressing `t` copies the title and `b`
@@ -53,27 +63,27 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Changed
 
-- Stage 8's squash suggestion block in the PR body now emits the title and
-  body inside separate fenced code blocks (info string `text`) instead of
-  bold-labeled plain Markdown (`**Title:** …` / `**Body:** …`).  GitHub
-  renders a one-click copy icon on fenced blocks and does not reinterpret
-  Markdown characters inside, so the user can paste the suggestion verbatim
-  into the "Squash and merge" dialog.  The agent chooses each fence length
+- Stage 8's squash suggestion comment now emits the title and body inside
+  separate fenced code blocks (info string `text`) instead of bold-labeled
+  plain Markdown (`**Title:** …` / `**Body:** …`).  GitHub renders a
+  one-click copy icon on fenced blocks and does not reinterpret Markdown
+  characters inside, so the user can paste the suggestion verbatim into the
+  "Squash and merge" dialog.  The agent chooses each fence length
   dynamically per the CommonMark rule
   (`max(longest backtick run in content, 2) + 1`, minimum 3) so commit
   bodies containing their own triple-backtick samples survive unchanged.
 - Stage 8 verdict keywords are now `SQUASHED_MULTI` / `SUGGESTED_SINGLE` /
   `BLOCKED` (previously `COMPLETED` / `BLOCKED`).  When the verdict is
   ambiguous after a clarification retry, the handler runs a deterministic
-  fallback chain — commit-count decrease, then PR-body marker presence,
-  then BLOCKED.
+  fallback chain — commit-count decrease, then suggestion-comment marker
+  presence, then BLOCKED.
 - `RunState` now persists a `squashSubStep` field tracking progress through
   the squash stage's substates so resume can re-enter at the correct point.
   `RUN_STATE_VERSION` bumped from 2 to 3 (the new field defaults to
   `undefined` for older state files; no destructive migration).
 - Stage 9 merge-confirm screen now includes a conditional one-line tip
-  (`pipeline.mergeConfirmSquashTip`) when a squash suggestion is live in
-  the PR body.
+  (`pipeline.mergeConfirmSquashTip`) when a squash suggestion comment is
+  live on the PR.
 
 ### Fixed
 
@@ -104,9 +114,9 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   derived verdict.  Previously the event was only emitted when the
   agent response parsed into a concrete keyword, so the fallback
   branch silently skipped telemetry.
-- Stage 8 now validates the squash suggestion block strictly before
-  accepting `SUGGESTED_SINGLE` / `applied_in_pr_body`.  A bare start
-  marker or a block missing the `**Title**` label / the end marker
+- Stage 8 now validates the squash suggestion comment strictly before
+  accepting `SUGGESTED_SINGLE` / `applied_via_github`.  A bare start
+  marker or a comment missing the `**Title**` label / the end marker
   is treated as malformed and fails closed (or re-runs planning on
   resume) instead of completing with `squash.messageAppended` and
   leaving Stage 9 unable to render the inline preview.
@@ -117,7 +127,7 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   earlier.  Without this, a resume where the user picks "agent
   squashes now" could re-send the follow-up on the older planning
   session rather than the exact conversation that drafted the
-  PR-body suggestion.
+  squash-suggestion comment.
 
 ## [0.1.0] - 2026-04-18
 
--- a/docs/pipeline.md
+++ b/docs/pipeline.md
@@ -142,15 +142,16 @@ The standard PR sync instructions appear in stages 5 through 8:
 
 The squash stage extends this contract with a marker-delimited
 **squash suggestion block**.  When a single commit is the right
-shape for the branch, the agent writes the proposed title and body
-into the PR description between the markers
-`<!-- agentcoop:squash-suggestion:start -->` and
+shape for the branch, the agent posts the proposed title and body
+as a PR comment whose body contains the marker block
+`<!-- agentcoop:squash-suggestion:start -->` …
 `<!-- agentcoop:squash-suggestion:end -->` instead of force-pushing.
-The block is idempotent — re-runs replace any prior block between
-the same markers rather than stacking duplicates.  The Stage 9
-merge-confirm screen reads this block back so the user can apply
-the message via GitHub's "Squash and merge" without opening the
-browser.
+The comment is updated idempotently — re-runs PATCH any prior
+squash-suggestion comment rather than appending a new one so the
+timeline stays close to the "Squash and merge" dropdown.  The
+Stage 9 merge-confirm screen reads this block back so the user can
+apply the message via GitHub's "Squash and merge" without opening
+the browser.
 
 ### Issue-implementation reconciliation
 
@@ -1258,8 +1259,8 @@ avoiding re-execution of already-completed work within a round.
 
 **Agent:** A\
 **Purpose:** Decide whether the branch is best presented as a
-single squash commit (in which case the agent writes the suggested
-message into the PR body and lets GitHub's "Squash and merge"
+single squash commit (in which case the agent posts the suggested
+message as a PR comment and lets GitHub's "Squash and merge"
 apply it at merge time) or several meaningful commits (rewrite
 history and force-push).  Skipped automatically if the branch has
 only one commit.
@@ -1276,11 +1277,15 @@ The work prompt asks the agent to:
 2. Decide whether the work belongs in **one** commit or **several**.
 3. Branch on that decision:
    - **Single commit appropriate:** do not rewrite history.  Draft
-     the squash title and body, then write them into the PR body
-     between the markers
-     `<!-- agentcoop:squash-suggestion:start -->` and
-     `<!-- agentcoop:squash-suggestion:end -->` (replacing any prior
-     block between the same markers).  Do not force-push.
+     the squash title and body, then post them as a PR comment whose
+     body contains the marker block
+     `<!-- agentcoop:squash-suggestion:start -->` …
+     `<!-- agentcoop:squash-suggestion:end -->`.  Post idempotently:
+     if a prior comment with the same start marker exists, edit it
+     via `gh api --method PATCH
+     /repos/{owner}/{repo}/issues/comments/{id}` so the timeline
+     does not accumulate duplicates.  The PR body is left untouched.
+     Do not force-push.
    - **Multiple commits appropriate:** consolidate the branch
      commits — using `git reset --soft {baseSha}` + `git commit` or
      interactive rebase — write clear messages, and force-push
@@ -1295,8 +1300,8 @@ of the following keywords:
 - SQUASHED_MULTI — if you rewrote history into multiple meaningful
   commits and force-pushed
 - SUGGESTED_SINGLE — if a single commit is appropriate and you
-  wrote the suggested title/body into the PR body marker block
-  (no force-push)
+  posted the suggested title/body as a PR comment containing the
+  marker block (no force-push)
 - BLOCKED — if you could not complete either path and need user
   intervention
 
@@ -1313,23 +1318,25 @@ because the three-way distinction is squash-specific.
   behaviour, up to 3 internal fix attempts) and finish with
   `squash.completed` when CI passes.
 - `SUGGESTED_SINGLE` → verify a fully parseable suggestion block is
-  present in the PR body before asking the user.  "Parseable" means
-  both markers are present **and** `parseSquashSuggestionBlock`
-  succeeds (the block contains a `**Title**` label followed by a
-  well-formed fenced block).  A bare start marker or a block missing
-  the `**Title**` label / the end marker fails closed with `blocked`
-  rather than completing as if the suggestion were valid — Stage 9
-  reads the same block via
+  present in the squash-suggestion PR comment before asking the
+  user.  "Parseable" means both markers are present **and**
+  `parseSquashSuggestionBlock` succeeds (the block contains a
+  `**Title**` label followed by a well-formed fenced block).  A
+  bare start marker or a block missing the `**Title**` label / the
+  end marker fails closed with `blocked` rather than completing as
+  if the suggestion were valid — Stage 9 reads the same block via
   `parseSquashSuggestionBlock` to render the inline preview, so a
   malformed block would leave the merge-confirm screen blank.  Once
-  validated, ask the user via `chooseSquashApplyMode`:
+  validated, ask the user via `chooseSquashApplyMode` (skipped when
+  the startup `squashApplyPolicy` is `"auto"`, in which case the
+  handler proceeds as if the user picked `"agent"`):
   - **"Agent squashes now"** → send a follow-up prompt on the
     same session telling the agent to perform the squash using
     the message it already drafted, then run the post-squash CI
     poll loop.
   - **"Apply via GitHub"** → finish the stage with
     `squash.messageAppended` and persist
-    `squashSubStep === "applied_in_pr_body"` so Stage 9 surfaces
+    `squashSubStep === "applied_via_github"` so Stage 9 surfaces
     the suggestion in the merge-confirm screen.  No CI rerun.
 - `BLOCKED` → existing blocked flow.
 
@@ -1341,8 +1348,8 @@ fails, the handler runs a **deterministic fallback chain**:
    snapshot taken at stage entry, treat as `SQUASHED_MULTI` (the
    force-push has already happened — that hard-to-undo side
    effect must be detected first).
-2. Else, fetch the PR body and check for a fully parseable
-   suggestion block (markers present **and**
+2. Else, fetch the squash-suggestion PR comment and check for a
+   fully parseable suggestion block (markers present **and**
    `parseSquashSuggestionBlock` returns a value).  If present,
    treat as `SUGGESTED_SINGLE`; a malformed block (e.g. only the
    start marker) is treated as missing.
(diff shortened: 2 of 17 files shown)


sehkone:
**[Reviewer Round 1]**
I found one correctness regression to fix before merging.

- `src/stage-squash.ts:723` now re-enters the `awaiting_user_choice` resume path by re-reading the squash-suggestion comment first, and only then calling `guardIfPrMerged()`. The problem is that `readSuggestionCommentBody()` depends on `findPrNumber()` (`src/stage-squash.ts:623`), and `findPrNumber()` uses `gh pr list` without `--state all` (`src/pr.ts:14`), so it stops finding the PR once that PR has been merged or otherwise closed.
- In that case, resuming a run that was paused at `awaiting_user_choice` after the user has merged on GitHub no longer short-circuits to `squash.alreadyMerged`; it falls through as if the suggestion comment were missing and starts a fresh planning run instead. The same ordering issue exists in the immediate `SUGGESTED_SINGLE` branch at `src/stage-squash.ts:845`, where a concurrent merge can now turn into `BLOCKED` instead of the existing merged short-circuit.
- I think the fix is to check `guardIfPrMerged()` before any comment lookup on those paths, or to make the comment lookup use a saved PR number / a resolver that can still see closed PRs. I’d also add a regression test for “resume from `awaiting_user_choice` after the PR has already been merged”, because the current tests cover `squashing` + merged, but not this new comment-based branch.

sehkone:

sehkone:
**[Author Round 1]**

- **Fixed** — `awaiting_user_choice` resume path (`src/stage-squash.ts:723`) and the immediate `SUGGESTED_SINGLE` branch (`src/stage-squash.ts:845`) now call `guardIfPrMerged()` before `readSuggestionCommentBody()`. You correctly identified that `findPrNumber()` uses `gh pr list` without `--state all`, so once the PR is merged the comment lookup would fall through — bypassing the `squash.alreadyMerged` short-circuit and either starting a fresh planning run or flipping the verdict to `BLOCKED`. Reordering is the minimal fix and preserves the "suggestion comment missing/malformed → fresh planning" fallback for open PRs.
- **Fixed** — Added two regression tests: `awaiting_user_choice with PR already merged → alreadyMerged, no fresh planning, no user choice` asserts the guard runs first (neither `findPrNumber` nor `findSuggestionCommentBody` is consulted on a merged PR), and `SUGGESTED_SINGLE + concurrent merge hides the PR from findPrNumber → alreadyMerged, not blocked` covers the companion ordering issue on the immediate branch using a `findPrNumber` mock that returns `undefined` to simulate the `gh pr list` filtering behaviour.

Verified locally: `pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm vitest run` (1873 tests) all pass.

sehkone:
**[Reviewer Round 2]**

Thanks — the Round 1 ordering fix is in place in the two paths I called out, and the new tests cover those cases. I found a few follow-ups that still look worth addressing before merge:

- `src/stage-squash.ts:817` still has the same merged-PR regression in the deterministic fallback path. When the agent never returns a parseable verdict, the code falls back to `readSuggestionCommentBody()` at `src/stage-squash.ts:822` before checking `guardIfPrMerged()`. That helper still depends on `findPrNumber()` at `src/stage-squash.ts:623`, which shells out to open-only `gh pr list` in `src/pr.ts:14`. So if the PR gets merged between the clarification turn and this fallback classification, the suggestion lookup now disappears and the stage resolves to `BLOCKED` instead of short-circuiting to `squash.alreadyMerged`. This is the same failure mode as Round 1, just in the third code path. I think the fix is the same too: guard for merged before the comment lookup here, and add a regression test for “ambiguous verdict + PR already merged before deterministic fallback”.

- The new prompt example at `src/stage-squash.ts:268` is not actually safe for long PR timelines. `gh api .../comments --paginate --jq '[.[] | select(...)] | last'` applies the jq expression page-by-page unless you also slurp, so on a PR with multiple comment pages it can miss the actual latest matching squash-suggestion comment and nudge the agent into posting a duplicate. That undercuts the main “update in place, don’t clutter the timeline” goal of this issue. The production helper in `src/pr-comments.ts:68`/`src/pr-comments.ts:125` already handles pagination by using `--paginate --slurp`; I’d either mirror that pattern in the prompt example or remove the example command so we don’t hand the agent a brittle lookup.

- I also don’t see coverage for the new shared startup/resume join point in `src/index.ts:440`. The new tests in `src/stage-squash.test.ts:1685` prove Stage 8 respects an injected `squashApplyPolicy`, but they don’t protect the new product requirement that the question is asked once per run on both the fresh-start and resume paths. Since this behavior intentionally lives outside `runStartup()`, I’d like one test around the fresh path and one around the resume path so a future refactor can’t silently stop asking it.

sehkone:

sehkone:

sehkone:
**[Author Round 2]**

Thanks for the careful re-read — all three follow-ups land. Disposition below.

- **Fixed** — Deterministic fallback now guards for a merged PR before the suggestion-comment lookup (`src/stage-squash.ts:817`). Same root cause as Round 1: `findPrNumber` uses open-only `gh pr list`, so a merge between the clarification turn and this fallback turned a missing comment into `BLOCKED` instead of taking the `squash.alreadyMerged` short-circuit. Reordering inside the `else` branch keeps `SQUASHED_MULTI` detection (commit-count drop) intact while making the merged short-circuit win over the comment lookup. Regression test added: `ambiguous verdict + PR merged before fallback → alreadyMerged, not BLOCKED` asserts the lifecycle guard runs first and neither `findPrNumber` nor `findSuggestionCommentBody` is consulted on a merged PR.

- **Fixed** — Prompt example now uses `--paginate --slurp` with a `.[] | .[]` flatten (`src/stage-squash.ts:268`), matching the production helper in `src/pr-comments.ts:68`. You correctly identified that `--paginate` alone applies the `--jq` filter per page, so `| last` would return the last match within each page rather than the latest match overall — defeating the "update in place" goal. Kept the example rather than removing it because the marker-block instructions a few lines above are abstract; a concrete command (and an explanation of *why* `--slurp` is required) is more likely to keep the agent on the safe path than dropping the example entirely.

- **Fixed** — Extracted the join-point prompt into `promptSquashApplyPolicy()` in `src/startup.ts` so the call site is a single named target. `src/index.ts:440` now reads `params.squashApplyPolicy = await promptSquashApplyPolicy()` instead of an inline `confirm()`. Two layers of coverage protect the per-run-on-both-paths requirement: (1) unit tests for the helper itself in `startup.test.ts` (default Enter → `auto`, explicit no → `ask`, default flag asserted), and (2) a structural test suite in `index.test.ts` that reads `src/index.ts` and asserts the helper is imported from `startup.js`, called exactly once, sits after BOTH the resume branch (`resuming: true,`) and the fresh branch's `params ??= await` block, and assigns to `params.squashApplyPolicy`. A refactor that silently drops the call on either path will fail the structural suite.

CI status: `pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm vitest run` (1878 tests) all pass locally.

sehkone:
**[Reviewer Round 3]**

Round 2 follow-ups look resolved to me: the merged-PR guard now covers the deterministic fallback, the prompt example uses `--paginate --slurp`, and the startup policy prompt has dedicated coverage at the shared join point.

I found one remaining doc/comment sweep gap:

- `src/stage-squash.ts:811` still says the resume path needs to continue "the exact conversation that drafted the PR-body suggestion", and `docs/pipeline.md:1426` repeats the same "PR-body suggestion" wording. After this change the source of truth is the squash-suggestion PR comment, so these are now stale in exactly the place the issue called out as a trap for future readers.

sehkone:

sehkone:
**[Author Round 3]**

- **Fixed** — `src/stage-squash.ts:811` and `docs/pipeline.md:1426` now say "drafted the squash-suggestion comment" instead of "drafted the PR-body suggestion". The two phrasings sit in matching spots (the same paragraph about persisting the verdict turn's session id before `awaiting_user_choice`), so they stay in sync for future readers. I also swept the rest of the tree and updated two more stale matches while I was there: `src/stage-squash.test.ts:1415` (the comment on the "verdict session id persistence" describe block, which mirrors the prod paragraph) and `CHANGELOG.md:47` / `CHANGELOG.md:130` (the unreleased-section bullets that described the Stage 9 hint and the session-id-persistence fix in terms of the old PR-body suggestion path). `pnpm tsc --noEmit` and `pnpm biome check` pass on the touched files.

sehkone:
**[Reviewer Round 4]**

The Round 3 item is fixed: `src/stage-squash.ts:808` and `docs/pipeline.md:1426` now both refer to the squash-suggestion comment.

I found one remaining documentation sweep gap:

- `CHANGELOG.md:39`, `CHANGELOG.md:66`, `CHANGELOG.md:78`, `CHANGELOG.md:84`, and `CHANGELOG.md:117` still describe the current unreleased squash flow in terms of the PR body / `applied_in_pr_body`. Because these bullets are under `[Unreleased]`, they now document the current behavior and are stale after this PR. They should be updated to match the shipped behavior: suggestion lives in a PR comment, the deterministic fallback checks suggestion-comment presence, the Stage 9 tip appears when a suggestion comment exists, and the sub-step is `applied_via_github`.

sehkone:

sehkone:
**[Author Round 4]**

- **CHANGELOG.md:39 / :66 / :78 / :84 / :117 — Fixed.** All five `[Unreleased]` bullets now describe shipped behavior:
  - Line 39 (Added): "writes … into a marker-delimited block in the PR body" → "posts … inside a marker-delimited PR comment".
  - Line 66 (Changed): "Stage 8's squash suggestion block in the PR body" → "Stage 8's squash suggestion comment".
  - Line 78 (Changed): fallback chain now reads "commit-count decrease, then suggestion-comment marker presence, then BLOCKED" (was "PR-body marker presence").
  - Line 84 (Changed): tip condition reworded from "when a squash suggestion is live in the PR body" to "when a squash suggestion comment is live on the PR".
  - Line 117 (Fixed): validation entry updated from `applied_in_pr_body` → `applied_via_github` and "block" → "comment".

I left the existing v2→v3 RunState bullet (lines 80–83) and the v2→v3 migration-guard bullet (lines 96–100) alone — they are accurate descriptions of the v3 work that is also part of this `[Unreleased]` band. The v3→v4 sub-step rename added by this PR is implicitly covered by the new top-of-section "Stage 8 … posts the single-commit suggestion as a PR comment …" entry plus the renamed-substep mention in the line-117 Fixed bullet, and the PR description spells out the v4 bump explicitly. Happy to add a separate explicit "RUN_STATE_VERSION bumped from 3 to 4" line if you'd prefer the changelog to call that out independently.

sehkone:
**[Reviewer Round 5]**

The Round 4 item is fixed: `CHANGELOG.md:11`, `CHANGELOG.md:39`, `CHANGELOG.md:78`, `CHANGELOG.md:84`, and `CHANGELOG.md:117` now describe the comment-based squash flow / `applied_via_github` state accurately.

One small new nit:
- `src/stage-squash.test.ts:1513` still says the deterministic fallback derives the verdict from “commit count and PR body”, but this branch now reads the squash suggestion from PR comments. Please update that comment so the test file stays in sync with the implementation and the rest of the documentation sweep.

sehkone:

sehkone:
**[Author Round 5]**

- **Fixed** — `src/stage-squash.test.ts:1513`: updated the comment block above the `pipeline:verdict emission from deterministic fallback` describe to say the fallback derives the verdict from "commit count and the squash-suggestion PR comment" instead of "commit count and PR body". Pushed in 5ed0841.

sehkone:
**[Reviewer Round 6]**

Thanks — the Round 5 nit is fixed. `src/stage-squash.test.ts:1513` now describes the deterministic fallback correctly in terms of the squash-suggestion PR comment.

I re-reviewed the updated diff against #274, including the comment-based squash-suggestion flow, the `applied_via_github` state migration, the Stage 9 hint path, and the startup/resume wiring for the per-run squash-apply policy. I did not find any still-unresolved items to carry forward from Round 5, and I did not spot any new issues in this revision.

sehkone:

PR #278, opened 2026-04-21
Track, display, and check CLI versions (#276)

### LINKED ISSUE #276: Track, display, and check CLI versions for claude and codex
## Summary

AgentCoop currently assumes `claude` and `codex` CLIs are installed and in `$PATH`, but does not know or surface which versions are actually being executed, nor does it help users stay current. This issue proposes recording the versions AgentCoop most recently ran with, displaying them in the agent panes, and prompting the user at startup when a newer version is available.

## Motivation

- The user has no feedback on which CLI version AgentCoop is driving — regressions or feature gaps caused by a CLI upgrade/downgrade are hard to diagnose.
- CLIs are updated frequently; users miss updates unless they remember to run the package manager themselves.
- Recording the version actually used for each run makes postmortem and issue reproduction reliable.

### 1. Record the executed versions

- Before spawning an agent, run `claude --version` / `codex --version` and capture the version string.
- Persist the most recent observed version per CLI in `~/.agentcoop/config.json` as `lastKnownVersions: { claude, codex }` (used for the startup comparison only).
- **Also record the version used for each individual run** in `run-state` (`~/.agentcoop/runs/…json`) and/or `run-log` (`~/.agentcoop/logs/…jsonl`). The config value gets overwritten across runs; the per-run record is what survives for postmortem and reproduction.

### 2. Show the version in each pane

- Agent pane header shows the CLI name with its version, e.g. `Agent A (Claude v1.2.3)` / `Agent B (Codex v0.4.1)`.
- Display the version detected at pipeline start so it does not flicker mid-run.
- Injection path: detected versions are passed into `renderApp` alongside the existing `modelNameA`/`cliTypeA` static metadata (see `index.ts` around the `renderApp` call) and composed into the pane header in `AgentPane.tsx` next to `label`/`modelName`. No adapter changes required.

### 3. Startup update check

**When to run the check.** Only after the CLIs actually used in this run are known — i.e. after the fresh/resume branches join and `params` (with `agentAConfig.cli` / `agentBConfig.cli`) is finalized in `index.ts`. `selectTarget()` internally calls `loadConfig()` and `runStartup()` is where agent CLI choice is determined, so running the check any earlier would prompt for a CLI that is not used this run. Only check the CLIs actually referenced by `agentAConfig` and `agentBConfig` (deduplicated).

**How to find the latest version.** Source depends on the CLI, because Codex in particular has multiple distribution channels that update independently:

- **Claude** (`@anthropic-ai/claude-code`): npm is the effective single channel. Query `https://registry.npmjs.org/@anthropic-ai/claude-code/latest` and read `version`.
- **Codex**: distributed via **npm** (`@openai/codex`), **Homebrew formula**, **Homebrew cask**, and **standalone GitHub release binaries**, each updating on its own cadence (observed gap: formula `0.46.0` vs cask `0.122.0` at time of writing). Resolve the binary with `readlink -f` (follow symlinks) and match the real path against known layouts:
    - Under an npm global prefix (e.g. `…/lib/node_modules/@openai/codex/…`) → npm registry.
    - Under `/opt/homebrew/Cellar/…` or `/usr/local/Cellar/…` → **Homebrew formula** metadata.
    - Under `/opt/homebrew/Caskroom/…`, `/usr/local/Caskroom/…`, or `~/Applications/…` linked from a cask → **Homebrew cask** metadata.
    - Under a recognized standalone layout such as `~/.codex/bin/…` → GitHub releases for `openai/codex`.
    - Anything else (unrecognized hand-placed binary, custom install) → **inconclusive**: skip the update check for this CLI and log why. Do not guess a channel — a wrong prompt is worse than no prompt.

**Prompt flow.**

- Compare the installed version (from `--version`) with the channel-appropriate latest.
- If the installed version is older, prompt: `A newer version of Claude is available (v1.2.3 → v1.3.0). Update now?`
- If the user answers yes, AgentCoop pauses and shows: `Please update Claude, then press Enter to continue.`
- When the user confirms, AgentCoop re-runs `--version` and verifies the new version.
    - If the version matches or exceeds the latest, proceed.
    - **If the version is unchanged** (update failed, shell didn't rehash, wrong package manager, user pressed Enter by mistake), show `Version is still v1.2.3.` and offer three choices: **retry** (re-run `--version` after another pause), **skip** (proceed with the current version for this run), **abort** (exit AgentCoop). No silent fallback.
- If the user answers no to the original prompt, continue with the current version and do not re-prompt in the same session.

## Why npm registry for Claude rather than GitHub releases

- Claude is distributed primarily via npm, so the registry is the source of truth for what users actually install.
- No authentication and no practical rate limits for this use case.
- Simpler response shape: one JSON fetch, read `version`. GitHub releases would require handling drafts, prereleases, and tag-naming inconsistencies.
- An npm publish is the release; a GitHub release can lag or be skipped.

(For Codex, a single source is explicitly not appropriate — see the install-source detection above.)

## Open questions

- Should the update check be skippable via config (e.g. `skipVersionCheck: true`) or a CLI flag for CI-like environments? Suggested: yes, plus a `lastVersionCheckAt` timestamp to throttle checks (e.g. once per day).
- Should we treat network failure to reach the registry/releases endpoint as fatal, or silently continue with a log line? Suggested: silent, log-only.
- Are there other CLIs we expect to add soon (e.g. Gemini)? If so, the version-check module should be structured per-CLI, with each CLI declaring its own channel-detection and latest-version resolver.

### Code

- New module: `src/version-check.ts` — spawns `--version`, detects install source (for Codex), fetches the appropriate latest-version source, compares. Structured so each CLI plugs in its own resolver.
- `src/config.ts`: extend `Config` with `lastKnownVersions` and optional skip/throttle fields.
- `src/run-state.ts` (and/or `src/bootstrap-log.ts` / run-log): record the CLI versions actually used for each run, for postmortem.
- `src/index.ts`: insert the version check after the fresh/resume branches join and `params` is finalized (i.e. after the agent CLI choice is known), before `renderApp`. Pass detected versions into `renderApp` alongside the existing `modelNameA`/`cliTypeA` metadata.
- `src/ui/AgentPane.tsx`: extend the header composition to include the version next to the existing `label`/`modelName`. No adapter changes required.

### Documentation

- `CHANGELOG.md`: add an entry under `## [Unreleased]` → `### Added` describing the three user-visible changes (version in pane header, startup update check with per-channel resolution for Codex, per-run version recording). Follow the existing paragraph-style bullet convention.
- `README.md`:
    - Update the Terminal UI ASCII diagram (the "Agent A (author) — Claude Opus 4.6 ● [*]" header line) to show the new `Claude v1.2.3` suffix so the illustration matches the implementation.
    - In the `Configuration` table, add rows for any new fields introduced (`lastKnownVersions`, and whichever of `skipVersionCheck` / `lastVersionCheckAt` are accepted from Open Questions).
    - In `Prerequisites` or a short new subsection, document the startup update-check behavior and how to skip it (config field and/or env var), so users running in CI or offline know what to expect.

sehkone:
## Summary

Adds visibility and update prompts for the `claude` / `codex` CLIs that AgentCoop drives:

- **Pane header** now shows the CLI name and installed CLI version next to the model name (e.g. `Agent A (author) — Claude Opus 4.7 v1.2.3`). The CLI name disambiguates panes when model names are shared or user-configured. When `modelDisplayName` already starts with the CLI name (e.g. `models.json` entries like `Claude Opus 4.7`), `formatPaneHeader` collapses the duplicate so the header reads `Claude Opus 4.7` rather than `Claude Claude Opus 4.7`. The version is detected once at pipeline start so it does not flicker mid-run, and is threaded through `renderApp` so no adapter changes are required.
- **Per-run recording**: the detected version is saved into `RunState` (`agentA.cliVersion` / `agentB.cliVersion`) and into the run-log header under a `version` line. It is refreshed at resume time so a CLI upgrade between runs is captured, not silently discarded. A failed re-probe on resume (e.g. `--version` crashed or its output could not be parsed) intentionally keeps the previously recorded version rather than blanking it, preserving the postmortem record — and the run-log header reads from `runState` (via `runState.agentA.cliVersion` / `runState.agentB.cliVersion`) rather than the raw probe result, so a failed re-probe preserves the postmortem record in the log path too, not just `RunState`.
- **Startup update check** runs after the fresh/resume branches join and `params` is finalized, so it never prompts for a CLI that is not used this run. It deduplicates the two agents' CLIs before running.
- **Channel-appropriate latest resolution**:
  - Claude → npm registry (`@anthropic-ai/claude-code`).
  - Codex → resolved from the binary's realpath against the npm global prefix (anchored on `lib/node_modules/@openai/codex/` so a project-local install at `<repo>/node_modules/@openai/codex/...` — reachable via `node_modules/.bin/codex` — falls through to `inconclusive` rather than getting a global update prompt), the pnpm global prefix (anchored on `/pnpm/global/` plus `/node_modules/@openai/codex/` so a real `pnpm add -g @openai/codex` install under e.g. `~/Library/pnpm/global/<N>/node_modules/@openai/codex/…`, `~/.local/share/pnpm/global/<N>/node_modules/…`, or `$PNPM_HOME/global/<N>/node_modules/…` is routed to the npm registry without reopening the project-local false positive — project-local pnpm uses `<repo>/node_modules/.pnpm/…` with no `global` segment), Homebrew formula anchored on the `codex` package segment and root-anchored on the Homebrew-default prefixes (`/opt/homebrew/Cellar/codex/`, `/usr/local/Cellar/codex/`, `/home/linuxbrew/.linuxbrew/Cellar/codex/`), Homebrew cask anchored likewise (`/opt/homebrew/Caskroom/codex/`, `/usr/local/Caskroom/codex/`, `/home/linuxbrew/.linuxbrew/Caskroom/codex/`), or a standalone install rooted at `~/.codex/` — specifically `~/.codex/bin/`, `~/.codex/versions/<ver>/`, or the current `~/.codex/packages/standalone/releases/` layout. Homebrew matches run before the npm marker check, so a formula install that bundles `libexec/lib/node_modules/@openai/codex/…` under `Cellar/codex/` resolves to the formula channel rather than npm. Homebrew anchoring also prevents a custom wrapper formula/cask (e.g. `Cellar/my-codex-wrapper/…`, `Caskroom/custom-codex/…`) **and** a copied tree under a non-Homebrew root (e.g. `/tmp/opt/homebrew/Cellar/codex/…`, `/Users/me/sandbox/usr/local/Caskroom/codex/…`) from being misclassified as the official `codex` package. The three Homebrew prefixes are Homebrew's documented defaults: Apple Silicon macOS (`/opt/homebrew`), Intel macOS (`/usr/local`), and Linuxbrew (`/home/linuxbrew/.linuxbrew`). Only these known subpaths under the running user's home directory count as a standalone install. Any unrecognized subpath inside `~/.codex/` (e.g. `~/.codex/tools/codex`, `~/.codex/tmp/codex`) and any `.codex` directory outside `\$HOME` (e.g. `/tmp/.codex/…`) fall through to `inconclusive` so the check is skipped rather than wrong. Likewise, any raw `Applications/` path that did not resolve through `realpath` into a Caskroom directory — system-wide (`/Applications/…`), home-rooted (`~/Applications/…`), or rogue (`/tmp/Applications/…`) — is treated as inconclusive rather than guessed as a cask.
- **Prompt flow**: if a newer version exists, AgentCoop prompts to update, pauses for the user to run their package manager, and re-runs `--version` on return. If the version is unchanged it offers retry / skip (proceed with the current version for this run) / abort — no silent fallback.
- **Config**: new `lastKnownVersions`, `skipVersionCheck`, and `lastVersionCheckAt` fields. `skipVersionCheck: true` disables the check entirely; the 24h throttle avoids hitting the registry on every run. Throttled/skipped runs still probe the installed version for the pane header via a lightweight `--version` parse — they do NOT call into the registry / Homebrew / GitHub. `lastVersionCheckAt` only advances after at least one CLI reached a real installed-vs-latest comparison — a run where every channel was inconclusive, or where the registry fetch failed, re-checks on the next start rather than being silently throttled for 24h. Network failures on non-throttled runs are logged but never fatal. Persistence of these fields uses a narrow JSON patch (`patchVersionCheckState` in `src/config.ts`) rather than a full `saveConfig` round-trip, so unknown top-level keys in `~/.agentcoop/config.json` (user-added fields, fields from a newer AgentCoop version) are preserved — mirroring the "don't rewrite unless dirty" guard in `runStartup()`. The patcher also merges `lastKnownVersions` against the existing nested object instead of replacing it, so unknown nested CLI entries (e.g. a `gemini` field added by a newer AgentCoop version) survive a claude/codex-only check.
- **Docs**: `CHANGELOG.md`, README "Terminal UI" ASCII diagram, a new "CLI update check" subsection under Prerequisites, and the configuration table are updated to reflect the new fields and behavior.

Closes #276

## Test plan

- [x] \`pnpm check\` (Biome) passes
- [x] \`pnpm tsc --noEmit\` passes
- [x] \`pnpm vitest run\` — all tests pass, including new coverage for the header-duplication collapse, for failure/inconclusive runs not advancing \`lastVersionCheckAt\`, and for a mixed run (one CLI successful, one inconclusive) advancing it
- [x] Pane header renders \`Agent A (author) — Claude Opus 4.7 v<installed>\` and \`Agent B (reviewer) — Codex GPT-5.4 v<installed>\` when versions are detected, without the "Claude Claude ..." duplication that the models.json display name + cliName prefix would otherwise produce
- [x] Pane header falls back to \`Agent A (author) — Claude Opus 4.7\` when version is undefined (no regression)
- [x] Fresh run where at least one CLI reached installed-vs-latest comparison: version-check writes \`lastKnownVersions\` and \`lastVersionCheckAt\`, and the recorded value lands in \`~/.agentcoop/runs/*.json\` and the run-log header
- [x] Fresh run where **every** CLI was inconclusive (custom layout) or the fetch failed: \`lastVersionCheckAt\` is NOT advanced, so the next run re-checks rather than being silently throttled for 24h
- [x] Resume run: version is re-detected and the saved \`cliVersion\` in \`RunState\` is refreshed to the current CLI build
- [x] Resume run with failed re-probe: saved \`cliVersion\` is preserved (not blanked) in \`RunState\` **and** in the new run-log header (both read from \`runState.agent*.cliVersion\`)
- [x] \`skipVersionCheck: true\` in \`~/.agentcoop/config.json\` suppresses the check and all network calls (verified by vi.fn spies on fetchJson / resolveBinary)
- [x] \`lastVersionCheckAt\` within the last 24h throttles the check (no network call, no install-source detection)
- [x] When a newer Claude version is published to npm, the update prompt fires with \`v<from> → v<to>\`
- [x] Codex install on npm global prefix (\`<prefix>/lib/node_modules/@openai/codex/...\`) resolves latest from the npm registry
- [x] Codex install on pnpm global prefix (\`~/Library/pnpm/global/<N>/node_modules/@openai/codex/…\` on macOS, \`~/.local/share/pnpm/global/<N>/node_modules/…\` on Linux, \`\$PNPM_HOME/global/<N>/node_modules/…\` when set, and the realpath-resolved \`.pnpm/@openai+codex@<ver>/node_modules/@openai/codex/…\` shape) resolves latest from the npm registry
- [x] Project-local pnpm layout (\`<repo>/node_modules/.pnpm/@openai+codex@<ver>/node_modules/@openai/codex/...\`) is reported as inconclusive — not guessed as a global pnpm install
- [x] Codex install resolved to a project-local \`node_modules/@openai/codex/...\` path (no \`lib/\` segment) is reported as inconclusive — not guessed as a global npm install
- [x] Codex install on Homebrew formula (\`/opt/homebrew/Cellar/codex/\`, \`/usr/local/Cellar/codex/\`, \`/home/linuxbrew/.linuxbrew/Cellar/codex/\`) resolves latest from formula metadata — including a formula that bundles \`node_modules/@openai/codex/...\` internally
- [x] Custom Homebrew formula (e.g. \`Cellar/my-codex-wrapper/…\`) is reported as inconclusive — not guessed as the official \`codex\` formula
- [x] Homebrew formula lookalike under a non-Homebrew root (e.g. \`/tmp/opt/homebrew/Cellar/codex/…\`, \`/Users/me/sandbox/usr/local/Cellar/codex/…\`, \`/var/tmp/home/linuxbrew/.linuxbrew/Cellar/codex/…\`) is reported as inconclusive — not guessed as an actual Homebrew install
- [x] Codex install on Homebrew cask (\`/opt/homebrew/Caskroom/codex/\`, \`/usr/local/Caskroom/codex/\`, \`/home/linuxbrew/.linuxbrew/Caskroom/codex/\`) resolves latest from cask metadata
- [x] Custom Homebrew cask (e.g. \`Caskroom/custom-codex/…\`) is reported as inconclusive — not guessed as the official \`codex\` cask
- [x] Homebrew cask lookalike under a non-Homebrew root (e.g. \`/tmp/opt/homebrew/Caskroom/codex/…\`, \`/Users/me/sandbox/usr/local/Caskroom/codex/…\`, \`/var/tmp/home/linuxbrew/.linuxbrew/Caskroom/codex/…\`) is reported as inconclusive — not guessed as an actual Homebrew install
- [x] Codex install under the known \`~/.codex/\` layouts (\`bin\`, \`versions/<ver>\`, current \`packages/standalone/releases/...\`) resolves latest from GitHub releases
- [x] Unrecognized subpath inside \`~/.codex/\` (e.g. \`~/.codex/tools/codex\`, \`~/.codex/tmp/codex\`, \`~/.codex/codex\`) is reported as inconclusive — not guessed as a GitHub-release install
- [x] \`.codex\` directory outside \$HOME (e.g. \`/tmp/.codex/codex\`) is reported as inconclusive and the check is skipped — not guessed as a GitHub-release install
- [x] Any \`Applications/\` path that did not resolve into Caskroom (\`/Applications/Codex.app/…\`, \`~/Applications/Codex.app/…\`, \`/tmp/Applications/codex\`) is reported as inconclusive — not guessed as a cask install
- [x] Unrecognized Codex install layout is reported as inconclusive and the check is skipped for that CLI
- [x] \"Update now?\" → yes, version unchanged on return → retry / skip / abort options appear and each behaves as documented
- [x] Network failure to registry/releases is logged but does not abort startup
- [x] Updating version-check state preserves unknown top-level keys in \`~/.agentcoop/config.json\` (covered by a new \`patchVersionCheckState\` test that seeds the file with unknown fields and asserts they survive)
- [x] Updating \`lastKnownVersions\` preserves unknown nested CLI entries (e.g. a forward-compatible \`gemini\` field) — covered by a new \`patchVersionCheckState\` test that seeds \`lastKnownVersions.gemini\` and asserts it survives a claude/codex update

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,73 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- Agent pane headers now include the CLI name and installed CLI
+  version next to the model label (e.g.
+  `Agent A (author) — Claude Opus 4.6 v1.2.3`). The CLI name
+  disambiguates panes when model names are shared or user-configured.
+  The version is detected once at pipeline start so it does not flicker
+  mid-run, and is threaded through `renderApp` so no adapter changes are
+  required.
+- Startup now compares the installed `claude` / `codex` CLI against the
+  channel-appropriate "latest" version and prompts the user to update
+  when a newer release is available. The check runs after the CLIs
+  actually used this run are known (fresh and resume branches joined,
+  `params` finalized) so it never prompts for a CLI that is not in use.
+  Claude's latest is fetched from the npm registry. Codex's latest is
+  resolved by matching the installed binary's realpath against known
+  layouts (npm global prefix anchored on
+  `lib/node_modules/@openai/codex/` so a project-local install at
+  `<repo>/node_modules/@openai/codex/...` — reachable via
+  `node_modules/.bin/codex` — is treated as inconclusive rather than
+  pointed at the npm registry, Homebrew formula `Cellar/codex/`,
+  Homebrew cask `Caskroom/codex/`, or a standalone install rooted at
+  `~/.codex/` — specifically `~/.codex/bin/`,
+  `~/.codex/versions/<ver>/`, or the current
+  `~/.codex/packages/standalone/releases/` layout). Homebrew matches
+  are anchored on the `codex` package segment and root-anchored on the
+  Homebrew prefixes (`/opt/homebrew/`, `/usr/local/`,
+  `/home/linuxbrew/.linuxbrew/`),
+  so a custom wrapper formula/cask (e.g. `Cellar/my-codex-wrapper/...`,
+  `Caskroom/custom-codex/...`) and a copied tree under a non-Homebrew
+  root (e.g. `/tmp/opt/homebrew/Cellar/codex/...`,
+  `/Users/me/sandbox/usr/local/Caskroom/codex/...`) are not
+  misclassified as the official `codex` package. Only the known
+  subpaths under the running user's home directory are accepted as a
+  standalone install; any other
+  layout — including an unrecognized subpath inside `~/.codex/` (e.g.
+  `~/.codex/tools/codex`) or a raw `Applications/` path that did not
+  resolve through `realpath` into a Caskroom directory — is treated as
+  inconclusive and the check is skipped for that CLI rather than
+  guessed. When an update is
+  chosen, AgentCoop pauses for the user to run their package manager
+  and re-runs `--version` on return; if the
+  version is unchanged it offers retry / skip (proceed with the current
+  version) / abort. The check can be disabled with
+  `skipVersionCheck: true` in `~/.agentcoop/config.json` and is
+  throttled to once per 24h via `lastVersionCheckAt`. The throttle
+  timestamp only advances after at least one CLI reached a real
+  installed-vs-latest comparison, so a run where every channel was
+  inconclusive or the registry fetch failed still re-checks on the
+  next start rather than being silently throttled for 24h. Network
+  failures are logged but never fatal. The persistence writes
+  `lastKnownVersions` / `lastVersionCheckAt` as a narrow patch into
+  the raw JSON rather than round-tripping through the normalized
+  `Config` shape, so unknown top-level keys in
+  `~/.agentcoop/config.json` (user-added fields, fields from a newer
+  AgentCoop version, etc.) are preserved across the update. The patch
+  also merges `lastKnownVersions` against the existing nested object
+  rather than replacing it, so unknown nested CLI entries (e.g. a
+  forward-compatible `gemini` field added by a newer AgentCoop
+  version) survive a claude/codex-only check.
+- Per-run CLI versions are now recorded into `RunState`
+  (`agentA.cliVersion` / `agentB.cliVersion`) and into the run-log
+  header under a `version` line for each agent, so postmortem
+  reproduction can identify the exact CLI build that produced a run.
+  The values survive across resume and are refreshed at resume time so
+  a CLI upgrade between runs is captured, not silently discarded. A
+  failed re-probe on resume (e.g. `--version` crashed or its output
+  could not be parsed) intentionally keeps the previously recorded
+  version rather than blanking it, preserving the postmortem record.
 - Stage 8 (Squash) now posts the single-commit suggestion as a PR
   comment instead of editing the PR body. The agent looks up any
   prior squash-suggestion comment by its start marker and PATCHes it
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ minimal human involvement.
 ## Terminal UI
 
 ```text
-┌─ Agent A (author) — Claude Opus 4.6 ● [*] ─┬─ Agent B (reviewer) — GPT-5.4 ─┐
+┌─ Agent A (author) — Claude Opus 4.6 v1.2.3 ● [*] ─┬─ Agent B (reviewer) — Codex GPT-5.4 v0.46.0 ─┐
 │                                            │                                │
 │  (streamed output)                         │  (streamed output)             │
 │                                            │                                │
@@ -43,7 +43,9 @@ minimal human involvement.
 ```
 
 - **Agent panes** — Streamed output from each agent in real time.
-  The active agent is marked with `●`, the focused pane (for
+  The pane header shows the agent role, the CLI name (Claude / Codex),
+  the model name, and the installed CLI version (detected at pipeline
+  start). The active agent is marked with `●`, the focused pane (for
   scrolling) with `[*]`.
 - **TokenBar** — Per-agent token usage (input/output, with cached
   token counts when available).
@@ -243,6 +245,50 @@ to reuse it, clean up and recreate, or halt.
 - Any external services the agents use (e.g., Figma Desktop MCP)
   must be running and authenticated
 
+### CLI update check
+
+At startup (after the agents for this run are known) AgentCoop runs
+`claude --version` / `codex --version` and compares the installed
+version against the latest release from the channel-appropriate
+source. Claude always uses the npm registry
+(`@anthropic-ai/claude-code`). Codex's channel is resolved from the
+installed binary's realpath: npm global prefix, Homebrew formula
+(`Cellar/codex/`), Homebrew cask (`Caskroom/codex/`), or a standalone
+install rooted at `~/.codex/` — specifically `~/.codex/bin/`,
+`~/.codex/versions/<ver>/`, or the current
+`~/.codex/packages/standalone/releases/` layout. Homebrew matches are
+anchored on the `codex` package segment and root-anchored on the
+Homebrew prefixes (`/opt/homebrew/`, `/usr/local/`,
+`/home/linuxbrew/.linuxbrew/`), so
+a custom wrapper formula or cask (e.g. `Cellar/my-codex-wrapper/…`,
+`Caskroom/custom-codex/…`) and a copied tree under a non-Homebrew
+root (e.g. `/tmp/opt/homebrew/Cellar/codex/…`,
+`/Users/me/sandbox/usr/local/Caskroom/codex/…`) are not misclassified
+as the official `codex` package. Only the known
+subpaths under the running user's home directory are accepted as a
+standalone install; an unrecognized subpath inside `~/.codex/` (e.g.
+`~/.codex/tools/codex`) or any `.codex` directory outside `$HOME`
+(e.g. `/tmp/.codex/…`) is treated as inconclusive and the check is
+skipped rather than guessed.
+Likewise, a raw `Applications/` path that did not resolve through
+`realpath` into a Caskroom directory is treated as inconclusive rather
+than guessed as a cask, so a hand-placed `~/Applications/Codex.app/…`
+or `/tmp/Applications/codex` never produces a bogus update prompt.
+
+If the installed version is older, AgentCoop prompts to update and
+pauses until you press Enter so you can run your package manager in
+another tab. On return it re-runs `--version` and either proceeds, or
+— if the version is unchanged — offers retry / skip / abort.
+
+Skip the check in CI or offline environments by setting
+`skipVersionCheck: true` in `~/.agentcoop/config.json`. The check is
+also throttled to once per 24h via the `lastVersionCheckAt`
+timestamp. That timestamp only advances after at least one CLI
+reached a real installed-vs-latest comparison — a run where every
+channel was inconclusive, or where the registry fetch failed, will
+re-check on the next start rather than being silently throttled
+for 24h. Network failures are logged but never fatal.
+
 ## Installation & quick start
 
 ```bash
@@ -283,6 +329,10 @@ Settings are stored in `~/.agentcoop/config.json`:
 | `notifications.desktop` | Desktop notification | `false` |
 | `customModels.claude` | Extra Claude model entries | _(none)_ |
 | `customModels.codex` | Extra Codex/GPT model entries | _(none)_ |
+| `lastKnownVersions.claude` | Most recent `claude --version` output | _(first run)_ |
+| `lastKnownVersions.codex` | Most recent `codex --version` output | _(first run)_ |
+| `skipVersionCheck` | Skip the startup update check | `false` |
+| `lastVersionCheckAt` | Epoch ms of the last check (throttle) | _(first run)_ |
 
 Agent presets (CLI, model, context window, effort level) are also
 saved per agent slot.
--- a/src/config.test.ts
+++ b/src/config.test.ts
@@ -14,7 +14,8 @@ vi.mock("node:os", () => ({
   homedir: () => tmpHome,
 }));
 
-const { configPath, loadConfig, saveConfig } = await import("./config.js");
+const { configPath, loadConfig, patchVersionCheckState, saveConfig } =
+  await import("./config.js");
 
 describe("configPath", () => {
   test("returns path under ~/.agentcoop/", () => {
@@ -135,6 +136,41 @@ describe("loadConfig", () => {
     writeFileSync(configPath(), JSON.stringify({ owners: "not-an-array" }));
     const config = loadConfig();
     expect(config.owners).toEqual([]);
+  });
+
+  test("reads version-check fields when present", () => {
+    writeFileSync(
+      configPath(),
+      JSON.stringify({
+        owners: ["aicers"],
+        lastKnownVersions: { claude: "1.2.3", codex: "0.46.0" },
+        skipVersionCheck: true,
+        lastVersionCheckAt: 1_700_000_000_000,
+      }),
+    );
+    const config = loadConfig();
+    expect(config.lastKnownVersions).toEqual({
+      claude: "1.2.3",
+      codex: "0.46.0",
+    });
+    expect(config.skipVersionCheck).toBe(true);
+    expect(config.lastVersionCheckAt).toBe(1_700_000_000_000);
+  });
+
+  test("ignores malformed version-check fields", () => {
+    writeFileSync(
+      configPath(),
+      JSON.stringify({
+        owners: ["aicers"],
+        lastKnownVersions: "not-an-object",
+        skipVersionCheck: "yes",
+        lastVersionCheckAt: "not-a-number",
+      }),
+    );
+    const config = loadConfig();
+    expect(config.lastKnownVersions).toBeUndefined();
+    expect(config.skipVersionCheck).toBeUndefined();
+    expect(config.lastVersionCheckAt).toBeUndefined();
   });
 
   test("filters out non-string elements from owners array", () => {
@@ -762,5 +798,105 @@ describe("saveConfig", () => {
     saveConfig(original);
     const loaded = loadConfig();
     expect(loaded).toEqual(original);
+  });
+});
+
+describe("patchVersionCheckState", () => {
+  beforeEach(() => {
+    mkdirSync(join(tmpHome, ".agentcoop"), { recursive: true });
+  });
+
+  afterEach(() => {
+    rmSync(tmpHome, { recursive: true, force: true });
+  });
+
+  test("preserves unknown top-level keys when patching fields", () => {
+    writeFileSync(
+      configPath(),
+      JSON.stringify({
+        owners: ["aicers"],
+        cloneBaseDir: "~/code",
+        language: "en",
+        futureUnknownField: { nested: "value" },
+        anotherUnknown: 42,
+      }),
+    );
+    patchVersionCheckState({
+      lastKnownVersions: { claude: "1.3.0" },
+      lastVersionCheckAt: 1_700_000_000_000,
+    });
+    const raw = JSON.parse(readFileSync(configPath(), "utf-8"));
+    expect(raw.futureUnknownField).toEqual({ nested: "value" });
+    expect(raw.anotherUnknown).toBe(42);
+    expect(raw.lastKnownVersions).toEqual({ claude: "1.3.0" });
+    expect(raw.lastVersionCheckAt).toBe(1_700_000_000_000);
+    // Existing recognized fields are left untouched as well.
+    expect(raw.owners).toEqual(["aicers"]);
+    expect(raw.cloneBaseDir).toBe("~/code");
+    expect(raw.language).toBe("en");
+  });
+
+  test("writes only the supplied fields (merges rather than replaces)", () => {
+    writeFileSync(
+      configPath(),
+      JSON.stringify({
+        owners: [],
+        lastKnownVersions: { claude: "1.0.0", codex: "0.1.0" },
+        lastVersionCheckAt: 100,
+      }),
+    );
+    // Only updating `lastKnownVersions` — timestamp should stay.
+    patchVersionCheckState({
+      lastKnownVersions: { claude: "1.1.0", codex: "0.1.0" },
+    });
+    const raw = JSON.parse(readFileSync(configPath(), "utf-8"));
+    expect(raw.lastKnownVersions).toEqual({ claude: "1.1.0", codex: "0.1.0" });
+    expect(raw.lastVersionCheckAt).toBe(100);
+  });
+
+  test("creates the file when it does not exist", () => {
+    rmSync(configPath(), { force: true });
+    patchVersionCheckState({
+      lastKnownVersions: { claude: "1.3.0" },
+    });
+    const raw = JSON.parse(readFileSync(configPath(), "utf-8"));
+    expect(raw.lastKnownVersions).toEqual({ claude: "1.3.0" });
+  });
+
+  test("written file ends with newline", () => {
+    patchVersionCheckState({
+      lastKnownVersions: { claude: "1.3.0" },
+    });
+    expect(readFileSync(configPath(), "utf-8").endsWith("\n")).toBe(true);
+  });
+
+  test("preserves unknown nested CLI entries inside lastKnownVersions", () => {
+    // `loadLastKnownVersions` filters the runtime view down to
+    // `{ claude?, codex? }`, so an `updates.lastKnownVersions` built
+    // from that view does not know about other CLI entries.  The
+    // patcher must merge against the existing nested object — not
+    // replace it — so a forward-compatible field (e.g. `gemini` from
+    // a newer AgentCoop version) or a hand-added entry survives a
+    // claude/codex-only check.
+    writeFileSync(
+      configPath(),
+      JSON.stringify({
+        owners: [],
+        lastKnownVersions: {
+          claude: "1.0.0",
+          codex: "0.46.0",
+          gemini: "0.5.0",
+        },
+      }),
+    );
+    patchVersionCheckState({
+      lastKnownVersions: { claude: "1.3.0", codex: "0.46.0" },
+    });
+    const raw = JSON.parse(readFileSync(configPath(), "utf-8"));
+    expect(raw.lastKnownVersions).toEqual({
+      claude: "1.3.0",
+      codex: "0.46.0",
+      gemini: "0.5.0",
+    });
   });
 });
--- a/src/config.ts
+++ b/src/config.ts
@@ -50,6 +50,22 @@ export interface Config {
     claude?: Array<{ name: string; value: string }>;
     codex?: Array<{ name: string; value: string }>;
   };
+  /**
+   * Most recently observed CLI versions, updated after the startup
+   * version check.  Used as the "previous" side of the comparison on
+   * the next run so the user can spot regressions across runs.
+   */
+  lastKnownVersions?: {
+    claude?: string;
+    codex?: string;
+  };
+  /** When `true`, the startup version check is skipped entirely. */
+  skipVersionCheck?: boolean;
+  /**
+   * Epoch milliseconds of the last successful version check.  Used to
+   * throttle network calls to roughly once per 24h.
+   */
+  lastVersionCheckAt?: number;
 }
 
 const DEFAULT_CONFIG: Config = {
@@ -100,6 +116,22 @@ function loadModelEntries(
         typeof (e as Record<string, unknown>).value === "string",
     )
     .map((e) => ({ name: e.name, value: e.value }));
+}
+
+function loadLastKnownVersions(
+  raw: unknown,
+): Config["lastKnownVersions"] | undefined {
+  if (typeof raw !== "object" || raw === null || Array.isArray(raw)) {
+    return undefined;
+  }
+  const r = raw as Record<string, unknown>;
+  const result: NonNullable<Config["lastKnownVersions"]> = {};
+  if (typeof r.claude === "string") result.claude = r.claude;
+  if (typeof r.codex === "string") result.codex = r.codex;
+  if (result.claude === undefined && result.codex === undefined) {
+    return undefined;
+  }
+  return result;
 }
 
 function loadCustomModels(raw: unknown): Config["customModels"] | undefined {
@@ -199,6 +231,16 @@ export function loadConfig(): Config {
         ? raw.executionMode
         : undefined,
     customModels: loadCustomModels(raw.customModels),
+    lastKnownVersions: loadLastKnownVersions(raw.lastKnownVersions),
+    skipVersionCheck:
+      typeof raw.skipVersionCheck === "boolean"
+        ? raw.skipVersionCheck
+        : undefined,
+    lastVersionCheckAt:
+      typeof raw.lastVersionCheckAt === "number" &&
+      Number.isFinite(raw.lastVersionCheckAt)
+        ? raw.lastVersionCheckAt
+        : undefined,
   };
 }
 
@@ -261,4 +303,58 @@ export function saveConfig(config: Config): void {
   const path = configPath();
   mkdirSync(dirname(path), { recursive: true });
   writeFileSync(path, `${JSON.stringify(config, null, 2)}\n`);
+}
+
+/**
+ * Patch only the version-check state fields in `~/.agentcoop/config.json`
+ * without round-tripping through the normalized `Config` shape.
+ *
+ * `loadConfig` / `saveConfig` drop top-level keys that the schema does
+ * not recognize, so using `saveConfig` to persist these fields would
+ * silently delete any unknown entries a user or future version may have
+ * added.  This helper reads the raw JSON, merges the supplied fields,
+ * and writes the file back, preserving unknown keys.
+ */
+export function patchVersionCheckState(updates: {
+  lastKnownVersions?: NonNullable<Config["lastKnownVersions"]>;
+  lastVersionCheckAt?: number;
+}): void {
+  const path = configPath();
+  let raw: Record<string, unknown> = {};
+  if (existsSync(path)) {
+    try {
+      const parsed = JSON.parse(readFileSync(path, "utf-8"));
+      if (
+        typeof parsed === "object" &&
+        parsed !== null &&
+        !Array.isArray(parsed)
+      ) {
+        raw = parsed as Record<string, unknown>;
+      }
+    } catch {
+      // Unreadable or invalid JSON — fall through and write a fresh file.
+    }
+  }
+  if (updates.lastKnownVersions !== undefined) {
+    // Merge against the existing nested object instead of replacing
+    // it wholesale.  `loadLastKnownVersions` filters the runtime view
+    // down to `{ claude?, codex? }`, so an `updates.lastKnownVersions`
+    // built from that view does not know about other CLI entries
+    // (e.g. a `gemini` field a future AgentCoop version may write, or
+    // a hand-added user entry).  Replacing would silently drop those
+    // — the same class of failure that motivated this helper for
+    // unknown top-level keys.
+    const existing =
+      typeof raw.lastKnownVersions === "object" &&
+      raw.lastKnownVersions !== null &&
+      !Array.isArray(raw.lastKnownVersions)
+        ? (raw.lastKnownVersions as Record<string, unknown>)
+        : {};
+    raw.lastKnownVersions = { ...existing, ...updates.lastKnownVersions };
+  }
+  if (updates.lastVersionCheckAt !== undefined) {
+    raw.lastVersionCheckAt = updates.lastVersionCheckAt;
+  }
+  mkdirSync(dirname(path), { recursive: true });
+  writeFileSync(path, `${JSON.stringify(raw, null, 2)}\n`);
 }
--- a/src/i18n/en.ts
+++ b/src/i18n/en.ts
@@ -348,6 +348,30 @@ export const en: Messages = {
   "worktree.alreadyExists": (path) =>
     `Worktree already exists at ${path}. ` +
     "Provide a conflictChoice (reuse | clean | halt).",
+
+  // ---- version check -----------------------------------------------------
+
+  "versionCheck.checking": "Checking CLI versions...",
+  "versionCheck.inconclusive": (cli, reason) =>
+    `  ${cli}: update check skipped (${reason})`,
+  "versionCheck.fetchFailed": (cli, reason) =>
+    `  ${cli}: could not fetch latest version (${reason})`,
+  "versionCheck.upToDate": (cli, version) =>
+    `  ${cli} v${version} is up to date.`,
+  "versionCheck.updatePrompt": (cli, from, to) =>
+    `A newer version of ${cli} is available (v${from} → v${to}). Update now?`,
+  "versionCheck.updateWaiting": (cli) =>
+    `Please update ${cli}, then press Enter to continue.`,
+  "versionCheck.versionUnchanged": (version) => `Version is still v${version}.`,
+  "versionCheck.retrySkipAbortPrompt": "How would you like to proceed?",
+  "versionCheck.proceedingWith": (cli, version) =>
+    `  ${cli} updated to v${version}. Continuing.`,
+  "versionCheck.abortedByUser": "Update check aborted by user.",
+  "versionCheck.versionUnknown": (cli) =>
+    `  Could not determine ${cli} version — skipping update check.`,
+  "versionCheck.retry": "Retry",
+  "versionCheck.skip": "Skip (use current version)",
+  "versionCheck.abort": "Abort",
   "worktree.haltConflict":
     "User chose to halt \u2014 worktree conflict unresolved.",
 };
--- a/src/i18n/ko.ts
+++ b/src/i18n/ko.ts
@@ -390,6 +390,31 @@ export const ko: Messages = {
 
   "notification.title": "agentcoop",
 
+  // ---- version check -----------------------------------------------------
+
+  "versionCheck.checking": "CLI 버전 확인 중...",
+  "versionCheck.inconclusive": (cli, reason) =>
+    `  ${cli}: 업데이트 검사 건너뜀 (${reason})`,
+  "versionCheck.fetchFailed": (cli, reason) =>
+    `  ${cli}: 최신 버전을 가져올 수 없습니다 (${reason})`,
+  "versionCheck.upToDate": (cli, version) =>
+    `  ${cli} v${version} 최신 버전입니다.`,
+  "versionCheck.updatePrompt": (cli, from, to) =>
+    `${cli}의 새 버전이 있습니다 (v${from} → v${to}). 지금 업데이트하시겠습니까?`,
+  "versionCheck.updateWaiting": (cli) =>
+    `${cli}를 업데이트한 후 Enter 키를 눌러 계속 진행하세요.`,
+  "versionCheck.versionUnchanged": (version) =>
+    `버전이 여전히 v${version}입니다.`,
+  "versionCheck.retrySkipAbortPrompt": "어떻게 진행하시겠습니까?",
+  "versionCheck.proceedingWith": (cli, version) =>
+    `  ${cli}가 v${version}로 업데이트되었습니다. 계속 진행합니다.`,
+  "versionCheck.abortedByUser": "사용자에 의해 업데이트 검사가 중단되었습니다.",
+  "versionCheck.versionUnknown": (cli) =>
+    `  ${cli} 버전을 확인할 수 없어 업데이트 검사를 건너뜁니다.`,
+  "versionCheck.retry": "재시도",
+  "versionCheck.skip": "건너뛰기 (현재 버전 사용)",
+  "versionCheck.abort": "중단",
+
   // ---- worktree errors ---------------------------------------------------
 
   "worktree.alreadyExists": (path) =>
--- a/src/i18n/messages.ts
+++ b/src/i18n/messages.ts
@@ -315,4 +315,25 @@ export interface Messages {
 
   "worktree.alreadyExists": (path: string) => string;
   "worktree.haltConflict": string;
+
+  // ---- version check -----------------------------------------------------
+
+  "versionCheck.checking": string;
+  "versionCheck.inconclusive": (cli: string, reason: string) => string;
+  "versionCheck.fetchFailed": (cli: string, reason: string) => string;
+  "versionCheck.upToDate": (cli: string, version: string) => string;
+  "versionCheck.updatePrompt": (
+    cli: string,
+    from: string,
+    to: string,
+  ) => string;
+  "versionCheck.updateWaiting": (cli: string) => string;
+  "versionCheck.versionUnchanged": (version: string) => string;
+  "versionCheck.retrySkipAbortPrompt": string;
+  "versionCheck.proceedingWith": (cli: string, version: string) => string;
+  "versionCheck.abortedByUser": string;
+  "versionCheck.versionUnknown": (cli: string) => string;
+  "versionCheck.retry": string;
+  "versionCheck.skip": string;
+  "versionCheck.abort": string;
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 
-import { confirm, select } from "@inquirer/prompts";
+import { confirm, input, select } from "@inquirer/prompts";
 
 import type { AgentAdapter, AgentStream } from "./agent.js";
 import { type BootstrapLog, createBootstrapLog } from "./bootstrap-log.js";
@@ -24,6 +24,7 @@ import {
   assembleReviewStage,
   assembleSquashStage,
   loadConfig,
+  patchVersionCheckState,
 } from "./config.js";
 import { createDonePromptOptions } from "./done-prompt-options.js";
 import { getGitHubUsername, getIssue } from "./github.js";
@@ -77,6 +78,12 @@ import {
   selectTarget,
 } from "./startup.js";
 import { renderApp } from "./ui/render-app.js";
+import {
+  type CliName,
+  refreshAgentCliVersion,
+  runStartupVersionCheck,
+  VersionCheckAbortError,
+} from "./version-check.js";
 import {
   bootstrapRepo,
   createWorktree,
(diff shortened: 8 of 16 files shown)


sehkone:
**[Reviewer Round 1]**

1. `src/version-check.ts:502` defeats both `skipVersionCheck` and the 24h throttle. `runStartupVersionCheck()` computes `throttled`, but it still calls `checkCliVersion()` before branching on it. `checkCliVersion()` always resolves the install source and, for recognized layouts, fetches the remote latest version, so a throttled/skipped run still hits npm/GitHub/Homebrew. That contradicts the issue, README/CHANGELOG, and the PR test plan. The existing `skipVersionCheck=true` coverage in `src/version-check.test.ts:507` only checks that no prompt is shown; it never asserts that `fetchJson` was suppressed, so this slipped through.

2. `src/version-check.ts:149` misses the current standalone Codex install layout, so the GitHub-release update check is skipped on a real standalone install. On this machine, `realpath(command -v codex)` resolves to `~/.codex/packages/standalone/releases/0.122.0-aarch64-apple-darwin/codex`, but `detectCodexSource()` only recognizes `.codex/bin` and `.codex/versions`. That means the implementation treats a real standalone install as `inconclusive`, which drops one of the issue's required channels. The test at `src/version-check.test.ts:106` only covers the older `.codex/bin/...` shape, so it doesn't catch the regression.

sehkone:

sehkone:
**[Author Round 1]**

1. **Fixed.** `runStartupVersionCheck` previously called `checkCliVersion()` unconditionally and then branched on `throttled`, so a throttled/skipped run still did install-source detection and the network fetch. Split the flow: a new `probeInstalledVersion()` does only `runVersion` + `parseVersion` (no `resolveBinary`, no `fetchJson`), and the orchestrator uses that on the throttled/skipped path. Strengthened the existing `skipVersionCheck=true` test to spy on `fetchJson` **and** `resolveBinary` and assert neither is called, and added a separate test for the `lastVersionCheckAt` within-24h throttle path asserting the same. Also suppressed the `checking` log line on throttled runs so the UX doesn't claim a check is happening when it isn't. (`src/version-check.ts`, `src/version-check.test.ts`)

2. **Fixed.** `detectCodexSource` previously only recognized `~/.codex/bin/` and `~/.codex/versions/`, which misses the current standalone installer layout `~/.codex/packages/standalone/releases/<ver>-<triple>/codex`. Broadened the match to any path containing a `/.codex/` segment — all three shapes (`bin`, `versions`, `packages/standalone/releases`) resolve to `github-releases`, and anything outside `~/.codex/` still falls through to `inconclusive`. Added regression tests for both the legacy `~/.codex/versions/<ver>/` layout and the current `~/.codex/packages/standalone/releases/...` layout. README/CLI-update-check section and PR description updated to reflect the broader match. (`src/version-check.ts`, `src/version-check.test.ts`, `README.md`)

`pnpm check`, `pnpm tsc --noEmit`, and `pnpm vitest run` (1927 tests) all pass.

sehkone:
**[Reviewer Round 2]**

I re-checked the two round-1 issues. Both are fixed: throttled/skipped runs now use `probeInstalledVersion()` without network/install-source resolution, and the current `~/.codex/packages/standalone/releases/...` layout is covered.

1. `src/ui/App.tsx:355-361`, `src/ui/AgentPane.tsx:116-126`: the pane header still does not show the CLI name, only the model name plus a version suffix. For Codex that renders as `Agent B (reviewer) — GPT-5.4 v0.46.0`, so the header still does not answer “which CLI build is driving this pane?” when model names are ambiguous or shared. The issue explicitly asked for the CLI name with its version, and `cliTypeA` / `cliTypeB` are already available here but unused. The new assertions in `src/ui/components.test.tsx:55-70` currently lock in the model+version form, so this gap would persist unnoticed.

2. `src/version-check.ts:149-156`: the standalone Codex detection fix is now too broad. Matching any realpath that contains a `/.codex/` segment classifies it as `github-releases`, which means a hand-placed/custom binary like `/tmp/.codex/codex` would now get an official-release update prompt instead of the required `inconclusive` skip. The issue was explicit that unrecognized/custom layouts must not be guessed because a wrong prompt is worse than no prompt. The added tests in `src/version-check.test.ts:106-129` only cover `~/.codex/...` positives plus one unrelated negative path, so this false-positive case is still untested.

sehkone:

sehkone:
**[Author Round 2]**

1. **Fixed.** The pane header was only showing the model name plus a version suffix, so `GPT-5.4 v0.46.0` on the Codex pane didn't answer "which CLI build is driving this pane?" when model names overlap. `formatPaneHeader` now takes an optional `cliName` argument and renders it ahead of the model name: `Agent A (author) — Claude Opus 4.6 v1.2.3` / `Agent B (reviewer) — Codex GPT-5.4 v0.46.0`. `AgentPane` gained a matching `cliName` prop, and `App.tsx` derives it from the existing `cliTypeA` / `cliTypeB` via the already-exported `cliDisplayName`, so no new plumbing from `index.ts` is required. The new rendering test `src/ui/components.test.tsx` asserts the `Codex GPT-5.4 v0.46.0` shape (replacing the previous gap where the assertions only locked in model+version), plus three new `formatPaneHeader` unit cases for CLI+model+version, CLI+version-only, and CLI-only. The README ASCII diagram and CHANGELOG entry were updated to match. (`src/ui/AgentPane.tsx`, `src/ui/App.tsx`, `src/ui/components.test.tsx`, `README.md`, `CHANGELOG.md`)

2. **Fixed.** `detectCodexSource` previously matched any realpath containing a `/.codex/` segment, which would classify `/tmp/.codex/codex` or any other rogue `.codex` directory as a `github-releases` install and produce a wrong update prompt — the exact "guess a channel" failure the issue warns against. The check is now anchored to `${homedir()}${sep}.codex${sep}`: only paths rooted at the current user's home directory count as a standalone install; everything else falls through to `inconclusive` so the check is skipped rather than wrong. Added a regression test `a rogue /.codex/ path outside \$HOME is inconclusive, not guessed` covering `/tmp/.codex/codex`, `/var/tmp/fake/.codex/bin/codex`, and `/opt/shared/.codex/versions/0.46.0/codex` — all classified as inconclusive. The three existing positive-standalone tests were switched from hardcoded `/Users/user/.codex/...` paths to `\`\${homedir()}/.codex/...\`` so they continue to pass in any CI environment. README and CHANGELOG updated to clarify that only home-rooted layouts count. (`src/version-check.ts`, `src/version-check.test.ts`, `README.md`, `CHANGELOG.md`)

`pnpm check`, `pnpm tsc --noEmit`, `pnpm markdownlint-cli2`, and `pnpm vitest run` (1932 tests) all pass.

sehkone:
**[Reviewer Round 3]**

I re-checked the two Round 2 items; both are fixed in the updated diff.

1. `src/version-check.ts:141-147` still guesses the Codex install channel too broadly for Homebrew cask. Any resolved binary path containing `/Applications/` is classified as `homebrew-cask`, so a hand-placed/custom binary such as `~/Applications/Codex.app/Contents/MacOS/codex` or `/tmp/Applications/codex` would get an official cask update prompt instead of the required `inconclusive` skip. The issue was explicit that unrecognized/custom layouts must not be guessed because a wrong prompt is worse than no prompt. This detection needs the same kind of tightening you applied for the `~/.codex/` path, and it should have a regression test covering a non-cask `Applications` path.

sehkone:

sehkone:
**[Author Round 3]**

1. **Fixed.** `detectCodexSource` previously matched any path containing `${sep}Applications${sep}` as `homebrew-cask`, which would classify hand-placed or custom binaries like `~/Applications/Codex.app/Contents/MacOS/codex` and `/tmp/Applications/codex` as official cask installs and emit a bogus update prompt — the same "guess a channel" failure the issue warns against. I went one step stronger than mirroring the `~/.codex/` home-anchoring: because we already `realpathSync` the binary before classifying it, a real cask install always resolves into a Caskroom directory under the Homebrew prefix. If we still see a raw `Applications/` path at detection time, the binary is provably not a cask symlink, so *any* `Applications/` classification would be a guess regardless of whether it's home-rooted. Dropped the `Applications` check entirely — only `/opt/homebrew/Caskroom/`, `/usr/local/Caskroom/`, and `/linuxbrew/Caskroom/` count as cask now; anything else falls through to `inconclusive`. Added a regression test `any /Applications/ path is inconclusive, not guessed as cask` covering `/Applications/Codex.app/Contents/MacOS/codex`, `${homedir()}/Applications/Codex.app/Contents/MacOS/codex`, `/tmp/Applications/codex`, and `/opt/shared/Applications/codex/bin/codex` — all return `inconclusive`. Extended the existing `Homebrew cask path` test to also cover `/usr/local/Caskroom/…` so the valid cask layouts remain pinned. README's "CLI update check" section and CHANGELOG entry updated to document the stricter cask rule and explain that an unresolved `Applications/` path is intentionally inconclusive. PR description also refreshed to match. (`src/version-check.ts`, `src/version-check.test.ts`, `README.md`, `CHANGELOG.md`)

`pnpm check`, `pnpm tsc --noEmit`, `pnpm markdownlint-cli2`, and `pnpm vitest run` (1933 tests) all pass.

sehkone:
**[Reviewer Round 4]**

The Round 3 `Applications`/cask issue is fixed in the updated diff.

1. `src/version-check.ts:175-177` still treats any binary under `~/.codex/` as an official `github-releases` install. That is broader than the issue's "recognized standalone layout" rule and will mis-prompt on hand-placed/custom binaries such as `~/.codex/tools/codex` or `~/.codex/tmp/codex`. The requirement was to recognize the known standalone layouts (`bin`, `versions/<ver>`, `packages/standalone/releases/...`) and treat everything else as inconclusive because a wrong prompt is worse than no prompt. The current tests in `src/version-check.test.ts:134-172` pin the positive layouts and the outside-`$HOME` negatives, but there is no regression case for an unknown subpath inside `~/.codex/`, so this false positive still passes.

sehkone:

sehkone:
**[Author Round 4]**

1. **Fixed.** `detectCodexSource` previously matched any binary whose realpath started with `${homedir()}/.codex/`, which would classify a hand-placed or custom binary like `~/.codex/tools/codex` or `~/.codex/tmp/codex` as an official `github-releases` install — exactly the "guess a channel" failure the issue warns against (home-rooting rejected the outside-`$HOME` rogues but was necessary, not sufficient). Tightened the match to the three documented standalone layouts only: `~/.codex/bin/`, `~/.codex/versions/`, and `~/.codex/packages/standalone/releases/`. Any other subpath under `~/.codex/` now falls through to `inconclusive` so the check is skipped rather than wrong. Added a regression test `unknown subpath inside ~/.codex/ is inconclusive, not guessed` covering `~/.codex/tools/codex`, `~/.codex/tmp/codex`, the bare `~/.codex/codex`, and a non-standalone `~/.codex/packages/custom/codex` — all classified as inconclusive. The three existing positive-standalone tests (`bin`, `versions/<ver>`, current `packages/standalone/releases/...`) and the outside-`$HOME` negative test continue to pass, pinning the known layouts. README's "CLI update check" section and the CHANGELOG entry updated to state that only the known subpaths count and that unrecognized subpaths inside `~/.codex/` are intentionally inconclusive. PR description also refreshed to match. (`src/version-check.ts`, `src/version-check.test.ts`, `README.md`, `CHANGELOG.md`)

`pnpm check`, `pnpm tsc --noEmit`, `pnpm markdownlint-cli2`, and `pnpm vitest run` (1934 tests) all pass.

sehkone:
**[Reviewer Round 5]**

The round-4 `~/.codex` false-positive is fixed in the current diff. I still see two unresolved issues:

1. `src/version-check.ts:134-161` still guesses the Homebrew channel too broadly. Any resolved binary path under a Homebrew `Cellar` or `Caskroom` is treated as the official `codex` formula/cask, so a custom wrapper like `/opt/homebrew/Cellar/my-codex-wrapper/1.0/bin/codex` or `/opt/homebrew/Caskroom/custom-codex/1.0/codex` would get an official Codex update prompt instead of the required `inconclusive` skip. The issue has been explicit that unrecognized/custom installs must not be guessed because a wrong prompt is worse than no prompt. This needs to verify the `codex` path segment itself (`.../Cellar/codex/...`, `.../Caskroom/codex/...`) or otherwise prove the package name. `src/version-check.test.ts:92-108` currently only covers the positive `codex` paths, so this false-positive case would still pass.

2. `src/index.ts:771-775` combined with `src/version-check.ts:575-579` can erase the per-run version record on resume. `runStartupVersionCheck()` deliberately returns `undefined` when `--version` fails or the output cannot be parsed, but resume then unconditionally writes that `undefined` back into `savedState.agentA.cliVersion` / `savedState.agentB.cliVersion`. That discards the version already recorded for the run in `~/.agentcoop/runs/...json`, which is the postmortem record issue #276 asked to preserve. Refreshing on resume makes sense when a new version string is actually detected, but a failed re-probe should keep the existing saved value rather than blank it out. I also do not see a regression test for the "saved cliVersion + failed refresh" path.

sehkone:

sehkone:

sehkone:
**[Author Round 5]**

1. **Fixed.** `detectCodexSource` previously matched any Homebrew path that included a generic `Cellar/` or `Caskroom/` segment, so a custom wrapper like `/opt/homebrew/Cellar/my-codex-wrapper/1.0/bin/codex` or `/opt/homebrew/Caskroom/custom-codex/1.0/codex` would have been classified as the official `codex` formula/cask and produced a wrong update prompt — the exact "guess a channel" failure the issue warns against. Anchored the match on the `codex` package segment directly after `Cellar/` / `Caskroom/`, matching Brew's `<prefix>/Cellar/<formula-name>/<version>/…` and `<prefix>/Caskroom/<cask-name>/<version>/…` layout. The three Homebrew prefixes (`/opt/homebrew/`, `/usr/local/`, `/linuxbrew/`) are preserved. Added two regression tests covering custom formulas (`my-codex-wrapper`, `codex-extras`, `not-codex` under all three prefixes) and custom casks (`custom-codex`, `codex-preview`) — all classified as inconclusive. The existing positive `codex` formula/cask tests continue to pass, pinning the valid paths. README "CLI update check" and the CHANGELOG entry updated to state the anchoring explicitly. (`src/version-check.ts`, `src/version-check.test.ts`, `README.md`, `CHANGELOG.md`)

2. **Fixed.** The resume branch previously wrote `runStartupVersionCheck()`'s result unconditionally back into `savedState.agentA.cliVersion` / `savedState.agentB.cliVersion`, so a failed re-probe (where the check deliberately returns `undefined` after `--version` crashed or the output could not be parsed) erased the per-run version already recorded for the run — exactly the postmortem record issue #276 asked to preserve. Extracted a small `refreshAgentCliVersion(agent, detected)` helper in `src/version-check.ts` that overwrites the saved value only when the probe produced a defined string, and routed both agents through it in `src/index.ts`. A successful upgrade is still captured; a failed re-probe now leaves the previously recorded version intact. Added four unit tests covering the overwrite path, the regression (failed re-probe preserves saved value), the undefined-to-undefined path, and first-successful-probe on state that predates version tracking. CHANGELOG updated to call out the preservation behavior. (`src/version-check.ts`, `src/index.ts`, `src/version-check.test.ts`, `CHANGELOG.md`)

`pnpm check`, `pnpm tsc --noEmit`, `pnpm markdownlint-cli2`, and `pnpm vitest run` (1940 tests) all pass.

sehkone:

sehkone:
**[Reviewer Round 6]**

I re-checked the two Round 5 items. The `codex` package-name anchoring and the resume-time `cliVersion` preservation are both fixed in the current diff.

1. `src/version-check.ts:141-145` and `src/version-check.ts:168-173` still classify Homebrew formula/cask installs with `String.prototype.includes()`, so the detector is still broader than the issue's “under `/opt/homebrew/...` or `/usr/local/...`” rule. A hand-placed binary under a copied tree like `/tmp/opt/homebrew/Cellar/codex/0.46.0/bin/codex` or `/Users/me/sandbox/usr/local/Caskroom/codex/0.122.0/codex` is not actually installed under the Homebrew prefix, but this code will treat it as the official formula/cask and prompt against the wrong channel instead of returning `inconclusive`. Given how explicitly issue #276 says “a wrong prompt is worse than no prompt,” these checks still need to be root-anchored (`startsWith()` on the exact prefixes, or equivalent path-segment parsing). `src/version-check.test.ts:93-139` should also grow regression coverage for these prefixed lookalikes.

sehkone:

sehkone:
**[Author Round 6]**

1. **Fixed.** `detectCodexSource` previously classified Homebrew formula/cask paths with `String.prototype.includes()`, so a hand-placed binary under a copied tree like `/tmp/opt/homebrew/Cellar/codex/0.46.0/bin/codex` or `/Users/me/sandbox/usr/local/Caskroom/codex/0.122.0/codex` would have been treated as the official formula/cask and produced a wrong update prompt — exactly the "guess a channel" failure the issue warns against. Switched both checks to `String.prototype.startsWith()` on the exact Homebrew prefixes (`/opt/homebrew/Cellar/codex/`, `/usr/local/Cellar/codex/`, `/linuxbrew/Cellar/codex/` for formulas, and the corresponding `Caskroom` paths for casks), so only binaries actually rooted under a Homebrew prefix qualify and prefixed lookalikes fall through to `inconclusive`. The existing positive tests for real Homebrew paths and the existing "custom formula/cask name" negatives continue to pass. Added a new regression test `prefixed Homebrew lookalikes are inconclusive, not guessed` covering both Cellar and Caskroom variants under all three prefixes: `/tmp/opt/homebrew/Cellar/codex/...`, `/Users/me/sandbox/usr/local/Cellar/codex/...`, `/var/tmp/linuxbrew/Cellar/codex/...`, and the matching Caskroom triples — all classified as inconclusive. README's "CLI update check" section, the CHANGELOG entry, and the PR description updated to state the root-anchoring explicitly. (`src/version-check.ts`, `src/version-check.test.ts`, `README.md`, `CHANGELOG.md`)

`pnpm check`, `pnpm tsc --noEmit`, `pnpm markdownlint-cli2`, and `pnpm vitest run` (1941 tests) all pass.

sehkone:
**[Reviewer Round 7]**

The Round 6 Homebrew root-anchoring fix is present in the updated diff. I still see two unresolved issues:

1. `src/version-check.ts:145-150` and `src/version-check.ts:177-182` hard-code `/linuxbrew/...` as the Linux Homebrew prefix. Homebrew's supported default Linux prefix is `/home/linuxbrew/.linuxbrew`, so a normal Linux install at `/home/linuxbrew/.linuxbrew/Cellar/codex/...` falls through to `inconclusive` and never gets the update check this feature is meant to provide. The new tests in `src/version-check.test.ts:93-171` also lock in `/linuxbrew/...`, so they would not catch the real default layout.

2. `src/index.ts:778-779` preserves `runState.agentA.cliVersion` / `runState.agentB.cliVersion` on resume when the re-probe returns `undefined`, but `src/index.ts:815-827` still passes the raw `agentAVersion` / `agentBVersion` into `createRunLog()`. On a resume where `--version` fails, the saved run-state keeps the previous version as intended while the new run-log header drops it entirely. That breaks the advertised “failed re-probe preserves the postmortem record” behavior for the log path. I also do not see regression coverage for the `version  : ...` header or this resume-preservation case in `src/run-log.test.ts`.

sehkone:
[Review Verdict Round 7: NOT_APPROVED]

sehkone:
[Review Verdict Round 7: NOT_APPROVED]

sehkone:
**[Author Round 7]**

1. **Fixed.** The Linux Homebrew prefix was hard-coded as `/linuxbrew/`, which is not Homebrew's documented default — the supported default Linux prefix is `/home/linuxbrew/.linuxbrew`, so a normal Linuxbrew install at `/home/linuxbrew/.linuxbrew/Cellar/codex/...` (and the matching `Caskroom/codex/...`) fell through to `inconclusive` and never got the update check this feature is meant to provide. Replaced the bogus prefix with the documented default in both the formula and cask checks, in both README "CLI update check" and the CHANGELOG entry. Extended the existing `Homebrew formula path` / `Homebrew cask path` positive tests to cover the new prefix, updated the `custom Homebrew formula` negative to use the real Linuxbrew prefix so the package-segment anchoring is still pinned, and updated the `prefixed Homebrew lookalikes are inconclusive` regressions so the "copied tree" cases now use `/var/tmp/home/linuxbrew/.linuxbrew/...` — i.e. the same lookalike pattern applied to the real Linux prefix. (`src/version-check.ts`, `src/version-check.test.ts`, `README.md`, `CHANGELOG.md`)

2. **Fixed.** Correct diagnosis: `refreshAgentCliVersion` already preserved `runState.agent*.cliVersion` on a failed re-probe, but `createRunLog()` was still being passed the raw `agentAVersion` / `agentBVersion` from the probe, so a failed re-probe on resume would drop the `version  : ...` line from the new run-log header while the saved `RunState` kept the previous build. That breaks the advertised "failed re-probe preserves the postmortem record" behavior for the log path. Threaded the header's `cliVersion` through `runState.agentA.cliVersion` / `runState.agentB.cliVersion` — the refresh helper already owns the preservation decision, so the log path now inherits it for free. Added two regression tests in `src/run-log.test.ts`: `records cliVersion in header when provided` (covers the `version  : 1.2.3` and `version  : 0.46.0` lines) and `omits version line when cliVersion is undefined` (covers the fresh-probe-failed case where the header must not emit a stray `version  : undefined`). Combined with the existing `refreshAgentCliVersion` unit tests — specifically `keeps the saved version when the re-probe returns undefined` — the resume-preserves-on-failed-re-probe path is now covered end-to-end: refresh preserves the saved value in `runState` → `runState.agent*.cliVersion` is what `createRunLog` reads → the run-log header records it. (`src/index.ts`, `src/run-log.test.ts`, updated PR description)

`pnpm check`, `pnpm tsc --noEmit`, `pnpm markdownlint-cli2`, and `pnpm vitest run` (1943 tests) all pass.

sehkone:
**[Reviewer Round 8]**

I re-checked the two Round 7 items. Both are fixed: Linuxbrew now matches `/home/linuxbrew/.linuxbrew`, and the run-log header now reads `cliVersion` from `runState`, so a failed resume-time re-probe no longer drops the saved version from the log path.

1. `src/version-check.ts:131-157` still gives npm precedence over Homebrew formula detection. A realpath like `/opt/homebrew/Cellar/codex/0.46.0/libexec/lib/node_modules/@openai/codex/bin/codex.js` is still "under `/opt/homebrew/Cellar/...`" per the issue's channel rules, so it should resolve to the Homebrew formula channel, but `detectCodexSource()` returns npm before it ever checks the `Cellar/codex` prefix because the `node_modules/@openai/codex` substring match runs first. That points a formula install at the npm registry instead of formula metadata and can produce the wrong update prompt when those channels diverge, which is exactly the case this feature is meant to avoid. `src/version-check.test.ts:80-109` only covers the direct `.../Cellar/codex/.../bin/codex` shape, so this overlap case still passes untested.

sehkone:
[Review Verdict Round 8: NOT_APPROVED]

sehkone:
**[Author Round 8]**

1. **Fixed.** You're right — a Homebrew formula that bundles its JS internally (e.g. `/opt/homebrew/Cellar/codex/0.46.0/libexec/lib/node_modules/@openai/codex/bin/codex.js`) was being routed to the npm channel because the `node_modules/@openai/codex/` substring match ran before the `Cellar/codex/` prefix check. Reordered `detectCodexSource()` in `src/version-check.ts` so the Homebrew formula and cask checks (both root-anchored on the documented Homebrew prefixes) run before the npm substring match. Added a regression test in `src/version-check.test.ts` (`Homebrew formula bundling node_modules resolves to formula, not npm`) covering all three Homebrew prefixes (`/opt/homebrew`, `/usr/local`, `/home/linuxbrew/.linuxbrew`) for the bundled-`node_modules` shape. `pnpm vitest run`, `pnpm tsc --noEmit`, and `pnpm biome check` all pass.

sehkone:
**[Reviewer Round 9]**

I re-checked the Round 8 Homebrew-formula-vs-npm overlap; that fix is present. I still see two issues in the current diff:

1. `src/ui/App.tsx:356-361`, `src/startup.ts:97-110`, `models.json:2-4`, `src/ui/AgentPane.tsx:123-133`: the new header composition duplicates Claude's name. `modelDisplayName(agentAConfig)` already resolves to strings like `Claude Opus 4.7`, and `App` now prepends `cliDisplayName("claude")` before passing the header through `formatPaneHeader()`. On a normal Claude run this renders `Agent A (author) — Claude Claude Opus 4.7 ...`, not the `Claude Opus ...` form described in the issue, README, and CHANGELOG. The tests only cover `formatPaneHeader` in isolation and a Codex-pane render, so this regression is untested. Either strip the CLI prefix from the model display string before composing the pane header, or avoid prepending `cliName` when the display name already starts with it.

2. `src/version-check.ts:694-699`, `src/config.ts:65-68`, `README.md:283-286`: `runStartupVersionCheck()` writes `lastVersionCheckAt` on every non-throttled attempt, even when every check was inconclusive or the network fetch failed. That means one transient registry outage (or a custom Codex layout that always returns `inconclusive`) suppresses all re-checks for the next 24h, even though no successful latest-version comparison ever happened. That conflicts with the code/docs' own "last successful version check" semantics and makes the update prompt much easier to miss. The timestamp should only advance when at least one CLI actually reached a definitive installed-vs-latest comparison / completion path, and there should be a regression test for the failure/inconclusive case.

sehkone:
[Review Verdict Round 9: NOT_APPROVED]

sehkone:
**[Author Round 9]**

1. **Fixed.** You're right — `models.json` entries like `Claude Opus 4.7` already embed the CLI name, so `App.tsx` prepending `cliDisplayName("claude")` produced `Agent A (author) — Claude Claude Opus 4.7 ...`. Fixed in `src/ui/AgentPane.tsx` by teaching `formatPaneHeader()` to drop the CLI prefix when `modelName` already begins with `cliName` (case-insensitive, with a word-boundary check so a hypothetical `Claudette Turbo` model still gets the prefix). Codex panes are unaffected because `GPT-5.4` does not start with `Codex`, so the disambiguating prefix is preserved there. Added three regression tests in `src/ui/components.test.tsx`: the `Claude Opus 4.7` duplication case, a case-insensitive variant, and the `Claudette` word-boundary guard.

2. **Fixed.** Reworded the semantics exactly as you described. `runStartupVersionCheck()` in `src/version-check.ts` now tracks a `didDefinitiveCheck` flag that is set only when a CLI produced both an installed version AND a latest version (i.e. reached `compareVersions()` / `upToDate` / the interactive update flow). `lastVersionCheckAt` is only advanced when that flag is true AND the run wasn't throttled — so a fully-inconclusive run, or one where the registry fetch failed for every CLI, re-checks on the next start rather than being silently throttled for 24h. A mixed run where at least one CLI reached comparison still advances, which matches the "last successful version check" semantics. Added three regression tests in `src/version-check.test.ts`: inconclusive-only run doesn't advance, network-fetch-failure doesn't advance, mixed success does advance. Updated `CHANGELOG.md` and `README.md` to document the refined semantics.

`pnpm vitest run` (1948 tests), `pnpm tsc --noEmit`, and `pnpm biome check` all pass.

sehkone:
**[Reviewer Round 10]**

I re-checked the two Round 9 items; both are fixed in the updated diff.

1. `src/version-check.ts:305-318`, `src/version-check.ts:647-650`, `src/version-check.test.ts:374-382`: the standalone Codex update check still parses the GitHub latest-release version incorrectly. As of 2026-04-21, `gh api repos/openai/codex/releases/latest` returns `tag_name: rust-v0.122.0` and `name: 0.122.0`. `fetchGitHubReleasesLatest()` currently strips only a leading `v`, so it turns that into `rust-v0.122.0` instead of a semver. When `runStartupVersionCheck()` compares an installed standalone build like `0.121.0` against that string, `compareVersions()` hits the first segment mismatch (`0` vs `rust`) and treats the installed version as newer, so the GitHub-release channel will never prompt for updates. The new test coverage still hard-codes the older `v0.122.0` shape, so it does not protect the real API response. This needs to normalize the GitHub release value to an actual version string before comparison (for example by using `name` when it is semver-like, or by extracting the semver from `tag_name`) and add a regression test for the current `rust-v...` tag format.

sehkone:
[Review Verdict Round 10: NOT_APPROVED]

sehkone:
**[Author Round 10]**

1. **Fixed.** You're right — `fetchGitHubReleasesLatest()` was doing a bare `^v` strip, which leaves `rust-v0.122.0` intact and makes `compareVersions()` treat any real `0.x.y` install as ahead of the latest. Updated it to extract the first semver substring from `name` first (upstream often sets `name` to the bare semver) and fall back to `tag_name` when `name` is absent or non-semver — both go through `parseVersion()`, so any prefix like `rust-v` or `codex-`, or a trailing build suffix, is handled uniformly. Added three regression tests in `src/version-check.test.ts`: `rust-v0.122.0` tag with `0.122.0` name (the observed shape), `rust-v...` tag with `name` missing (fallback path), and a non-semver release (`nightly`) that should return `undefined` so the check is skipped rather than wrong.

sehkone:
**[Reviewer Round 11]**

I re-checked the Round 10 GitHub-release normalization issue; that fix is present in the updated diff.

1. `src/version-check.ts:695-719` reintroduces the config-clobbering problem that `startup.ts` explicitly avoids. `runStartupVersionCheck()` mutates a `Config` returned by `loadConfig()` and then calls `saveConfig()` whenever `lastKnownVersions` changes or `lastVersionCheckAt` advances. But `loadConfig()` drops unknown top-level keys (`src/config.ts:191-245`), and `saveConfig()` rewrites only the normalized shape (`src/config.ts:302-305`). So the first successful version check will silently delete any unrecognized config fields from `~/.agentcoop/config.json`, even though `runStartup()` goes out of its way not to rewrite the file on every run for exactly this reason (`src/startup.ts:284-303`). Please preserve unknown config fields when updating the version-check metadata, or patch just the three new fields into the raw JSON instead of round-tripping through `Config`.

sehkone:
[Review Verdict Round 11: NOT_APPROVED]

sehkone:
**[Author Round 11]**

1. **Fixed** — `runStartupVersionCheck()` no longer round-trips through `Config`. Added `patchVersionCheckState()` in `src/config.ts` that reads the raw JSON, merges only `lastKnownVersions` / `lastVersionCheckAt`, and writes the file back, so unknown top-level keys survive the update. The version-check module's `saveConfig: (c: Config) => void` option is replaced with a narrower `persistVersionCheckState(updates)` callback that receives only the patchable fields, and `index.ts` is wired to `patchVersionCheckState`. A new `patchVersionCheckState` suite in `config.test.ts` seeds the file with unknown top-level entries and asserts they pass through; `version-check.test.ts` also asserts the callback receives only the two patch fields, not a full `Config`. CHANGELOG entry updated to call out that persistence uses a narrow patch rather than a full `saveConfig` round-trip.

sehkone:
**[Reviewer Round 12]**

I re-checked the Round 11 config-clobbering item; the narrow `patchVersionCheckState()` wiring is present, so that concern is resolved.

1. `src/version-check.ts:196-204` still guesses the npm channel too broadly. The issue says npm only when the resolved binary is under an npm global prefix, but the current check accepts any path containing `node_modules/@openai/codex/`. A project-local install such as `/worktree/node_modules/@openai/codex/bin/codex.js` (which `realpath(command -v codex)` can hit via `node_modules/.bin`) would therefore get an npm update prompt instead of the required `inconclusive` skip. That is the same “wrong prompt is worse than no prompt” failure we’ve been tightening elsewhere. `src/version-check.test.ts:79-90` only pins positive npm cases and does not cover a repo-local / custom `node_modules` path.

2. `src/config.ts:121-134`, `src/version-check.ts:708-738`, and `src/config.ts:338-340` still clobber unknown entries inside `lastKnownVersions`. `loadLastKnownVersions()` filters that object down to `{ claude?, codex? }`, `runStartupVersionCheck()` builds `merged` from that filtered view, and `patchVersionCheckState()` then replaces `raw.lastKnownVersions` wholesale. So if a newer AgentCoop version later adds `lastKnownVersions.gemini`, running this build and checking Claude/Codex will silently delete the Gemini entry even though the changelog now says version-check persistence preserves fields from newer versions. The new config tests at `src/config.test.ts:813-870` only cover unknown top-level keys, not unknown nested CLI entries under `lastKnownVersions`.

sehkone:
[Review Verdict Round 12: NOT_APPROVED]

sehkone:
**[Author Round 12]**

1. **Fixed.** `detectCodexSource` previously matched any path containing `${sep}node_modules${sep}@openai${sep}codex${sep}`, which `realpath(command -v codex)` can hit via a project-local `node_modules/.bin/codex` symlink (e.g. `/worktree/node_modules/@openai/codex/bin/codex.js`). That would have routed the update prompt at the global npm registry for a dependency the project deliberately pinned per-package — exactly the "wrong prompt is worse than no prompt" failure. The marker now requires `${sep}lib${sep}node_modules${sep}@openai${sep}codex${sep}`. Every npm-style global layout (system npm, nvm, asdf, Volta, custom `--prefix`) places packages under `<prefix>/lib/node_modules/...`, while a project-local install never has the `lib/` segment in front of `node_modules`, so the marker is both necessary and sufficient. The Homebrew formula check still runs first, so a formula that bundles its JS at `Cellar/codex/.../libexec/lib/node_modules/@openai/codex/...` still resolves to the formula channel rather than npm. Added `detectCodexSource > project-local node_modules is inconclusive, not guessed as npm` covering `/worktree/...`, `/Users/me/proj/...`, and `/tmp/sandbox/...` shapes. Updated the comment block and PR description to reflect the new anchor.

2. **Fixed.** `patchVersionCheckState` previously did `raw.lastKnownVersions = updates.lastKnownVersions`, which silently dropped unknown nested CLI keys because `runStartupVersionCheck` derives `updates.lastKnownVersions` from `loadLastKnownVersions`, which filters the runtime view down to `{ claude?, codex? }`. So a future `gemini` field — or a hand-added user entry — would have been wiped on every claude/codex-only check, defeating the same forward-compatibility guarantee the helper exists to provide for top-level keys. The patcher now reads the existing `raw.lastKnownVersions` (when present and an object), spreads it under the supplied updates, and writes the merged result. This mirrors the merge semantics the top-level patch already provides. Added `patchVersionCheckState > preserves unknown nested CLI entries inside lastKnownVersions` that seeds `lastKnownVersions = { claude, codex, gemini }`, runs a claude/codex-only patch, and asserts `gemini` survives. CHANGELOG and PR description updated to document the nested-merge behavior.

CI: `pnpm check`, `pnpm tsc --noEmit`, `pnpm vitest run` (all 1957 tests) all green locally.

sehkone:
**[Reviewer Round 13]**

I re-checked the two Round 12 items; both fixes are present in the updated diff.

1. `src/version-check.ts:196-217`, `src/version-check.test.ts:80-101`: the npm-channel detection is now too narrow for global `pnpm` installs. The new marker requires `.../lib/node_modules/@openai/codex/...`, but a real `pnpm add -g @openai/codex` install resolves under `.../pnpm/global/<store>/node_modules/@openai/codex/...` instead. In this environment, `pnpm root -g` is `/Users/sehkone/Library/pnpm/global/5/node_modules`, so the current logic would classify that install as `inconclusive` and skip the Codex update check entirely, even though it is still the npm-registry channel the issue wants covered. The current tests only pin `lib/node_modules/...` positives plus repo-local negatives, so this false negative is still untested. Please broaden the npm-source detection to cover known global `pnpm` layouts without reopening the project-local `node_modules/.bin` false positive, and add a regression test for a `pnpm root -g`-style path.

sehkone:
[Review Verdict Round 13: NOT_APPROVED]

sehkone:
[Review Verdict Round 13: NOT_APPROVED]

sehkone:
**[Author Round 13]**

1. **Fixed.** `detectCodexSource` previously recognized only the `lib/node_modules/@openai/codex/` marker for the npm channel, which covers system npm / nvm / asdf / Volta / custom `--prefix` but misses pnpm: `pnpm root -g` returns layouts like `~/Library/pnpm/global/<N>/node_modules` (macOS), `~/.local/share/pnpm/global/<N>/node_modules` (Linux), `%LOCALAPPDATA%/pnpm/global/<N>/node_modules` (Windows), or `$PNPM_HOME/global/<N>/node_modules` (when set) — none of which contain a `lib/` segment. A real `pnpm add -g @openai/codex` install therefore fell through to `inconclusive` and the update check was silently skipped even though it is still the npm-registry channel.

   Added a pnpm-specific marker in `src/version-check.ts` that requires both `/pnpm/global/` and `/node_modules/@openai/codex/` to appear in the realpath. The `/pnpm/global/` segment is unique to the global install location — project-local pnpm uses `<repo>/node_modules/.pnpm/...` with no `global` segment — so the existing project-local false-positive guard (the reason `lib/node_modules/` was required in the first place) stays closed. Both the top-level realpath `<pnpm-root>/@openai/codex/...` and the virtual-store shape `<pnpm-root>/.pnpm/@openai+codex@<ver>/node_modules/@openai/codex/...` match.

   New regression tests in `src/version-check.test.ts`:
   - positives for macOS (`~/Library/pnpm/global/5/node_modules/...`), Linux (`~/.local/share/pnpm/global/...`), custom `$PNPM_HOME`, and the realpath-resolved `.pnpm/...` shape,
   - negatives for project-local pnpm paths (`<repo>/node_modules/.pnpm/@openai+codex@<ver>/node_modules/@openai/codex/...`) to prove the project-local false positive is not reopened.

   The PR description's "Channel-appropriate latest resolution" bullet and the test plan are updated to mention the pnpm global layout alongside the existing npm/Homebrew/standalone handling.

   Full check passes: `pnpm check`, `pnpm tsc --noEmit`, `pnpm vitest run` (all 1959 tests pass).

sehkone:
**[Reviewer Round 14]**

I re-checked the Round 13 pnpm-global detection item. The `detectCodexSource()` change is present in `src/version-check.ts:220-240`, and `src/version-check.test.ts:93-147` now covers both the real `pnpm/global/<store>/node_modules/...` positives and the project-local `.pnpm/...` negatives that would have caught the previous miss.

I do not see any remaining issues in the updated diff against #276. The version recording, pane-header rendering, startup prompt flow, config patching, and documentation now line up with the issue as written.

sehkone:
[Review Verdict Round 14: APPROVED]

sehkone:
## Suggested squash commit

**Title**

```text
Track, display, and check CLI versions
```

**Body**

```text
AgentCoop previously had no visibility into which claude / codex
binary was driving a run, and no way to tell the user when a newer
version was available. Diagnosing regressions across CLI upgrades
and reproducing past runs both depended on the user remembering
what they had installed, and staying current depended on them
remembering to run their package manager.

Capture the installed version from `--version` before the pipeline
starts, thread it through `renderApp` into the pane header next to
the CLI name and model, and record it per-run in RunState and the
run-log header so the postmortem record survives future upgrades.
Collapse a duplicated CLI prefix in the pane header when the
configured model display name already starts with the CLI name.

After the fresh / resume branches join and the agent CLIs are
finalized, run a startup update check against the channel
appropriate source: npm registry for Claude, and for Codex a
realpath-based classification into npm global, pnpm global,
Homebrew formula, Homebrew cask, a known `~/.codex/` standalone
layout, or inconclusive. Homebrew matches are anchored on the
`codex` package segment and root-anchored on the documented
Homebrew prefixes (Apple Silicon, Intel macOS, Linuxbrew) so
lookalike paths under a non-Homebrew root and custom wrapper
packages both fall through to inconclusive instead of being
guessed. The npm marker requires the `lib/node_modules/@openai/
codex/` anchor so a project-local dependency does not get the
global update prompt, and the pnpm global marker requires both a
`/pnpm/global/` segment and the `@openai/codex` package segment so
project-local `.pnpm/` layouts likewise fall through. Homebrew
detection runs before the npm marker check so a formula that
bundles `node_modules/@openai/codex/` internally still resolves to
the formula channel. Only the known `~/.codex/bin`,
`~/.codex/versions/<ver>`, and current
`~/.codex/packages/standalone/releases/` subpaths count as a
standalone install, and `.codex` must be under the running user's
`$HOME`; anything else is inconclusive. GitHub release tags are
parsed with a semver extractor so prefixed names like `rust-v0.46.0`
still compare correctly.

When the installed version is older, prompt to update, pause for
the user to run their package manager, and re-run `--version` on
return. If the version is unchanged, offer retry / skip / abort;
there is no silent fallback. Network failure is logged but never
fatal.

Add `lastKnownVersions`, `skipVersionCheck`, and `lastVersionCheckAt`
to the config. `skipVersionCheck: true` and the 24h throttle both
suppress all network / install-source work while still probing the
installed version cheaply for the pane header. `lastVersionCheckAt`
advances only after at least one CLI reached a real
installed-vs-latest comparison, so an all-inconclusive run or a
failed registry fetch re-checks on the next start rather than being
silently throttled. Persistence uses a narrow JSON patch rather
than a full `saveConfig` round-trip so unknown top-level keys and
unknown nested `lastKnownVersions` entries (forward-compatibility
with other CLIs, user-added fields) survive the update.

On resume the version is re-probed and the RunState entry is
refreshed to the current build; a failed re-probe intentionally
preserves the previously recorded value in both RunState and the
run-log header rather than blanking the postmortem record.

Update the README Terminal UI diagram, add a `CLI update check`
section under Prerequisites documenting the skip / throttle knobs,
extend the Configuration table with the new fields, and add a
`CHANGELOG.md` entry under `[Unreleased]` > `Added`.

Closes #276
```

PR #279, opened 2026-04-21
Harden Codex adapter for newer CLI releases

### LINKED ISSUE #277: Follow-up: Harden Codex adapter for newer Codex CLI releases
## Summary

This is a follow-up technical issue to #276, scoped specifically to Codex adapter robustness across newer Codex CLI releases.

#276 covers user-facing version tracking, display, and update prompting. This issue is intentionally narrower: it focuses only on hardening the internal Codex integration so AgentCoop behaves predictably once users move to newer Codex CLI versions.

## Why this is a separate issue

#276 is about product behavior:

- record installed and used CLI versions
- display versions in the UI
- check for newer versions and prompt the user to update

This issue is about adapter correctness and compatibility:

- explicit Codex execution policy
- safer structured parsing for resumed sessions
- improved failure detection
- better compatibility coverage in tests

These changes support #276, but are not required to express its user-facing behavior and are better tracked as a follow-up implementation hardening task.

## Problem

AgentCoop currently appears broadly compatible with both older Codex CLI builds such as `0.46.0` and newer builds such as `0.122.0`, but the adapter still relies on several fragile assumptions.

As of April 21, 2026, local verification against `codex-cli 0.122.0` shows that `codex exec resume --help` advertises both `--json` and `-m`, while the current adapter and tests still assume resume is plain-text-only.

Current integration risks include:

- Codex runs do not explicitly set `approval_policy`, so behavior can depend on the user's local `~/.codex/config.toml`.
- Resume handling still relies on plain-text scraping even though newer Codex CLI versions expose structured JSON mode for resumed turns.
- Top-level JSON `error` events are declared but not treated carefully enough; newer Codex can emit retry-related `error` events before a later terminal turn event.
- The plain-text resume parser is marker-based and can misparse edge cases if the assistant output contains marker-like lines such as standalone `codex` or `tokens used` lines.
- Compatibility tests currently encode the obsolete assumption that `codex exec resume` cannot use `--json`, which makes adapter drift easier to miss.

These are internal integration risks rather than version-tracking or UI concerns.

### 1. Explicitly set approval policy for Codex runs

- Pass `-c approval_policy=never` from AgentCoop for unattended pipeline execution.
- Apply this consistently to both fresh `invoke()` calls and `resume()` calls.
- This is intentionally more specific than `--dangerously-bypass-approvals-and-sandbox`; the goal is to lock approval behavior without changing the sandbox mechanism beyond what AgentCoop already controls.
- Preserve the current sandbox behavior explicitly as well, noting the CLI asymmetry:
- fresh `codex exec` can continue using `-s danger-full-access`
- `codex exec resume` does not accept `-s`, so sandbox must continue to be passed via `-c sandbox_mode=danger-full-access`
- Document that this intentionally overrides user-local Codex approval policy for AgentCoop-managed unattended runs.

### 2. Prefer JSON mode for resumed Codex turns

- Treat JSON resume support as a current capability in newer Codex releases rather than a speculative future feature. For example, `codex-cli 0.122.0` already advertises `codex exec resume --json` in `--help`.
- Use JSON mode for `codex exec resume` on newer Codex CLI versions.
- Implement this as optimistic try-then-fallback rather than version sniffing:
- first attempt JSON resume
- if the installed CLI explicitly rejects `--json` on `resume`, or otherwise fails in a way that clearly indicates unsupported JSON resume, retry through the existing plain-text resume path
- do not retry through the plain-text path for unrelated runtime failures; surface those failures normally
- when JSON resume is used, route display streaming and structured result parsing through the same JSON event path used for fresh `invoke()` calls
- keep the existing plain-text parser only as a backward-compatibility fallback for older Codex versions

This keeps the issue independent of #276's version-detection work and avoids coupling runtime behavior to a separate version gate.

### 3. Treat top-level JSON `error` events as failures only when terminal

- Extend the JSON parser so top-level `type: "error"` is not silently ignored.
- Do not treat every `error` event as an immediate terminal failure. Newer Codex can emit retry-related `error` events and then later recover.
- Instead, surface a top-level `error` as a failure only when it is the last terminal signal for the turn, meaning no later `turn.completed` or `turn.failed` arrives.
- If a later terminal turn event exists, that later terminal event wins.
- Apply the same rule to both fresh JSON `invoke()` output and JSON-based `resume()` output.

### 4. Keep plain-text resume parsing as fallback only

- Retain plain-text parsing only for legacy compatibility.
- Make the fallback parser more defensive against marker-like content inside the assistant response.
- Prefer parsing relative to the validated banner boundary and post-banner structure, rather than scanning the whole output with a global `lastIndexOf` marker search.
- Add explicit edge-case coverage for assistant output containing standalone `codex` lines, standalone `tokens used` lines, or similar confusing marker-like content.

### 5. Add stronger compatibility coverage

Add tests for:

- JSON-mode `resume`
- JSON resume display streaming and usage forwarding when JSON events provide them
- fallback from JSON resume to plain-text resume only for older Codex versions that explicitly reject JSON resume
- no fallback for unrelated JSON resume failures
- top-level JSON `error` events that are terminal failures
- retry-style top-level JSON `error` events followed by later success, which must remain successful
- explicit `approval_policy=never` in generated Codex args for both fresh and resumed runs
- plain-text resume responses containing confusing marker-like content

## Non-goals

This issue does not cover:

- recording CLI versions
- displaying CLI versions in the UI
- startup update checks or update prompts
- install-source detection for Codex version checks

Those remain in #276.

## Open question

- Should the plain-text resume fallback remain indefinitely for backward compatibility, or should it be removed once AgentCoop defines a minimum supported Codex CLI version?

## Scope / touch points

- `src/codex-adapter.ts`
- `src/codex-adapter.test.ts`
- `src/spawn-agent.test.ts`
- `CHANGELOG.md` or equivalent release notes, to document the explicit Codex approval-policy override for AgentCoop-managed runs

## Acceptance criteria

- Codex runs always pass `-c approval_policy=never` under AgentCoop.
- AgentCoop preserves its explicit sandbox behavior on both fresh and resumed Codex turns, despite the CLI flag asymmetry between `exec` and `exec resume`.
- On newer Codex CLI versions, resume prefers JSON mode and uses structured parsing and streaming rather than plain-text scraping.
- Plain-text fallback is attempted only when the CLI clearly indicates that JSON resume is unsupported; unrelated resume failures are surfaced as failures.
- When JSON resume emits `agent_message` or usage events, AgentCoop forwards them through the same structured path used for fresh `invoke()` output; plain-text fallback continues to omit ambiguous combined totals.
- Older Codex CLI versions still work through a tested optimistic fallback path to plain-text resume parsing.
- Top-level JSON `error` events are surfaced as failures only when they are terminal for the turn.
- Retry-related JSON `error` events followed by later terminal success do not cause false failures.
- Plain-text resume parsing is retained only as a compatibility fallback and has stronger edge-case test coverage.

sehkone:
## Summary
- Pin AgentCoop-managed Codex runs to `approval_policy=never` while preserving the adapter's explicit sandbox settings for fresh and resumed turns.
- Prefer `codex exec resume --json` on newer Codex CLIs, route resumed streaming and usage through the structured JSON path, and fall back to the legacy plain-text parser only when the CLI explicitly rejects JSON resume.
- Harden terminal JSON error handling and plain-text resume parsing, and expand compatibility tests around retry-style errors, resume fallback behavior, and marker-like plain-text output.

Closes #277

## Test plan
- [x] `pnpm check`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm dlx markdownlint-cli2 "**/*.md"`
- [x] Verified fresh invoke args include `-c approval_policy=never` alongside `-s danger-full-access`.
- [x] Verified resumed runs prefer `--json`, keep `-c sandbox_mode=danger-full-access`, and forward streamed agent text and usage from JSON events.
- [x] Verified resume falls back to plain text only when the CLI explicitly rejects `--json`, and unrelated JSON resume failures remain terminal errors.
- [x] Verified top-level JSON `error` events only fail the turn when they are terminal, while retry-style errors followed by `turn.completed` remain successful.
- [x] Verified plain-text resume parsing handles marker-like `codex` and `tokens used` lines in assistant output without truncating the response.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -154,6 +154,13 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- AgentCoop-managed Codex runs now pin `approval_policy=never`
+  explicitly instead of inheriting the user's local Codex approval
+  setting. Fresh `codex exec` runs keep using
+  `-s danger-full-access`, while `codex exec resume` now prefers
+  `--json` for structured parsing and streaming on newer CLIs and
+  falls back to the legacy plain-text parser only when the installed
+  CLI explicitly rejects JSON resume.
 - Resuming Stage 8 from the `squashing` substate no longer re-sends the
   full planning prompt.  The handler now checks whether the squash
   already landed (commit count collapsed to 1) and jumps straight to
--- a/src/agent.ts
+++ b/src/agent.ts
@@ -31,7 +31,7 @@ export interface AgentResult {
   usage?: TokenUsage;
 }
 
-export interface AgentStream {
+export interface AgentStream<TResult extends AgentResult = AgentResult> {
   /** Async iterator that yields output chunks as they arrive. */
   [Symbol.asyncIterator](): AsyncIterator<string>;
 
@@ -39,7 +39,7 @@ export interface AgentStream {
    * Resolves when the process exits with the final structured result.
    * Consuming the iterator is optional; `result` always resolves.
    */
-  result: Promise<AgentResult>;
+  result: Promise<TResult>;
 
   /** The underlying child process, exposed for cancellation. */
   child: ChildProcess;
--- a/src/codex-adapter.test.ts
+++ b/src/codex-adapter.test.ts
@@ -3,6 +3,7 @@ import { describe, expect, test, vi } from "vitest";
 import type { AgentResult, AgentStream } from "./agent.js";
 import {
   buildCodexInvokeArgs,
+  buildCodexPlainTextResumeArgs,
   buildCodexResumeArgs,
   CodexStreamTransformer,
   detectCodexError,
@@ -194,7 +195,52 @@ describe("parseCodexJsonl", () => {
     expect(result.status).toBe("success");
   });
 
-  test("detects error events with retry messages", () => {
+  test("treats a terminal top-level error event as failure", () => {
+    const lines = [
+      JSON.stringify({ type: "thread.started", thread_id: "sess-error" }),
+      JSON.stringify({ type: "turn.started" }),
+      JSON.stringify({
+        type: "error",
+        message: "stream error: unexpected status 500",
+      }),
+    ].join("\n");
+
+    const result = parseCodexJsonl(lines);
+    expect(result.status).toBe("error");
+    expect(result.responseText).toBe("stream error: unexpected status 500");
+    expect(result.sessionId).toBe("sess-error");
+  });
+
+  test("ignores retry-style top-level error when a later success completes the turn", () => {
+    const lines = [
+      JSON.stringify({ type: "thread.started", thread_id: "sess-retry-ok" }),
+      JSON.stringify({ type: "turn.started" }),
+      JSON.stringify({
+        type: "error",
+        message: "stream error: unexpected status 400; retrying 1/5",
+      }),
+      JSON.stringify({
+        type: "item.completed",
+        item: { id: "item_0", type: "agent_message", text: "recovered" },
+      }),
+      JSON.stringify({
+        type: "turn.completed",
+        usage: { input_tokens: 20, output_tokens: 5 },
+      }),
+    ].join("\n");
+
+    const result = parseCodexJsonl(lines);
+    expect(result.status).toBe("success");
+    expect(result.responseText).toBe("recovered");
+    expect(result.sessionId).toBe("sess-retry-ok");
+    expect(result.usage).toEqual({
+      inputTokens: 20,
+      outputTokens: 5,
+      cachedInputTokens: 0,
+    });
+  });
+
+  test("later turn.failed overrides an earlier top-level error event", () => {
     const lines = [
       JSON.stringify({ type: "thread.started", thread_id: "sess-retry" }),
       JSON.stringify({ type: "turn.started" }),
@@ -210,7 +256,36 @@ describe("parseCodexJsonl", () => {
 
     const result = parseCodexJsonl(lines);
     expect(result.status).toBe("error");
+    expect(result.responseText).toBe("unexpected status 400");
     expect(result.sessionId).toBe("sess-retry");
+  });
+
+  test("ignores a later top-level error once turn.completed has already ended the turn", () => {
+    const lines = [
+      JSON.stringify({ type: "thread.started", thread_id: "sess-done" }),
+      JSON.stringify({
+        type: "item.completed",
+        item: { id: "item_0", type: "agent_message", text: "finished" },
+      }),
+      JSON.stringify({
+        type: "turn.completed",
+        usage: { input_tokens: 12, output_tokens: 3 },
+      }),
+      JSON.stringify({
+        type: "error",
+        message: "stream error after completion",
+      }),
+    ].join("\n");
+
+    const result = parseCodexJsonl(lines);
+    expect(result.status).toBe("success");
+    expect(result.responseText).toBe("finished");
+    expect(result.sessionId).toBe("sess-done");
+    expect(result.usage).toEqual({
+      inputTokens: 12,
+      outputTokens: 3,
+      cachedInputTokens: 0,
+    });
   });
 });
 
@@ -284,6 +359,79 @@ describe("extractCodexResumeResponse", () => {
     ].join("\n");
 
     expect(extractCodexResumeResponse(text)).toBe("");
+  });
+
+  test("does not treat standalone codex lines inside the response as a new marker", () => {
+    const text = [
+      BANNER,
+      "user",
+      "Show the raw transcript markers",
+      "codex",
+      "First line",
+      "codex",
+      "Still part of the response",
+      "tokens used",
+      "500",
+    ].join("\n");
+
+    expect(extractCodexResumeResponse(text)).toBe(
+      "First line\ncodex\nStill part of the response",
+    );
+  });
+
+  test("does not treat standalone tokens used lines inside the response as the footer", () => {
+    const text = [
+      BANNER,
+      "user",
+      "Show the raw transcript markers",
+      "codex",
+      "First line",
+      "tokens used",
+      "not the footer",
+      "Still part of the response",
+      "tokens used",
+      "500",
+    ].join("\n");
+
+    expect(extractCodexResumeResponse(text)).toBe(
+      "First line\ntokens used\nnot the footer\nStill part of the response",
+    );
+  });
+
+  test("does not strip a footer-like tail when the count is not a plain integer", () => {
+    const text = [
+      BANNER,
+      "user",
+      "Show the raw transcript markers",
+      "codex",
+      "First line",
+      "tokens used",
+      "123abc",
+    ].join("\n");
+
+    expect(extractCodexResumeResponse(text)).toBe(
+      "First line\ntokens used\n123abc",
+    );
+  });
+
+  test("parses bare fallback output that starts with codex", () => {
+    const text = ["codex", "Done.", "tokens used", "50"].join("\n");
+    expect(extractCodexResumeResponse(text)).toBe("Done.");
+  });
+
+  test("uses the known prompt to disambiguate prompt-side codex markers", () => {
+    const prompt = ["Show these markers verbatim:", "codex"].join("\n");
+    const text = [
+      BANNER,
+      "user",
+      ...prompt.split("\n"),
+      "codex",
+      "Actual answer",
+      "tokens used",
+      "120",
+    ].join("\n");
+
+    expect(extractCodexResumeResponse(text, prompt)).toBe("Actual answer");
   });
 });
 
@@ -703,7 +851,15 @@ describe("buildCodexInvokeArgs", () => {
   test("builds invoke args with '-' placeholder for stdin", () => {
     const args = buildCodexInvokeArgs({});
 
-    expect(args).toEqual(["exec", "-s", "danger-full-access", "--json", "-"]);
+    expect(args).toEqual([
+      "exec",
+      "-s",
+      "danger-full-access",
+      "--json",
+      "-c",
+      "approval_policy=never",
+      "-",
+    ]);
   });
 
   test("includes -m when model is specified", () => {
@@ -721,10 +877,10 @@ describe("buildCodexInvokeArgs", () => {
     expect(args).not.toContain("--model");
   });
 
-  test("does not include -a flag (not supported by CLI)", () => {
+  test("uses approval_policy=never instead of a legacy approval flag", () => {
     const args = buildCodexInvokeArgs({});
     expect(args).not.toContain("-a");
-    expect(args).not.toContain("never");
+    expect(args).toContain("approval_policy=never");
   });
 
   test("includes -c reasoning effort when specified", () => {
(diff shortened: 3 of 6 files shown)


sehkone:
**[Reviewer Round 1]**

Overall this looks like a well-structured hardening pass that hits each of the issue's acceptance criteria: approval-policy pinning, JSON resume with optimistic fallback, retry-aware terminal-error handling, and more defensive plain-text parsing are all present with matching tests. Below are concerns I'd want addressed before merging, in roughly decreasing importance.

1. **PR title contains the issue number.** CLAUDE.md says "Do NOT include issue or PR numbers in the title," and recent commits follow that (`Harden Codex resume compatibility`, `Track, display, and check CLI versions`, `Forward hotkeys through confirmMerge wrapper`). The committed commit title is already clean, but the PR title `Harden Codex adapter for newer CLI releases (#277)` will propagate to the squash-merged commit. Drop `(#277)` — `Closes #277` in the body already handles linkage.

2. **`parseCodexInvokeOutput` has an unrelated behavior change that isn't scoped to resume and isn't pinned by a test.** `src/codex-adapter.ts:587-600` now replaces an empty `responseText` with `output.trim() || stderrText.trim() || `codex exited with code ${code}`` whenever JSONL parsed cleanly but exit was non-zero. That affects fresh `invoke()` errors too, and none of the updated assertions pin the new string — e.g. `spawn-agent.test.ts:1703-1717` (`invoke handles invalid JSONL output with non-zero exit`) still just checks `status` and `errorType`. It's arguably a nicer error message, but it's not required by #277. Either localize the fallback to `parseCodexResumeJsonOutput` where it's actually needed, or add an explicit assertion so the new behavior is intentional and test-covered.

3. **`sawStructuredJson` is carried on `AgentResult` via a runtime-only cast.** `CodexParsedResult` (src/codex-adapter.ts:77-79) extends `AgentResult` with the optional flag, but `parseCodexResumeJsonOutput` advertises a plain `AgentResult` return, and `isUnsupportedJsonResumeFailure` reads the flag back with `(result as CodexParsedResult).sawStructuredJson` (src/codex-adapter.ts:680). It works today only because nothing between `parseResult` and the predicate drops extra fields — there's no compile-time help. Prefer threading `CodexParsedResult` through explicitly, or decide the "saw JSON?" question by re-running `hasStructuredCodexJsonEvent` on the captured stdout at predicate time instead of relying on propagated metadata.

4. **`isUnsupportedJsonResumeFailure` matches on phrases that can co-occur without being the actual argument rejection.** The predicate (src/codex-adapter.ts:682-696) only requires `--json` and one of `unexpected|unknown|unrecognized argument/option/flag` to appear *anywhere* in the combined response+stderr — not in the same line or message. The `sawStructuredJson` guard is what keeps e.g. `turn.failed { message: "tool failed: unknown option '--json'" }` from tripping it, and that guard is exactly what test `resume does not fall back when structured JSON output mentions --json` leans on. If a future Codex release ever dumps combined text that mentions `--json` in one sentence and `unknown option` somewhere else (e.g. a toolcall error log), we'd silently swap to plain-text and lose usage data. Consider requiring the two phrases to be within the same line, or anchoring on stderr only — CLI argument rejection is always on stderr anyway.

5. **Nested fallback can spawn up to four children on old CLIs with xhigh effort.** With `reasoningEffort: "xhigh"` on a legacy Codex, the composition in `resume` (src/codex-adapter.ts:827-837) is inner-xhigh→high wrapped by outer JSON→plain-text. Worst path: JSON+xhigh rejects `xhigh` as a variant → JSON+high rejects `--json` → plain-text+xhigh rejects `xhigh` again → plain-text+high succeeds. That's four spawns, with the user seeing error output from each attempt. Once JSON+high establishes that the CLI rejects `--json`, the "xhigh is unsupported" signal is already known and the plain-text fallback can skip xhigh and go straight to `high`. Not a correctness bug, but worth either a comment acknowledging the cost or collapsing reasoning across the two modes.

6. **`withConditionalFallback` isn't safe if the wrapped stream is iterated twice.** `chunks()` (src/codex-adapter.ts:661-667) does `yield* stream` then conditionally `yield* ensureRetryStream()`. On a second iteration both streams are already drained, so the generator silently yields nothing. Today no caller iterates `AgentStream` twice, but `withXhighFallback` is exported and this wrapper is now used by multiple layers. An `iteratedOnce` guard that throws on re-iteration (or a comment spelling out the single-consumer assumption) would make the contract explicit.

7. **Minor surface-area / doc polish.**
   - `buildCodexPlainTextResumeArgs` is exported only to be asserted-on in tests (src/codex-adapter.test.ts:1000-1015). If it's not meant to be public, an `@internal` JSDoc or folding the assertions into a `createCodexAdapter.resume` integration test would keep the surface area smaller.
   - The JSDoc for `extractCodexResumeResponse` (src/codex-adapter.ts:298-326) documents the prompt-aware disambiguation in prose but the signature line isn't updated to mention `expectedPrompt`. Tiny readability nit.

Positives worth calling out:
- The retry-error handling in `parseCodexJsonl` (tracking `pendingErrorMessage` and clearing it on any terminal turn event) is exactly what the issue asks for, and the three test cases (`treats a terminal top-level error event as failure`, `ignores retry-style top-level error when a later success completes the turn`, `ignores a later top-level error once turn.completed has already ended the turn`) cover the shape cleanly.
- The new plain-text parsing via `stripTrailingTokenFooter` + `parseTrailingIntegerLine` + validated banner boundary is meaningfully safer than the old `lastIndexOf` approach, and the adversarial tests (`parses bare fallback output that starts with codex`, `does not strip a footer-like tail when the count is not a plain integer`, `uses the known prompt to disambiguate prompt-side codex markers`) are the ones this file has needed.
- `withConditionalFallback` is a clean generalization of `withXhighFallback`, and the prompt-aware fallback through `parseCodexResumePlainTextOutput` is a nice defensive move against prompt/response marker collision.

Nothing above is "wrong answer" blocking; my biggest two asks are the PR title and scoping (or explicitly pinning) the invoke responseText change.

sehkone:

sehkone:
**[Author Round 1]**

1. Fixed: Updated the PR title to `Harden Codex adapter for newer CLI releases` so the squash title no longer carries `(#277)`. The PR body still matches the branch and still closes `#277`, so I left it unchanged.

2. Fixed: Kept the broader `parseCodexInvokeOutput` fallback behavior, but pinned it explicitly in `spawn-agent.test.ts` by asserting that invalid JSONL plus a non-zero exit preserves the raw output (`garbage output`) in `responseText`. That makes the shared invoke/resume behavior intentional instead of incidental.

3. Fixed: Made `AgentStream` and `spawnAgent` generic and threaded `CodexParsedResult` through the JSON resume path, so `isUnsupportedJsonResumeFailure` now reads `sawStructuredJson` from a typed result instead of a runtime cast.

4. Fixed: Tightened JSON-resume fallback detection to look for actual CLI rejection phrases on the same line as `--json`, and added coverage for the split-line case so unrelated stderr text does not trigger the plain-text fallback.

5. Partially addressed: I added a code comment documenting why transport fallback and reasoning fallback remain independent, including the extra-probe tradeoff on very old CLIs. I did not collapse the two retry layers, because that would add cross-layer state plumbing for a non-correctness optimization and would make the fallback chain harder to reason about.

6. Fixed: Added a single-consumer guard to `withConditionalFallback` so a second iterator request now throws instead of silently yielding nothing, and covered it through `withXhighFallback` in `codex-adapter.test.ts`.

7. Fixed: Added `@internal` JSDoc to `buildCodexPlainTextResumeArgs` and updated the `extractCodexResumeResponse` docs to call out the `expectedPrompt` disambiguation parameter explicitly.

Local checks: `pnpm check`, `pnpm typecheck`, `pnpm test`, and `pnpm dlx markdownlint-cli2 "**/*.md"` all pass.

sehkone:
**[Reviewer Round 2]**

All seven Round 1 items are adequately resolved. Walking through each:

1. **PR title** — Fixed. `gh pr view` shows `Harden Codex adapter for newer CLI releases` with no `(#277)`.

2. **`parseCodexInvokeOutput` fallback pinned** — Fixed. `spawn-agent.test.ts:1754-1762` now asserts `result.responseText === "garbage output"` for the invalid-JSONL-plus-non-zero-exit path. The behavior exercised is the new `fallbackResponseText = result.responseText || output.trim() || …` branch at `src/codex-adapter.ts:593-606` (since `parseCodexJsonl` silently skips malformed lines and returns `status: "success"`, that's the path that executes, not the catch block). Behavior is now intentional.

3. **`sawStructuredJson` threaded via types** — Fixed. `AgentStream<TResult extends AgentResult = AgentResult>` in `src/agent.ts:34`, `spawnAgent<TResult>` in `src/spawn-agent.ts:34`, `parseCodexResumeJsonOutput` now returns `CodexParsedResult` at `src/codex-adapter.ts:626`, and `isUnsupportedJsonResumeFailure(result: CodexParsedResult)` at `src/codex-adapter.ts:727` reads the flag directly with no cast. The `as TResult` casts in `spawn-agent.ts` for the error paths (`cli_not_found`, `execution_error`) are benign: `sawStructuredJson` is optional, and those paths never produce responseText with the `--json` rejection phrases that would trigger a false fallback.

4. **Line-level rejection match** — Fixed. `hasUnsupportedJsonResumeArgumentLine` at `src/codex-adapter.ts:703-725` iterates line-by-line and requires `--json` and a rejection phrase on the **same** line. The new test `resume does not fall back when --json and unknown option are on different lines` at `src/spawn-agent.test.ts:1565-1607` exercises the split-line case (stderr has `For JSON output, pass '--json' to the CLI.\ntool failed: unknown option '--sandbox'`) and asserts `mockSpawn` is called exactly once. Would have tripped the old predicate.

5. **4-spawn worst case** — Partially addressed with a comment (`src/codex-adapter.ts:872-876`), which is the option I offered. I accept the tradeoff: collapsing reasoning state across transport layers adds plumbing for a non-correctness optimization. The comment's phrasing is slightly indirect (it says layers are independent because rejections can be independent, rather than naming the spawn-count cost), but the intent reads through and a future maintainer would understand why the layers are kept separate.

6. **Double-iteration guard** — Fixed. `withConditionalFallback` now throws on the second `[Symbol.asyncIterator]()` call (`src/codex-adapter.ts:689-695`), and the test at `src/codex-adapter.test.ts:1338-1348` covers the contract.

7. **JSDoc polish** — Fixed. `buildCodexPlainTextResumeArgs` has `@internal` (`src/codex-adapter.ts:549-552`), and `extractCodexResumeResponse` now documents `expectedPrompt` (`src/codex-adapter.ts:325-327`).

No new findings. `Closes #277` with all acceptance criteria satisfied is appropriate; the open question about removing the plain-text fallback entirely is correctly left out of scope.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Harden Codex resume compatibility
```

**Body**

```text
Pin unattended Codex runs to approval_policy=never so AgentCoop
does not inherit approval behavior from a user's local Codex
configuration.

Prefer JSON for resumed Codex turns and fall back to the legacy
plain-text path only when the CLI clearly rejects --json. This
keeps newer CLIs on the structured streaming and parsing path
while preserving compatibility with older releases.

Treat top-level JSON error events as terminal only when no later
turn result arrives, and tighten the plain-text fallback parser
around the validated banner structure so marker-like response
lines are not misparsed.

Expand compatibility coverage so resume fallback, usage
forwarding, and retry-style error recovery stay locked down.

Closes #277
```

PR #294, opened 2026-04-25
Cap Stage 9 merge prompt to keep choice lines visible (#293)

### LINKED ISSUE #293: Stage 9 merge prompt overflows terminal when squash body is long, hiding choices
## Problem

After Stage 8 ends with a `SUGGESTED_SINGLE` verdict, Stage 9's merge confirmation prompt (`confirmMerge`) inlines the suggested squash commit **body** into the prompt message. When the body is long, the InputArea grows past the terminal height and the `1 — …` / `2 — …` choice lines are pushed below the visible viewport. Ink renders in-place (no alt-screen, no scrollback), so scrolling up does not reveal the hidden choices either.

## Root cause

1. `askMerge` in `src/pipeline.ts` joins `getSquashMergeHint().body` verbatim into the prompt message (around `tipSections.push(\`${label}\n${hint.body}\`)`).
2. `InputArea` (`src/ui/InputArea.tsx`) uses `flexShrink={0}` on the choices box and has no internal viewport or height cap, so it expands to the full message height.
3. The App root Box pins `height={terminalHeight}` (`src/ui/App.tsx`) and Ink does not use the alt-screen. When InputArea exceeds the terminal height, the AgentPane area first collapses to a negative height, and any remaining overflow (the choice lines at the bottom of InputArea) is pushed off-screen with no way to reveal it.

## Proposal

**Drop the body from the prompt message and surface it via the existing hotkey + PR link only.** This caps the message length at the input stage, so we do not need a scroll viewport or pagination to close the regression.

Concretely, in `askMerge` (`src/pipeline.ts`):

- Do not join `hint.body` into the prompt.
- Replace it with a one-line hint such as "Body: N lines — `[b]` copy / view in PR comment". Keep the existing `[b]` hotkey behavior (clipboard copy).
- Keep `hint.title` since it is a single line, but ellipsize if it would wrap to more than one row.
- When the clipboard is unavailable, show only "View body in PR comment" and keep `hint.prUrl` visible.

Additional safety net:

- Add a cap to `inputAreaHeight` in `src/ui/App.tsx` (e.g. `terminalHeight - statusBar - MIN_PANE_CONTENT*2`) and truncate the message tail with a "…(truncated)" marker when it would exceed the cap. This protects against future prompts that might grow long.

## Scope of changes

- Message assembly in `askMerge` (`src/pipeline.ts`).
- New i18n keys (`src/i18n/{en,ko,messages}.ts`).
- (Optional) Height cap in `src/ui/App.tsx` `inputAreaHeight` and message truncation in `src/ui/InputArea.tsx`.
- Related tests: `src/pipeline.test.ts`, `src/stage-e2e.test.ts`, `src/stage-squash.test.ts`, `src/ui/components.test.tsx`.

## Reproduction

Produce a `SUGGESTED_SINGLE` suggestion comment with a long body (e.g. 60+ lines) and enter Stage 9's merge confirmation prompt. Shrink the terminal to 50 rows or fewer — the `1 — …` choice line is no longer visible and cannot be scrolled to.

sehkone:
## Summary

Stage 9's merge-confirm prompt inlined the suggested squash commit body verbatim. When the body was long the InputArea grew past the terminal height, pushing the `1 — …` / `2 — …` choice lines off the bottom of the viewport. Ink renders in-place without an alt-screen, so the hidden choices could not be scrolled back into view and the prompt appeared frozen.

- `askMerge` (`src/pipeline.ts`) now renders a one-line `Suggested body: N lines` summary instead of inlining the body. The full body stays reachable via the existing `[b]` clipboard hotkey, and the linked PR comment is shown as the fallback.
- When the clipboard is unavailable, the body line shows a `view in the PR comment` hint instead of a copy affordance.
- Suggested titles are ellipsized to 120 chars (and any embedded newlines collapsed) so a pathological title can't wrap to multiple rows and reintroduce the same overflow.
- Added new i18n keys (`pipeline.suggestedSquashBodyLines`, `pipeline.suggestedSquashBodyCopyHint`, `pipeline.suggestedSquashBodyViewInPr`, `input.truncated`) for both en and ko.

As a defensive safety net for any future prompt that grows long:

- `inputAreaHeight` in `src/ui/App.tsx` accepts an optional `terminalHeight` and clamps to `computeInputAreaCap(...)`, reserving the status-bar minimum + `MIN_PANE_CONTENT * 2` rows for the agent panes. A `MIN_INPUT_AREA_ROWS = 2` floor prevents the input from collapsing entirely.
- `InputArea` accepts a `maxRows` prop and truncates tail-first via `truncateMessageToFit`, replacing the dropped tail with a single `…(truncated)` marker so the choice / text-input line is always visible.
- Each rendered prompt line (message rows and choice rows alike) is rendered with `wrap="truncate-end"` so a long single-line title or hint cannot wrap to multiple rendered rows on a narrow terminal. Without this, a single source line could span two or more terminal rows even though `truncateMessageToFit` and `inputAreaHeight` count it as one — `wrap="truncate-end"` keeps the source-line / rendered-row mapping 1:1 (visible `…` marker on truncation).

Closes #293

## Test plan

- [x] `pnpm biome check` is clean for changed files
- [x] `pnpm vitest run` — full suite (2004 tests across 44 files) green
- [x] New regression test: 80-line body produces a `<15`-line prompt with no `body line N` content rendered
- [x] New test: pathologically long title is ellipsized (`Fix x{100,}…`)
- [x] New test: clipboard-unavailable path shows `view in the PR comment` instead of copy affordance
- [x] New test: `InputArea maxRows` truncates a 30-line message to fit a 6-row cap, preserves head + choice lines, drops the tail, renders `…(truncated)` marker
- [x] New tests cover `computeInputAreaCap` floor + reservation math and `truncateMessageToFit` edge cases (no cap, fits, head+marker, marker-only, non-positive budget)
- [x] New regression test: a 200-char single-line message renders as exactly one row at the default 100-col render width (would previously wrap to two rows and push the choice lines off-screen)
- [x] New regression test: a long message line containing a hotkey sentinel still renders as one row (verifies that the outer `<Text wrap="truncate-end">` truncates the composed text including nested `<Text>` children)
- [x] Manual: run a `SUGGESTED_SINGLE` flow with a 60+ line body in a 50-row terminal — the `1 — …` / `2 — …` choice lines remain visible; pressing `[b]` copies the full body

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -109,24 +109,36 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   user can then either let the agent perform the squash (which reruns CI)
   or apply the suggestion via GitHub's "Squash and merge" at merge time
   (which avoids the extra CI cycle).
-- Stage 9 (Done) merge-confirm screen now prints the suggested squash title,
-  body, and PR URL inline when the squash stage finished via the
+- Stage 9 (Done) merge-confirm screen now surfaces the suggested squash
+  title and PR URL inline when the squash stage finished via the
   squash-suggestion comment path, so the user can copy-paste without opening
-  the browser.
+  the browser.  The body is represented by a one-line summary (`Suggested
+  body: N lines`) rather than being inlined verbatim, since a long body
+  otherwise pushed the merge-confirm choice lines off the bottom of the
+  terminal viewport with no way to scroll them back into view.  The full
+  body remains accessible through the `[b]` clipboard hotkey and the linked
+  PR comment.
 - Stage 9 (Done) merge-confirm screen now renders `[t] copy` / `[b] copy`
-  hotkey hints next to the suggested squash title and body when the terminal
-  can write to the system clipboard.  Pressing `t` copies the title and `b`
-  copies the body (each independently, so the values line up with GitHub's
-  separate "Squash and merge" title / body fields).  A small clipboard
-  utility detects the environment and returns an ordered candidate list
-  (`pbcopy` / `wl-copy` / `xclip` on local sessions, OSC 52 first on SSH
-  sessions, OSC 52 as fallback everywhere stdout is a TTY), and the writer
-  tries candidates in order until one succeeds.  When no candidate is
-  reachable, the hints are not rendered at all — the user falls back to
-  drag-select as before, rather than seeing a hint that silently does
-  nothing.  Per-hotkey status (`copy` / `copied` / `copy failed`) is
-  reflected in the label; `copied` auto-reverts after ~1s, `copy failed`
-  persists until the next re-render.
+  hotkey hints next to the suggested squash title and the body summary
+  when the terminal can write to the system clipboard.  Pressing `t` copies
+  the title and `b` copies the body (each independently, so the values line
+  up with GitHub's separate "Squash and merge" title / body fields).  A
+  small clipboard utility detects the environment and returns an ordered
+  candidate list (`pbcopy` / `wl-copy` / `xclip` on local sessions, OSC 52
+  first on SSH sessions, OSC 52 as fallback everywhere stdout is a TTY),
+  and the writer tries candidates in order until one succeeds.  When no
+  candidate is reachable, the hints are not rendered at all — the user
+  falls back to opening the PR comment, rather than seeing a hint that
+  silently does nothing.  Per-hotkey status (`copy` / `copied` / `copy
+  failed`) is reflected in the label; `copied` auto-reverts after ~1s,
+  `copy failed` persists until the next re-render.
+- Stage 9's merge-confirm prompt now caps the height of the InputArea so
+  even an unexpectedly long prompt cannot push the choice / input lines
+  past the bottom of the terminal viewport.  When the message would
+  overflow, the tail is replaced with a single `…(truncated)` marker so
+  the choice lines stay visible.  Pathologically long suggested squash
+  titles (>120 characters or containing embedded newlines) are also
+  ellipsized at the assembly stage.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,7 +138,10 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   overflow, the tail is replaced with a single `…(truncated)` marker so
   the choice lines stay visible.  Pathologically long suggested squash
   titles (>120 characters or containing embedded newlines) are also
-  ellipsized at the assembly stage.
+  ellipsized at the assembly stage.  Each rendered prompt line — message
+  rows and choice rows alike — is rendered with `wrap="truncate-end"` so
+  a long single-line title or hint cannot wrap to multiple terminal rows
+  on a narrow terminal, which would otherwise break the row budget.
 
 ### Changed
 
--- a/docs/pipeline.md
+++ b/docs/pipeline.md
@@ -1569,7 +1569,10 @@ Cleanup
    long, the InputArea also caps its height so the choice /
    text-input line is always visible.  When the message would
    exceed that cap, the tail is replaced with a single
-   `…(truncated)` marker.
+   `…(truncated)` marker.  Each rendered line — message rows and
+   choice rows — also uses `wrap="truncate-end"` so a long
+   single-line title or hint cannot wrap to multiple rendered
+   rows on a narrow terminal, keeping the row budget accurate.
 
 **Agent rebase prompt:**
 
--- a/docs/pipeline.md
+++ b/docs/pipeline.md
@@ -1545,16 +1545,31 @@ Cleanup
      the PR. Each action is individually selectable.
 
    When a squash suggestion is live in a PR comment, the
-   merge-confirm screen also renders `[t] copy` / `[b] copy`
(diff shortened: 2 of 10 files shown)


sehkone:
**[Reviewer Round 1]**

I found one issue that I think should be fixed before merging.

- `src/ui/InputArea.tsx:273`, `src/ui/App.tsx:113`, and `src/pipeline.ts:1250`: the new cap/truncation logic counts only newline-delimited source lines, but Ink wraps `<Text>` by terminal width. That means `inputAreaHeight` and `truncateMessageToFit` can both decide the prompt fits while the rendered rows exceed `maxRows`, and the suggested title can still wrap: `ellipsizeTitle` caps at 120 characters, which is still more than one rendered row on a normal 80-column terminal once `Suggested title:` and the optional `[t]` hint are included. This leaves the defensive safety net incomplete for future long prompts made of wrapped lines, and it does not fully satisfy the issue requirement to ellipsize the title when it would wrap to more than one row. Please make the budget width-aware, for example by using the same wrapped-row approach as `splitIntoRows` in `AgentPane`, or by truncating/no-wrapping the title and message lines before height budgeting. A regression test with a narrow terminal width and a long single-line title/prompt would catch this; the current tests only cover explicit newlines.

The main Stage 9 body change otherwise looks aligned with the issue: the squash body is no longer inlined, `[b]` still copies the full body when available, and the PR comment remains the fallback source.

sehkone:

sehkone:
**[Author Round 1]**

- **Fixed** — width-aware rendering for `InputArea`. Each per-line `<Text>` in `renderMessageLines` and each choice line now uses `wrap="truncate-end"`, so any single source line stays at exactly one rendered row regardless of terminal width. With that 1:1 source-line / rendered-row guarantee, the newline-based counting in `truncateMessageToFit` and `inputAreaHeight` accurately matches what the user sees — addressing the budget-vs-render skew. The 120-char `ellipsizeTitle` cap is still useful as a defensive upstream limit (it bounds the message string passed to `confirmMerge` for non-Ink consumers / log output), but the actual on-screen wrap is now bounded by the renderer.

  I picked the second branch of the suggestion (truncate / no-wrap before height budgeting) over a width-aware `splitIntoRows` budget because it's simpler and keeps the budget naturally accurate without threading `terminalWidth` through `inputAreaHeight`, `truncateMessageToFit`, and the App layout. The visible `…` marker also signals the truncation to the user.

- **Fixed** — added two regression tests for the narrow-terminal / long-single-line case the previous suite missed:
  - `a long single-line message renders as one row, choices stay visible` — feeds a 200-char single-line message into `InputArea` at the default 100-col render width and asserts the rendered frame is exactly 3 rows (1 message + 2 choices), the message row ends with `…`, and the choice rows are present.
  - `a long message line containing a sentinel still renders as one row` — verifies that the outer `<Text wrap="truncate-end">` truncation propagates through nested `<Text>` children where the hotkey hint label is rendered, so a sentinel-bearing long line is also bounded to one rendered row.

Verified locally: `pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm vitest run` (2004 tests across 44 files) all green.

sehkone:
**[Reviewer Round 2]**

No remaining findings from me.

The Round 1 width-budget concern is resolved. `src/ui/InputArea.tsx:179`, `src/ui/InputArea.tsx:235`, and `src/ui/InputArea.tsx:242` now render message and choice source lines with `wrap="truncate-end"`, so the newline-based `truncateMessageToFit` / `inputAreaHeight` budget is no longer invalidated by Ink wrapping a long single source line. The added regressions in `src/ui/components.test.tsx:3083` and `src/ui/components.test.tsx:3112` cover the long-line and sentinel-substitution cases that would have failed before.

I also rechecked the Stage 9 behavior against #293: `src/pipeline.ts:1259` through `src/pipeline.ts:1285` no longer inlines the squash body, keeps `[b]` copying the full body when clipboard support exists, shows the PR-comment fallback when clipboard support is absent, and keeps the PR URL visible when provided. The long-body, long-title, clipboard-available, and clipboard-unavailable tests in `src/pipeline.test.ts` cover the important paths.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Cap Stage 9 merge prompt to keep choices visible
```

**Body**

```text
Stage 9's merge-confirm prompt inlined the suggested squash body
verbatim. A long body grew InputArea past the terminal height and
pushed the `1 — …` / `2 — …` choices off-screen. Ink renders
in-place without an alt-screen, so the hidden choices could not be
scrolled back into view and the prompt appeared frozen.

Render only a one-line `Suggested body: N lines` summary in the
prompt; the full body stays reachable via the existing `[b]`
clipboard hotkey and the linked PR comment. When the clipboard is
unavailable, surface a `view in the PR comment` hint instead of an
inline copy affordance. Ellipsize pathologically long titles so a
title cannot wrap to multiple rows and reintroduce the overflow.

Add a defensive height cap on InputArea: reserve the status bar
plus `MIN_PANE_CONTENT * 2` rows for the agent panes, then truncate
any future overflow tail-first with a `…(truncated)` marker so the
choice / text-input line is always on screen. Render each per-line
`<Text>` with `wrap="truncate-end"` so a single long source line
cannot wrap to multiple rendered rows on a narrow terminal, keeping
the newline-based row budget accurate.

Closes #293
```

PR #305, opened 2026-04-30
Move squash-suggestion comment authoring from agent to code (#304)

### LINKED ISSUE #304: Move squash-suggestion comment authoring from agent to code
## Problem

Stage 8 (squash) currently delegates the authoring of the squash-suggestion PR comment entirely to the agent. The prompt at `src/stage-squash.ts:200-282` instructs the agent to:

- Wrap the suggestion in `<!-- agentcoop:squash-suggestion:start -->` / `<!-- agentcoop:squash-suggestion:end -->` markers.
- Use the `**Title**` / `**Body**` labels.
- Compute fence length dynamically (`max(longest run of backticks in content, 2) + 1`) so commit bodies containing triple-backtick samples don't break the outer fence.
- Find the prior squash-suggestion comment via `gh api ... | select(.body | contains(start_marker))` and either PATCH it or POST a fresh one for idempotent updates.

The code only validates the result after the fact (`hasValidSuggestionBlock` → `parseSquashSuggestionBlock` at `src/stage-squash.ts:377-491`). Any one missing piece — end marker, unterminated fence, missing label, mis-sized fence — fails the validator and the stage returns `outcome: "blocked"` with `requiresArtifact: true`, which leaves the user only "Give instruction" / "Halt" choices.

## Symptom

While shipping aicers/aice-web-next#377, agentcoop produced a squash-suggestion comment that contained the start marker and the labels but was missing both the closing body fence and the end marker. The verdict turn cleanly returned `SUGGESTED_SINGLE`, but the post-verdict validation at `src/stage-squash.ts:881-888` rejected the comment, so the stage flipped to `blocked` and the TUI rendered the `BLOCKED:` prompt.

The agent's first-turn explanation in that run ("Branch commits and PR body unchanged since the prior squash decision; the existing single-commit suggestion comment ... remains accurate. No update needed.") shows the second failure mode of this design: when the agent skips re-writing the comment because it judges the existing one to be still valid, a previously corrupted comment silently propagates into the next run.

## Root cause

Marker placement, fence-length arithmetic, and idempotent-update bookkeeping are deterministic concerns that the agent has to reproduce by hand on every turn. LLM output is not a reliable substrate for this kind of mechanical formatting — one omitted line shows up as an opaque blocked outcome several turns later, with no targeted feedback to the agent. The strict fail-closed policy at `src/stage-squash.ts:484-488` is correct (Stage 9 reads the same block and a malformed block would render the wrong text in the inline preview), but the system relies on the agent to never make a formatting mistake.

## Proposed change

Split responsibilities so the agent only produces the natural-language part:

1. **Add a `buildSquashSuggestionComment(suggestion: SquashSuggestion): string` helper** in `src/stage-squash.ts` that takes a `{ title, body }` pair and returns the canonical marker block — including markers, labels, and fences sized via the existing `max(longest backtick run, 2) + 1` rule, computed independently for title and body. Use the existing `SQUASH_SUGGESTION_START_MARKER` / `SQUASH_SUGGESTION_END_MARKER` constants.

2. **Widen `findLatestCommentWithMarker` to return the comment id alongside the body.** [src/pr-comments.ts:125-142](src/pr-comments.ts:125) currently returns `string | undefined`, which is enough for the existing read-only callers but not enough to PATCH. Change the return type to `{ id: number; body: string } | undefined` (or add a sibling helper if churn is a concern) and update the two existing callers — verdict-fallback at `src/stage-squash.ts:837` and resume-from-`awaiting_user_choice` at `src/stage-squash.ts:739` — to read `.body`. The widened helper is what `postOrUpdateSquashSuggestion` needs to decide between PATCH and POST. Decide explicitly whether the existing injectable `SquashStageOptions.findSuggestionCommentBody` option ([src/stage-squash.ts:99-104](src/stage-squash.ts:99)) keeps its read-only `string`-returning shape (and a separate write-side option is added) or is widened to the richer return type so reads and writes share one entry point.

3. **Add a `postOrUpdateSquashSuggestion(...)` helper** that calls the widened `findLatestCommentWithMarker`, then PATCHes via `gh api ... --method PATCH /repos/{owner}/{repo}/issues/comments/{id}` when a prior comment exists or POSTs via `gh pr comment` otherwise. This is the single entry point for writing the comment. Follow the existing convention in this file — `getCiStatus`, `findPrNumber`, and similar helpers are exposed as injectable options on `SquashStageOptions` rather than calling `execFileSync` directly, which keeps the unit tests in `src/stage-squash.test.ts` from needing a real `gh` binary. The new helper should follow the same pattern.

4. **Simplify the squash work prompt** at `src/stage-squash.ts:200-282`. Remove the marker/fence/idempotent-update instructions for the SUGGESTED_SINGLE branch. Ask the agent to reply with title and body in a small machine-parseable envelope (e.g. `<<<TITLE>>>...<<</TITLE>>>` / `<<<BODY>>>...<<</BODY>>>`) and *not* call `gh` itself for the comment. The SQUASHED_MULTI branch (history rewrite + force-push) is unaffected.

5. **Decide where the envelope is parsed and the comment is posted, relative to the verdict turn.** Today Stage 8 is a two-turn dance — work prompt then verdict prompt — and the verdict's `SUGGESTED_SINGLE` keyword has implicit "the agent posted the suggestion" semantics. Once authoring moves into the code, that semantic drift needs an explicit resolution. Two viable shapes:

   - **Keep the verdict turn, infer the branch from the work response.** Parse the envelope from `squashResult.responseText`. If present → SUGGESTED_SINGLE branch: the code posts the comment immediately, and the verdict turn is either skipped or downgraded to a confirmation. If absent → fall through to the existing verdict / clarification flow, which decides between SQUASHED_MULTI, BLOCKED, and the deterministic fallback chain — envelope absence on its own is not a SQUASHED_MULTI signal (the agent could equally be in BLOCKED, or have produced a malformed response). Smallest change to the surrounding state machine; preserves the clarification / deterministic-fallback chain in [src/stage-squash.ts:507-528](src/stage-squash.ts:507) and [src/stage-squash.ts:824-855](src/stage-squash.ts:824).
   - **Collapse work + verdict into one turn.** The work prompt asks for the decision *and* (when SUGGESTED_SINGLE) the envelope in the same response. Cleaner end state but a larger behavioural change with knock-on effects on `resolveVerdict`, the clarification follow-up, and the deterministic fallback chain.

   The first option is preferred for this issue. The second is fine but should be scoped separately if pursued.

6. **Parse the agent's title/body envelope, then call `postOrUpdateSquashSuggestion`**. If the envelope is missing or malformed when the agent appears to have taken the SUGGESTED_SINGLE branch, fall through to the existing clarification / blocked path with a precise error message.

7. **Demote `hasValidSuggestionBlock` to an assertion** on the path immediately after the code wrote the comment itself. It still serves as defense-in-depth and remains the gate when we resume from `awaiting_user_choice` against a comment we did not just author.

## Tests

The existing `parseSquashSuggestionBlock` test suite already covers most of the negative cases that come to mind. Audit it first and only add what is missing — the screenshot incident specifically combined "missing end marker" with "unterminated body fence", so a test that exercises both at once is worth adding regardless.

The high-value new tests are:

- **Round-trip `buildSquashSuggestionComment` → `parseSquashSuggestionBlock`** for plain bodies, bodies containing 3-backtick fences, bodies containing 5-backtick fences, and bodies with HTML comments. Round-trip is the single strongest guarantee the formatter and parser stay in lock-step.
- **`postOrUpdateSquashSuggestion` against a stubbed `gh` adapter**: PATCHes via the right comment id when a prior comment exists, POSTs otherwise.
- **A light prompt assertion** that the SUGGESTED_SINGLE work prompt no longer references `squash-suggestion:start` / `gh pr comment` / fence-length math. Use negative `not.toContain` checks; avoid asserting on positive prose so the test does not break on copy edits.
- **Malformed envelope handling**: when the agent takes the SUGGESTED_SINGLE branch but its response cannot be parsed into title/body, the stage funnels into the existing clarification / blocked path with an actionable message. Cover at least missing close tag (`<<<TITLE>>>...` with no `<<</TITLE>>>`), empty title, and empty body.

## Out of scope (follow-up)

Self-healing for already-corrupted comments on existing PRs (run a one-shot follow-up to ask the agent to repair the comment when validation fails but the start marker is present). Track separately so it can ship after the main change has soaked.

## Why now

A user run on aicers/aice-web-next#377 was blocked by exactly this failure mode and required a manual `gh api PATCH` to repair the comment before the pipeline could resume. The risk of repeat failure on other PRs is non-trivial because the formatting rules the agent has to reproduce are intricate (CommonMark fence-length math in particular).

sehkone:
## Summary

Stage 8's SUGGESTED_SINGLE branch used to delegate authoring of the marker-delimited squash-suggestion PR comment to the agent — marker placement, CommonMark fence sizing, and idempotent PATCH/POST bookkeeping all had to be reproduced on every turn. A single missed line would silently flip the stage to `blocked` after the verdict cleanly returned `SUGGESTED_SINGLE`, as observed on aicers/aice-web-next#377 (missing closing body fence and end marker, repaired by hand).

This PR splits responsibilities so the agent only produces the natural-language title and body inside a small `<<<TITLE>>> / <<<BODY>>>` envelope, and agentcoop deterministically renders the marker block and PATCH/POSTs it idempotently.

Key changes:

- `buildSquashSuggestionComment(suggestion)` renders the canonical marker block, sizing title/body fences independently as `max(longest backtick run, 2) + 1` (min 3).
- `parseSquashSuggestionBlock` anchors both markers to whole lines and requires the end-marker line at or after the body's closing fence.  This prevents a literal end marker embedded inside the body fenced block from truncating the parse — the body fence absorbs that occurrence as content, and only a free-standing end-marker line counts as the delimiter.  Round-trip tests cover bodies that mention either marker as a literal.
- `parseSquashEnvelope(text)` extracts the agent's reply, distinguishing three outcomes:
  - **`missing`** — no `<<<TITLE>>>` tag on its own line.  Prose that mentions the tag names mid-sentence or in backticks (e.g. a multi-commit reply explaining why the envelope was not used) classifies here and falls through to the verdict flow.
  - **`malformed`** — envelope intent declared (TITLE_OPEN on its own line) but the structure is broken: a missing close tag, an absent body section, an empty title, or an empty / whitespace-only body.  Routes into the focused clarification turn so a recoverable formatting mistake — including the dropped close tag that motivated this issue — gets one more attempt before blocking.
  - **`ok`** — well-formed envelope with non-empty content.
- `postOrUpdateSquashSuggestion(...)` is the single write entry point: PATCH the prior comment by id when one with the start marker exists, otherwise POST. Backed by injectable `findLatest`/`patch`/`post` deps for unit-testability.  Errors from the lookup propagate to the caller — they are NOT swallowed into 'no prior comment'.  A transient auth / network / rate-limit failure during lookup would otherwise be indistinguishable from 'no matching comment' and would turn an idempotent PATCH into a duplicate POST.  The handler converts the thrown error into a `blocked` outcome instead.
- `findLatestCommentWithMarker` now returns `{ id, body }` so PATCH callers don't need a second lookup; the two read-only callers destructure `.body`.  Errors from the underlying `gh api` call propagate to the caller.  `getSquashMergeHint` in `index.ts` wraps with `try`/`catch` to silently degrade — the hint is purely cosmetic.  The squash handler's read-side `findSuggestionCommentBody` adapter no longer silently degrades: errors propagate to the caller, and the `awaiting_user_choice` resume path now blocks on a lookup failure (leaving `squashSubStep` at `awaiting_user_choice`) instead of falling through to a fresh planning run that could re-invoke the agent and re-author the suggestion or change the branch decision after the user had already been asked about one.
- New `patchPrComment` helper in `src/pr-comments.ts` sends the body via stdin as a JSON request payload so multi-line content with fences / Markdown-active characters survives verbatim.
- The SUGGESTED_SINGLE work prompt is simplified — no more marker, fence-length math, or `gh` invocation instructions for the agent. The SQUASHED_MULTI branch is unchanged.
- `createSquashStageHandler` runs the envelope-driven shortcut **before** the verdict turn:
  - Well-formed envelope → code authors the comment, asserts it round-trips through the parser, posts it (lookup failure → `blocked` rather than duplicate POST), emits a `pipeline:verdict` `SUGGESTED_SINGLE` event, and proceeds to the user-choice flow (verdict turn skipped).
  - Malformed envelope (intent declared, structure broken) → send one focused **clarification turn** asking for either a valid envelope or a `SQUASHED_MULTI` / `BLOCKED` keyword. Re-parse: ok envelope → author + post; SQUASHED_MULTI → CI poll; BLOCKED → blocked; anything else (still malformed, missing, or bare SUGGESTED_SINGLE) → blocked with both responses surfaced. This preserves the recoverable-mistake path the verdict-clarification round already provides.
  - Envelope absent → fall through to the existing verdict / clarification / deterministic-fallback chain (SQUASHED_MULTI and BLOCKED keep their behaviour; envelope absence is not a SUGGESTED_SINGLE signal).
- **No stale-comment promotion to SUGGESTED_SINGLE.**  A SUGGESTED_SINGLE outcome must be backed by a current envelope from this run.  When the verdict resolves to SUGGESTED_SINGLE without an envelope, the same focused envelope clarification turn fires (asking for either a valid envelope or SQUASHED_MULTI / BLOCKED) — agentcoop does not consume a historical squash-suggestion comment as evidence, which would re-introduce the stale-suggestion propagation problem issue #304 was meant to fix.  The deterministic fallback chain similarly drops the "valid marker block on PR → SUGGESTED_SINGLE" rule; without an envelope from this run, the only deterministic signals are commit-count collapse (SQUASHED_MULTI) or BLOCKED.  The fallback chain's PR-merged guard is preserved so a concurrent merge surfaces as `alreadyMerged` rather than `BLOCKED`.
- `hasValidSuggestionBlock` is demoted to defense-in-depth: an assertion on the just-authored path and the gate when resuming from `awaiting_user_choice` against a comment we did not just write.
- `docs/pipeline.md` and `CHANGELOG.md` updated to match.

Per the issue, option 1 ('keep the verdict turn, infer the branch from the work response') was chosen — the second option (collapsing work + verdict into one turn) is left out of scope.

Closes #304

## Test plan

- [x] \`pnpm tsc --noEmit\` passes
- [x] \`pnpm biome check\` passes
- [x] \`pnpm vitest run\` — full suite green (2056 tests)
- [x] Round-trip \`buildSquashSuggestionComment\` ↔ \`parseSquashSuggestionBlock\` for: plain body, triple-backtick body, five-backtick body, HTML comments in body, inline backticks in title, body containing the start marker as a literal string, body containing the **end marker** as a literal string on its own line, body containing **both** markers as literal strings
- [x] \`parseSquashEnvelope\` classification: ok; missing (no tags / inline backtick-quoted prose / inline mid-sentence prose); malformed for missing TITLE close tag / missing BODY close tag / missing BODY section / empty title / empty body / **whitespace-only body**
- [x] \`postOrUpdateSquashSuggestion\`: PATCHes when prior comment has an id, POSTs when no prior comment, falls back to POST when prior comment lacks an id, propagates errors from \`findLatest\` and does NOT call \`post\` on lookup failure
- [x] \`findLatestCommentWithMarker\`: propagates errors from the underlying \`gh\` call (no longer swallowed)
- [x] Combined regression test: \`parseSquashSuggestionBlock\` rejects a block missing **both** the closing body fence and the end marker (the screenshot incident)
- [x] Negative prompt assertions: SUGGESTED_SINGLE prompt no longer mentions \`squash-suggestion:start\` / \`squash-suggestion:end\` / \`gh pr comment\` / \`gh api repos\` / \`--method PATCH\` / \`fence_len\` / 'longest run of backticks'
- [x] Stage-handler end-to-end: envelope present → comment authored, verdict turn skipped, user-choice runs; envelope present + user picks agent → follow-up resumes the planning session; envelope present → \`pipeline:verdict\` \`SUGGESTED_SINGLE\` event emitted; malformed envelope (empty title or body) → clarification turn sent, then ok envelope on retry → comment authored / SQUASHED_MULTI on retry → CI poll / BLOCKED on retry → blocked / still unrecoverable on retry → blocked; **missing TITLE close tag** → clarification asked, recovers on valid retry; **missing BODY close tag** → clarification asked, SQUASHED_MULTI retry routes into CI poll; envelope absent → falls through to existing verdict flow; multi-commit reply mentioning tag names inline → falls through (no shortcut, no block); **postSuggestionComment lookup throw** → blocked with no duplicate POST
- [x] **Stale-comment fix:** verdict SUGGESTED_SINGLE without envelope + stale historical PR comment → clarification fires (stale comment NOT consumed); clarification retry yielding a valid envelope → comment authored from FRESH content (not stale); deterministic fallback chain with stale marker block + count unchanged → BLOCKED (not SUGGESTED_SINGLE); \`pipeline:verdict\` emits BLOCKED (not SUGGESTED_SINGLE) when only a stale block is present
- [x] **awaiting_user_choice resume:** lookup failure → blocked with detail surfaced, no fresh planning run, no \`chooseSquashApplyMode\` call, sub-step preserved at \`awaiting_user_choice\`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,25 @@ This file documents recent notable changes to this project. The format of this
 file is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and
 this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- Stage 8 (Squash) no longer asks the agent to author the
+  marker-delimited squash-suggestion PR comment.  The agent now
+  drafts the title and body inside a `<<<TITLE>>>` / `<<<BODY>>>`
+  envelope, and agentcoop builds the marker block (with correct
+  fence sizing) and PATCH/POSTs the comment idempotently.  The
+  formatter and parser are pinned in lock-step by a round-trip
+  unit test.  This eliminates a class of malformed-comment
+  failures where a missing end marker or unterminated fence in the
+  agent-authored comment would silently flip the stage to
+  `blocked` after the verdict cleanly returned `SUGGESTED_SINGLE`.
+- `findLatestCommentWithMarker` now returns `{ id, body }` instead
+  of just the body so callers that need to PATCH an existing
+  comment can do so without a second lookup.  Read-only callers
+  destructure `.body`.
+
 ## [0.2.0] - 2026-04-29
 
 ### Added
@@ -263,5 +282,6 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Initial public release of AgentCoop.
 
+[Unreleased]: https://github.com/aicers/agentcoop/compare/0.2.0...HEAD
 [0.2.0]: https://github.com/aicers/agentcoop/compare/0.1.0...0.2.0
 [0.1.0]: https://github.com/aicers/agentcoop/tree/0.1.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,19 +18,33 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   failures where a missing end marker or unterminated fence in the
   agent-authored comment would silently flip the stage to
   `blocked` after the verdict cleanly returned `SUGGESTED_SINGLE`.
-  Envelope detection is strict — all four tags must appear on
-  their own lines, in order — so a multi-commit reply that merely
-  mentions the tag names in prose still falls through to the
-  existing verdict flow instead of pre-empting it with a
-  malformed-envelope block.  When the shape is detected but
-  content is empty, the stage sends one focused clarification turn
-  asking for either a valid envelope or a SQUASHED_MULTI / BLOCKED
-  keyword before blocking, preserving the recoverable-mistake path
-  that the verdict-clarification round already provides.
+  Envelope detection keys off a `<<<TITLE>>>` open tag on its own
+  line — prose that merely mentions the tag names mid-sentence or
+  in backticks never produces a tag on a line by itself, so a
+  multi-commit reply that quotes the tags still falls through to
+  the existing verdict flow instead of pre-empting it with a
+  malformed-envelope block.  Once envelope intent is declared, any
+  structural break (a missing close tag, an absent body section,
+  empty title or body content) classifies as malformed and routes
+  into a focused clarification turn that asks for either a valid
+  envelope or a SQUASHED_MULTI / BLOCKED keyword.  This preserves
+  the recoverable-mistake path the verdict-clarification round
+  already provides — a dropped close tag is exactly the kind of
+  formatting-only mistake the clarification turn was meant to
+  repair.
 - `findLatestCommentWithMarker` now returns `{ id, body }` instead
   of just the body so callers that need to PATCH an existing
   comment can do so without a second lookup.  Read-only callers
-  destructure `.body`.
+  destructure `.body`.  Errors from the underlying `gh api` call
+  propagate to the caller — they are no longer swallowed into
+  `undefined`.  Write-side callers like `postOrUpdateSquashSuggestion`
+  must distinguish "no matching comment" from "lookup failed" so a
+  transient auth / network / rate-limit failure does not turn an
+  idempotent PATCH into a duplicate POST; the squash stage converts
+  any thrown lookup error into a `blocked` outcome instead.
+  Read-only callers (`getSquashMergeHint` in `index.ts`, the
+  in-handler `findSuggestionCommentBody` adapter) wrap the call in
+  `try`/`catch` to silently degrade to "no matching comment".
 
 ## [0.2.0] - 2026-04-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,13 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   legitimately documents the envelope contract (plausible for
   issue #304 itself) is not truncated at an in-body literal close
   marker; in-body literal tags are absorbed as content and only
-  the final own-line `<<</BODY>>>` terminates the envelope.
+  the final own-line `<<</BODY>>>` terminates the envelope.  The
+  structural close tag must additionally be the last non-blank
+  line of the response, so an envelope where the agent forgot the
+  real close tag while including an example `<<</BODY>>>` line in
+  body content classifies as malformed (and routes into the
+  clarification turn) instead of being silently accepted with a
+  truncated body.
 - `findLatestCommentWithMarker` now returns `{ id, body }` instead
   of just the body so callers that need to PATCH an existing
   comment can do so without a second lookup.  Read-only callers
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,15 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   failures where a missing end marker or unterminated fence in the
   agent-authored comment would silently flip the stage to
   `blocked` after the verdict cleanly returned `SUGGESTED_SINGLE`.
+  Envelope detection is strict — all four tags must appear on
+  their own lines, in order — so a multi-commit reply that merely
+  mentions the tag names in prose still falls through to the
+  existing verdict flow instead of pre-empting it with a
+  malformed-envelope block.  When the shape is detected but
+  content is empty, the stage sends one focused clarification turn
+  asking for either a valid envelope or a SQUASHED_MULTI / BLOCKED
+  keyword before blocking, preserving the recoverable-mistake path
+  that the verdict-clarification round already provides.
 - `findLatestCommentWithMarker` now returns `{ id, body }` instead
   of just the body so callers that need to PATCH an existing
   comment can do so without a second lookup.  Read-only callers
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,29 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   Read-only callers (`getSquashMergeHint` in `index.ts`, the
   in-handler `findSuggestionCommentBody` adapter) wrap the call in
   `try`/`catch` to silently degrade to "no matching comment".
+- `parseSquashSuggestionBlock` now anchors both markers to whole
+  lines and requires the end-marker line to appear at or after the
+  body's closing fence.  This prevents a literal end marker
+  embedded inside the body fenced block from truncating the parse:
+  the body fence absorbs that occurrence as content, and only a
+  free-standing end-marker line — emitted by the formatter after
+  the body fence closes — counts as the delimiter.  A round-trip
+  test pins the property for body content that mentions the end
+  marker.
+- The post-verdict path no longer promotes a historical
+  squash-suggestion comment on the PR to a SUGGESTED_SINGLE
+  outcome.  A SUGGESTED_SINGLE outcome must be backed by a current
+  `<<<TITLE>>> / <<<BODY>>>` envelope from this run.  When the
+  verdict resolves to SUGGESTED_SINGLE without an envelope, the
+  same focused clarification turn used for the malformed case
+  fires (asking for either a valid envelope or SQUASHED_MULTI /
+  BLOCKED).  The deterministic fallback chain similarly drops the
+  "valid marker block on PR → SUGGESTED_SINGLE" rule; without an
+  envelope from this run, the only deterministic signals are
+  commit-count collapse (SQUASHED_MULTI) or BLOCKED.  This closes
+  the stale-suggestion propagation problem issue #304 was meant to
+  fix, which an earlier revision still permitted via the verdict
+  path.
 
 ## [0.2.0] - 2026-04-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,9 +25,10 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   the existing verdict flow instead of pre-empting it with a
   malformed-envelope block.  Once envelope intent is declared, any
   structural break (a missing close tag, an absent body section,
-  empty title or body content) classifies as malformed and routes
-  into a focused clarification turn that asks for either a valid
-  envelope or a SQUASHED_MULTI / BLOCKED keyword.  This preserves
+  empty title, or empty / whitespace-only body content) classifies
+  as malformed and routes into a focused clarification turn that
+  asks for either a valid envelope or a SQUASHED_MULTI / BLOCKED
+  keyword.  This preserves
   the recoverable-mistake path the verdict-clarification round
   already provides — a dropped close tag is exactly the kind of
   formatting-only mistake the clarification turn was meant to
@@ -42,9 +43,17 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   transient auth / network / rate-limit failure does not turn an
   idempotent PATCH into a duplicate POST; the squash stage converts
   any thrown lookup error into a `blocked` outcome instead.
-  Read-only callers (`getSquashMergeHint` in `index.ts`, the
-  in-handler `findSuggestionCommentBody` adapter) wrap the call in
-  `try`/`catch` to silently degrade to "no matching comment".
+  `getSquashMergeHint` in `index.ts` wraps the call in `try`/`catch`
+  to silently degrade — the hint is purely cosmetic.  The squash
+  handler's read-side `findSuggestionCommentBody` adapter no longer
+  silently degrades: errors propagate to the caller, and the
+  `awaiting_user_choice` resume path now blocks on a lookup failure
+  instead of falling through to a fresh planning run.  Falling
+  through on this stateful path could re-invoke the agent and
+  re-author the suggestion or change the branch decision after the
+  user had already been presented with one; blocking leaves
+  `squashSubStep` at `awaiting_user_choice` so a retry once `gh`
+  recovers re-presents the existing choice.
 - `parseSquashSuggestionBlock` now anchors both markers to whole
   lines and requires the end-marker line to appear at or after the
   body's closing fence.  This prevents a literal end marker
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,12 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   the recoverable-mistake path the verdict-clarification round
   already provides — a dropped close tag is exactly the kind of
   formatting-only mistake the clarification turn was meant to
-  repair.
+  repair.  The body's structural close tag is anchored to the LAST
+  own-line `<<</BODY>>>` after BODY_OPEN, so a body that
+  legitimately documents the envelope contract (plausible for
+  issue #304 itself) is not truncated at an in-body literal close
+  marker; in-body literal tags are absorbed as content and only
+  the final own-line `<<</BODY>>>` terminates the envelope.
 - `findLatestCommentWithMarker` now returns `{ id, body }` instead
   of just the body so callers that need to PATCH an existing
   comment can do so without a second lookup.  Read-only callers
--- a/docs/pipeline.md
+++ b/docs/pipeline.md
@@ -1355,7 +1355,10 @@ and classifies the response as one of three outcomes:
   BODY_OPEN so a body that legitimately documents the envelope
   contract (plausible for issue #304 itself) is not truncated at
   an in-body literal close tag — only the final own-line
-  `<<</BODY>>>` terminates the envelope,
+  `<<</BODY>>>` terminates the envelope, and that close tag must
+  be the last non-blank line of the response so a forgotten real
+  close tag (with an in-body example present) classifies as
+  malformed instead of silently truncating the body,
   then `postOrUpdateSquashSuggestion` PATCHes the prior comment by
   id (when one exists) or POSTs a fresh one.  A transient lookup
   failure inside the write helper (auth, network, rate-limit) is
--- a/docs/pipeline.md
+++ b/docs/pipeline.md
@@ -1351,6 +1351,11 @@ and classifies the response as one of three outcomes:
   non-empty content) → agentcoop calls
   `buildSquashSuggestionComment` to render the canonical marker
   block (asserts round-trip parseability as defense-in-depth),
+  BODY_CLOSE is anchored to the LAST own-line `<<</BODY>>>` after
+  BODY_OPEN so a body that legitimately documents the envelope
+  contract (plausible for issue #304 itself) is not truncated at
+  an in-body literal close tag — only the final own-line
+  `<<</BODY>>>` terminates the envelope,
   then `postOrUpdateSquashSuggestion` PATCHes the prior comment by
   id (when one exists) or POSTs a fresh one.  A transient lookup
   failure inside the write helper (auth, network, rate-limit) is
--- a/docs/pipeline.md
+++ b/docs/pipeline.md
@@ -1338,7 +1338,12 @@ cleanup stays in Stage 9.
 
 **Envelope-driven SUGGESTED_SINGLE shortcut:** Before running the
 verdict turn, the handler scans the work response for the
-`<<<TITLE>>>` / `<<<BODY>>>` envelope.  Three outcomes:
+`<<<TITLE>>>` / `<<<BODY>>>` envelope.  Detection is **strict** —
+all four tags must appear on their own lines, in the order
+TITLE_OPEN → TITLE_CLOSE → BODY_OPEN → BODY_CLOSE — so that prose
+which merely mentions the tag names (e.g. backtick-quoted in a
+multi-commit reply explaining why the envelope was not used) is
+not mistaken for an envelope attempt.  Three outcomes:
 
 - **Well-formed envelope** → agentcoop calls
   `buildSquashSuggestionComment` to render the canonical marker
@@ -1349,12 +1354,22 @@ verdict turn, the handler scans the work response for the
   handler proceeds directly to the user-choice path.  A
   `pipeline:verdict` event with keyword `SUGGESTED_SINGLE` is
   emitted so telemetry consumers see the verdict.
-- **Malformed envelope** (at least one of the four tags present,
-  but the envelope cannot be parsed — missing close tag, empty
-  title, empty body, etc.) → fail closed with `blocked` and an
-  actionable error message.  The agent has signalled intent to
-  take the SUGGESTED_SINGLE branch; falling through to the verdict
-  turn would silently mask the formatting failure.
+- **Malformed envelope** (all four tags on their own lines and in
+  order, but content is empty — empty title or empty body) → send
+  one focused **clarification turn** asking the agent to reply
+  with either a valid envelope or a `SQUASHED_MULTI` /
+  `BLOCKED` keyword (no other commentary).  Re-parse the retry:
+  - Valid envelope → author and post the comment, proceed to the
+    user-choice path.
+  - `SQUASHED_MULTI` keyword → run the CI poll path.
+  - `BLOCKED` keyword → existing blocked flow.
+  - Anything else (still malformed, missing, or a bare
+    `SUGGESTED_SINGLE` keyword without an envelope) → fail closed
+    with `blocked` and surface both responses for diagnostics.
+  The clarification preserves the recoverable-mistake path the
+  verdict-clarification round already provides, so a single
+  formatting slip does not dump the user into "Give instruction /
+  Halt" with no context.
 - **Envelope absent** → fall through to the existing verdict /
   clarification chain.  Envelope absence on its own is not a
   SUGGESTED_SINGLE signal: the agent could be in the
@@ -1422,7 +1437,11 @@ the fallback chain.
 planning ──┬── (envelope ok)         → awaiting_user_choice
            │       ├── (agent)  → squashing → ci_poll → done
            │       └── (github) → applied_via_github (stage done)
-           ├── (envelope malformed)  → blocked
+           ├── (envelope malformed)  → clarify
+           │       ├── (envelope ok)        → awaiting_user_choice (as above)
+           │       ├── (SQUASHED_MULTI)     → ci_poll → done
+           │       ├── (BLOCKED)            → blocked
+           │       └── (still unrecoverable) → blocked
            └── (envelope absent)     → verdict
                 ├── (SQUASHED_MULTI)   → ci_poll → done
                 ├── (SUGGESTED_SINGLE) → awaiting_user_choice (as above)
--- a/docs/pipeline.md
+++ b/docs/pipeline.md
@@ -1338,27 +1338,34 @@ cleanup stays in Stage 9.
 
 **Envelope-driven SUGGESTED_SINGLE shortcut:** Before running the
 verdict turn, the handler scans the work response for the
-`<<<TITLE>>>` / `<<<BODY>>>` envelope.  Detection is **strict** —
-all four tags must appear on their own lines, in the order
-TITLE_OPEN → TITLE_CLOSE → BODY_OPEN → BODY_CLOSE — so that prose
-which merely mentions the tag names (e.g. backtick-quoted in a
-multi-commit reply explaining why the envelope was not used) is
-not mistaken for an envelope attempt.  Three outcomes:
+`<<<TITLE>>>` / `<<<BODY>>>` envelope.  Detection keys off a
+`<<<TITLE>>>` open tag on its own line — prose that merely names
+the tags mid-sentence or in backticks (e.g. a multi-commit reply
+explaining why the envelope was not used) never produces a tag on
+a line by itself, so it does not register as envelope intent.
+Once envelope intent is declared, the parser walks the structure
+and classifies the response as one of three outcomes:
 
-- **Well-formed envelope** → agentcoop calls
+- **Well-formed envelope** (TITLE_OPEN → TITLE_CLOSE → BODY_OPEN
+  → BODY_CLOSE all present on their own lines, in order, with
+  non-empty content) → agentcoop calls
   `buildSquashSuggestionComment` to render the canonical marker
   block (asserts round-trip parseability as defense-in-depth),
   then `postOrUpdateSquashSuggestion` PATCHes the prior comment by
-  id (when one exists) or POSTs a fresh one.  The verdict turn is
-  skipped — the envelope is the SUGGESTED_SINGLE signal — and the
-  handler proceeds directly to the user-choice path.  A
-  `pipeline:verdict` event with keyword `SUGGESTED_SINGLE` is
-  emitted so telemetry consumers see the verdict.
-- **Malformed envelope** (all four tags on their own lines and in
-  order, but content is empty — empty title or empty body) → send
-  one focused **clarification turn** asking the agent to reply
-  with either a valid envelope or a `SQUASHED_MULTI` /
-  `BLOCKED` keyword (no other commentary).  Re-parse the retry:
+  id (when one exists) or POSTs a fresh one.  A transient lookup
+  failure inside the write helper (auth, network, rate-limit) is
+  surfaced as a `blocked` outcome rather than POSTing a duplicate
+  suggestion comment.  The verdict turn is skipped — the envelope
+  is the SUGGESTED_SINGLE signal — and the handler proceeds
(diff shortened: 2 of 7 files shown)


sehkone:
**[Reviewer Round 1]**

1. `src/stage-squash.ts:554-606` and `src/stage-squash.ts:977-983` make the new shortcut too eager: any stray `<<<TITLE>>>` / `<<<BODY>>>` mention is treated as an attempted single-commit envelope. A perfectly valid multi-commit response like “I took the multi-commit path, so I did not use the `<<<TITLE>>>` / `<<<BODY>>>` envelope” now returns `blocked` before the verdict turn ever runs, because `parseSquashEnvelope()` sees one tag and classifies the whole response as `malformed`. That is a real false-negative path for `SQUASHED_MULTI`, and I do not see a regression test covering it. I think the shortcut needs a stricter shape check (for example, require all four tags in order on their own lines before treating the response as an envelope attempt) plus a test that proves plain prose mentioning the tags still falls through to the existing verdict flow.

2. `src/stage-squash.ts:977-983` also drops the clarification behavior that issue #304 explicitly tried to preserve. The preferred shape in the issue was “keep the verdict turn, infer the branch from the work response,” so a malformed single-commit reply can still go through the existing clarification/blocked path with targeted feedback. Here, a missing close tag or empty field hard-blocks immediately and sends the user back to “Give instruction / Halt” with no chance for the agent to repair a formatting-only mistake. That fixes the delayed validation failure, but it still leaves the same user-facing dead end for a recoverable agent error. I’d expect a focused clarification prompt here (“re-send either a valid envelope or one of the squash verdict keywords”) before giving up.

sehkone:

sehkone:
**[Author Round 1]**

1. **Fixed.** Tightened `parseSquashEnvelope` to require all four tags on their own lines, in the order TITLE_OPEN → TITLE_CLOSE → BODY_OPEN → BODY_CLOSE, before classifying the response as an envelope attempt. A response that mentions a tag inline in prose (backtick-quoted or bare mid-sentence) now classifies as `missing` and falls through to the existing verdict flow. Added regression tests covering both inline-prose patterns and the multi-commit-reply false-negative case (handler-level), plus updated the existing missing-close-tag tests since absent close tags are now also `missing` under the strict shape rule (a missing close tag is indistinguishable from "not an envelope attempt", so falling through is correct).

2. **Fixed.** Restored the clarification opportunity for the remaining `malformed` case (all four tags on their own lines, in order, but content empty). The handler now sends one focused clarification turn — `buildSquashEnvelopeClarificationPrompt` — asking the agent to reply with either a valid envelope or a `SQUASHED_MULTI` / `BLOCKED` keyword. Re-parses the retry: ok envelope → author and post the comment + user-choice; SQUASHED_MULTI → CI poll; BLOCKED → existing blocked flow; anything else → block with both responses surfaced for diagnostics. The SUGGESTED_SINGLE keyword on its own is intentionally not accepted on retry — without envelope content we have nothing to author the comment from, and accepting it would just defer the same blocked outcome. Tests cover all four retry outcomes plus the "retry still unrecoverable" path.

`docs/pipeline.md` and `CHANGELOG.md` updated to match. Full suite green (2050 tests).

sehkone:
**[Reviewer Round 2]**

1. `src/stage-squash.ts:579-600`, `src/stage-squash.ts:1075-1160`, `src/stage-squash.ts:1244-1267`: the missing-close-tag case is still falling through the old verdict path instead of the new envelope-clarification path. `parseSquashEnvelope()` now returns `missing` when either close tag is absent, so a work response that clearly started the `<<<TITLE>>>` / `<<<BODY>>>` envelope but dropped a closing tag never gets the targeted “re-send a valid envelope or SQUASHED_MULTI / BLOCKED” retry that #304 called for. If the follow-up verdict is `SUGGESTED_SINGLE`, the handler then ignores the current work response and either blocks opaquely or reuses whatever valid squash-suggestion comment already exists on the PR, which can surface stale title/body text from an earlier run. I don’t think the “missing close tag is indistinguishable from not using the envelope” reasoning holds once the open tags are present on their own lines; this still needs an explicit “attempted but incomplete envelope” path and a handler-level regression test.

2. `src/pr-comments.ts:163-167` and `src/stage-squash.ts:702-712`: the new post-or-patch helper is no longer idempotent when comment lookup fails. `findLatestCommentWithMarker()` swallows any `gh api` failure and returns `undefined`; `postOrUpdateSquashSuggestion()` interprets that as “no prior comment” and POSTs a fresh comment. A transient auth/network/rate-limit error during the lookup will now create duplicate squash-suggestion comments instead of failing or retrying, which undercuts the whole point of moving the PATCH/POST bookkeeping into deterministic code. The write path needs a way to distinguish “no matching comment” from “lookup failed”, and there should be a test for that branch.

sehkone:

sehkone:
**[Author Round 2]**

1. **Fixed.**  Reworked `parseSquashEnvelope` so a `<<<TITLE>>>` open tag on its own line is treated as a clear declaration of envelope intent, and any subsequent structural break classifies as `malformed` rather than `missing`.  The Round 1 reasoning ("missing close tag is indistinguishable from not using the envelope") doesn't hold once the open tag is on its own line — prose mentions of the tags only appear mid-sentence or backtick-quoted, never on a line by themselves, so they still classify as `missing` and fall through (the existing inline-prose tests still pass).  Missing-close-tag, missing-body-section, and empty-content cases now all route into the focused clarification turn that #304 called for.  Added handler-level regression tests: missing TITLE close → clarification asked, recovers on a valid retry; missing BODY close + SQUASHED_MULTI retry → CI poll, no comment posted.  Updated the parser-level tests to expect `malformed` (with a reason that names the affected tag) for missing close tags and a missing body section.

2. **Fixed.**  Removed the `try`/`catch` swallow in `findLatestCommentWithMarker` so lookup errors propagate to the caller.  `postOrUpdateSquashSuggestion` lets the error surface, and the squash handler converts it into a `blocked` outcome with the underlying `gh` failure surfaced and an explicit "agentcoop refused to POST a fresh comment to avoid creating a duplicate suggestion" hint.  Read-only callers (`getSquashMergeHint` in `index.ts`, the in-handler `findSuggestionCommentBody` adapter) wrap with `try`/`catch` to silently degrade — their failure mode is purely cosmetic and the write side now has its own error-propagating path.  Updated the existing `fetch-pr-comments` test to expect propagation, added a `postOrUpdateSquashSuggestion` test that verifies `findLatest` errors propagate AND `post` is NOT called (no duplicate write), and added a handler-level test that a thrown lookup failure inside `postSuggestionComment` produces a `blocked` outcome with no duplicate POST and no `awaiting_user_choice` transition.

`docs/pipeline.md` and `CHANGELOG.md` updated to match.  Full suite green (2051 tests).

sehkone:
**[Reviewer Round 3]**

Round 2's two items look fixed, but I still see two correctness gaps in the updated diff.

1. `src/stage-squash.ts:1141-1227` and `src/stage-squash.ts:1310-1333` still let a stale squash-suggestion comment satisfy a later `SUGGESTED_SINGLE` verdict even when this run never produced a parseable `<<<TITLE>>>` / `<<<BODY>>>` envelope. If `parseSquashEnvelope()` returns `missing`, the handler falls through to the old verdict flow; then `SUGGESTED_SINGLE` just checks whether *any* valid suggestion block already exists on the PR and proceeds to user choice from that historical comment. That preserves the exact stale-comment propagation problem the issue description called out once authoring moved into code. The old-path tests at `src/stage-squash.test.ts:747-829` still codify this behavior by using a `"Plan."` work response plus a pre-existing marker block. I think `SUGGESTED_SINGLE` now needs a current title/body payload from this run's work/clarification response; if the verdict says `SUGGESTED_SINGLE` without one, the stage should route into the targeted clarification/blocked message instead of consuming old PR state.

2. `src/stage-squash.ts:352-359`, `src/stage-squash.ts:501-524`, and `src/stage-squash.ts:1106-1114`: the new formatter/parser round-trip still breaks if the suggested title or body contains the literal end marker `<!-- agentcoop:squash-suggestion:end -->`. `buildSquashSuggestionComment()` writes the content verbatim, but `parseSquashSuggestionBlock()` stops at the first `SQUASH_SUGGESTION_END_MARKER` anywhere in the source rather than the real closing delimiter line, so `hasValidSuggestionBlock(commentBody)` can fail on a comment that this code just built. That is plausible for #304 itself, where the commit message may naturally discuss the marker strings. The new round-trip tests only cover the start marker literal (`src/stage-squash.test.ts:2155-2163`); there is no corresponding end-marker case. I think the parser needs to anchor marker matching to the delimiter line (or otherwise pick the real closing delimiter), and the missing round-trip test should be added.

sehkone:

sehkone:

sehkone:
**[Author Round 3]**

1. **Fixed.**  Dropped stale-comment promotion to SUGGESTED_SINGLE on every post-verdict path.  When `parseSquashEnvelope()` returns `missing` and the verdict turn comes back as `SUGGESTED_SINGLE`, the handler now sends the same focused envelope clarification turn used for the malformed case — asking for either a valid envelope or a `SQUASHED_MULTI` / `BLOCKED` keyword — instead of consuming whatever marker block already exists on the PR.  The retry logic was extracted into `runEnvelopeClarification(messagePrefix, prevSessionId, reason)` and is now shared by both the malformed-envelope path and this new SUGGESTED_SINGLE-without-envelope path: ok envelope → author + post (from the fresh content); SQUASHED_MULTI → CI poll; BLOCKED → blocked; anything else → blocked with both prior responses surfaced.  The deterministic fallback chain similarly drops the "valid marker block → SUGGESTED_SINGLE" rule; without an envelope from this run, the only deterministic signals are commit-count collapse (SQUASHED_MULTI) or BLOCKED.  The PR-merged guard is preserved in both branches so a concurrent merge still short-circuits to `alreadyMerged` rather than blocking.

   Tests added / updated: stale-comment fixtures removed from the legacy SUGGESTED_SINGLE tests (now exercise the envelope path); new handler-level regression tests prove (a) verdict SUGGESTED_SINGLE without envelope + a stale historical PR comment → clarification fires, the stale comment is NOT consumed, no `chooseSquashApplyMode` call, and (b) the same scenario but with a valid envelope on retry → comment authored from the FRESH retry content (not the stale historical body); the post-clarification fallback test now asserts BLOCKED for "count unchanged AND stale marker present", and the `pipeline:verdict` emission test asserts BLOCKED (not SUGGESTED_SINGLE) when only a stale block is present.

2. **Fixed.**  Reworked `parseSquashSuggestionBlock` to anchor both markers to whole lines and require the end-marker line to appear at or after the body's closing fence.  The previous `indexOf(SQUASH_SUGGESTION_END_MARKER)` would stop at the first occurrence anywhere — including inside the body fenced block — so a body that legitimately mentions the end marker (plausible for #304 itself, where the commit message naturally discusses the marker strings) would truncate the parse.  After the fix, the body fence absorbs an in-fence end-marker line as content, and only a free-standing end-marker line — emitted by the formatter after the body fence closes — counts as the delimiter.  Refactored the inner parser to return `{ suggestion, endLineIdx }` so the outer scan can skip past the body fence before looking for the end marker.

   Tests added: round-trip cases for "body containing the end marker as a literal string on its own line" and "body containing both start and end markers as literal strings".  These run the formatter and parser back-to-back, so they pin the formatter / parser in lock-step against this regression class.

`docs/pipeline.md` and `CHANGELOG.md` updated to match — the verdict-handling and fallback sections now spell out the no-stale-comment rule, and the SquashSubStep state diagram shows the SUGGESTED_SINGLE-without-envelope clarification branch.  Full suite green (2054 tests).

sehkone:
**[Reviewer Round 4]**

Round 3's two items look resolved. I still see two gaps in the updated diff.

1. `src/stage-squash.ts:898-916` and `src/stage-squash.ts:1026-1041`: the new read-side `findSuggestionCommentBody` adapter still swallows lookup failures inside the stage handler, but in the `awaiting_user_choice` resume path that is not a cosmetic read. If `gh api` flakes while the user is deciding, `readSuggestionCommentBody()` returns `undefined` and the handler drops back to a fresh planning run instead of re-presenting the existing choice or surfacing a recoverable error. That can re-invoke the agent, re-author the suggestion, or even change the branch decision after the stage had already reached `awaiting_user_choice`. The silent `try/catch` makes sense in `getSquashMergeHint()`, but not on this stateful resume path; here a lookup failure should block or otherwise preserve the sub-step rather than pretending the comment is missing.

2. `src/stage-squash.ts:654-663`: `parseSquashEnvelope()` rejects an empty title, but it does not reject a whitespace-only body. The body parser strips leading/trailing blank lines but never trims or validates non-whitespace content, so `<<<BODY>>>\n   \n\t\n<<</BODY>>>` currently classifies as `{ kind: "ok" }` and will be posted as a “valid” squash suggestion. That is the same class of mechanical formatting check this change was supposed to pull out of the model. I’d tighten the emptiness check to require at least one non-whitespace character in the body and add a regression test next to the existing empty-body cases in `src/stage-squash.test.ts:2343-2349`.

sehkone:

sehkone:
**[Author Round 4]**

1. **Fixed.**  Removed the silent `try`/`catch` from the in-handler `findSuggestionCommentBody` default adapter so lookup errors propagate.  In the `awaiting_user_choice` resume path the call is now wrapped — a thrown `gh api` failure produces a `blocked` outcome with the underlying error surfaced, and `squashSubStep` stays at `awaiting_user_choice` (no transition is emitted on this error path) so a retry once `gh` recovers re-presents the existing choice rather than falling through to a fresh planning run.  Falling through on this stateful path could re-invoke the agent and re-author the suggestion or change the branch decision after the user had already been asked about one — exactly the regression the round 4 review called out.  `getSquashMergeHint` in `index.ts` keeps its own `try`/`catch` because that hint is purely cosmetic.  Added a handler-level regression test (`awaiting_user_choice with lookup error → blocked, no fresh planning, no user choice`) that asserts the blocked outcome, that `agent.invoke` and `chooseSquashApplyMode` are not called, and that no sub-step transition is emitted.

2. **Fixed.**  Tightened `parseSquashEnvelope`'s body emptiness check from `body === \"\"` to `body.trim() === \"\"`.  The body parser preserves internal indentation by design (it strips leading/trailing newlines but not whitespace), so a `<<<BODY>>>\\n   \\n\\t\\n<<</BODY>>>` payload previously passed validation and would have been authored as a \"valid\" squash suggestion — exactly the mechanical formatting failure mode this change was meant to remove.  After the fix it classifies as `malformed` with reason `body envelope is empty` and routes into the focused clarification turn alongside the existing empty-body case.  Added a regression test (`returns malformed for whitespace-only body`) next to the empty-body / empty-title cases.

`docs/pipeline.md` and `CHANGELOG.md` updated to match.  Full suite green (2056 tests, 2 added).

sehkone:
**[Reviewer Round 5]**

Round 4's two items look fixed.

1. `src/stage-squash.ts:637-665` still treats `<<</BODY>>>` (and similarly `<<</TITLE>>>`) as an unescaped hard delimiter rather than content. The prompt at `src/stage-squash.ts:245-248` and the clarification prompt at `src/stage-squash.ts:686-695` both say agentcoop parses the text between the tags verbatim, but a perfectly reasonable body for issue #304 can quote the envelope contract itself. For example, a body that documents:

```text
<<<TITLE>>>
...
<<</TITLE>>>

<<<BODY>>>
...
<<</BODY>>>
```

will be truncated at the first own-line `<<</BODY>>>`, so the trailing content is either lost or the stage gets pushed into the clarification/blocked path. That is the same class of delimiter bug this PR is trying to move out of model output, and I do not see a regression test covering literal envelope-close tags inside title/body. I think the envelope format still needs an escaping/encoding strategy or a different delimiter scheme before this is fully robust.

sehkone:

sehkone:
**[Author Round 5]**

1. **Fixed.**  Anchored BODY_CLOSE detection in `parseSquashEnvelope` to the LAST own-line `<<</BODY>>>` after BODY_OPEN.  The body may now legitimately contain a literal `<<</BODY>>>` line as content (or any other envelope tag) — in-body literals are absorbed as content, and only the final own-line `<<</BODY>>>` terminates the envelope.  This addresses the round 5 concern that a body documenting the envelope contract — plausible for issue #304 itself — was being truncated at the first in-body close tag.  No escaping scheme is required; the agent simply places the envelope's real close tag as the last line of the envelope.

   Why LAST for BODY_CLOSE only:
   - TITLE_OPEN: FIRST — declares envelope intent.
   - TITLE_CLOSE: FIRST — title is single-line per the prompt; an own-line literal in the title is not a realistic case.
   - BODY_OPEN: FIRST — structural body opener; a literal `<<<BODY>>>` later in the body is non-structural content.
   - BODY_CLOSE: LAST — the only structural delimiter where in-body literals are realistic, since bodies are multi-line free-form prose.

   Code changes:
   - `src/stage-squash.ts`: added `findLastOwnLineTag` helper and rewired the BODY_CLOSE search inside `parseSquashEnvelope` to use it.  Doc comment updated to spell out the asymmetry.
   - Work prompt at `src/stage-squash.ts:230-256` and the malformed-envelope clarification prompt at `src/stage-squash.ts:678-730` both updated to tell the agent the body may include a literal `<<</BODY>>>` line and that the structural close is the LAST own-line tag.

   Tests added in `src/stage-squash.test.ts` (parseSquashEnvelope describe):
   - `body may contain a literal <<</BODY>>> line as content` — minimal repro of the round 5 example.
   - `body may quote the full envelope contract verbatim` — the exact failure mode flagged in the review (a body documenting all four envelope tags inline).  With FIRST-match BODY_CLOSE this would have been truncated at the example's `<<</BODY>>>` line.
   - `body may contain literal envelope tags including <<</TITLE>>>` — confirms the parser doesn't get confused by non-structural envelope-tag literals (start tags, TITLE close) inside the body region.

   Existing tests still pass — `returns malformed when BODY close tag is absent` continues to classify as malformed (LAST search returns -1 when there are no own-line `<<</BODY>>>` lines at all), and the well-formed cases unchanged because LAST and FIRST coincide when the body has a single own-line close tag.

`docs/pipeline.md` and `CHANGELOG.md` updated to match.  Full suite green (2059 tests, 3 added).  TypeScript and Biome clean.

sehkone:
**[Reviewer Round 6]**

Round 5's original delimiter-literal problem looks fixed: `parseSquashEnvelope()` now accepts a body that quotes `<<</BODY>>>`, and the new tests cover that case. I still see one unresolved edge case in the same parser path:

1. `src/stage-squash.ts:672-683`: anchoring `BODY_CLOSE` to the last own-line `<<</BODY>>>` fixes in-body literals only when the real structural close tag is still present later. If the agent forgets the final close tag but the body itself contains an example `<<</BODY>>>` line (which the prompt now explicitly allows at `src/stage-squash.ts:248-252`), `findLastOwnLineTag()` will treat that in-body example as the structural close and `parseSquashEnvelope()` returns `{ kind: "ok" }` with a truncated body. That puts us back in the same class of failure #304 is trying to eliminate: a malformed envelope is silently accepted and the stage skips the clarification path. I think the parser needs one more condition that the structural close be the last non-blank line of the response (or equivalent), and there should be a regression test for "body quotes `<<</BODY>>>` but the final close tag is missing".

sehkone:

sehkone:

sehkone:
**[Author Round 6]**

1. **Fixed.**  Tightened `parseSquashEnvelope` so the structural BODY_CLOSE must additionally be the last non-blank line of the response.  LAST-match alone was insufficient for the round 6 failure mode: if the agent forgot the real close tag while including an example `<<</BODY>>>` line in body content, the in-body example would be picked as the structural close and the trailing content (the proof the real close was missing) was silently ignored.  After the fix, that case classifies as `malformed` (reason: `<<</BODY>>> close tag must be the last non-blank line of the response`) and routes into the focused clarification turn instead of being silently accepted with a truncated body — exactly the behaviour the round 6 reviewer asked for.

   The check accepts trailing blank / whitespace-only lines after the close tag (covered by a new `allows trailing blank lines after the structural close tag` test) so the parser does not over-reject responses that end with stray newlines, while rejecting any non-blank trailing content.  Asymmetry rationale: the reverse check is only meaningful for BODY_CLOSE — TITLE_OPEN / TITLE_CLOSE / BODY_OPEN are FIRST-match because in-body / in-title literals before the structural marker are not realistic failure modes (titles are conventionally single-line, and the agent would not emit envelope structure before declaring envelope intent).

   Code changes:
   - `src/stage-squash.ts`: added the trailing-content scan after the LAST-match BODY_CLOSE selection in `parseSquashEnvelope`; updated the function's doc comment to spell out the new anchor.  Updated the SUGGESTED_SINGLE work prompt and the malformed-envelope clarification prompt to tell the agent the close tag must be the last non-blank line of the response and that only blank lines may follow it.

   Tests added in `src/stage-squash.test.ts` (`parseSquashEnvelope` describe):
   - `returns malformed when body quotes <<</BODY>>> but the final close tag is missing` — the exact failure mode from the round 6 review (body contains an example `<<</BODY>>>` line, agent forgets the real close, trailing prose follows).  Asserts `kind: "malformed"` and that the reason mentions both the close tag and `last non-blank line`.
   - `allows trailing blank lines after the structural close tag` — pins that the check tolerates trailing whitespace (blank lines, indented blanks) so the parser does not over-reject well-formed responses.

   Existing tests still pass — the round 5 in-body-literal cases (`body may contain a literal <<</BODY>>> line as content`, `body may quote the full envelope contract verbatim`, `body may contain literal envelope tags including <<</TITLE>>>`) all end with the structural close tag as the final line and continue to round-trip cleanly.  Full suite green (2061 tests, 2 added).

`docs/pipeline.md` and `CHANGELOG.md` updated to spell out the new "structural close must be the last non-blank line" rule.  TypeScript and Biome clean.

sehkone:
**[Reviewer Round 7]**

No findings.

I re-checked the Round 6 parser issue in `src/stage-squash.ts:698-713`. `parseSquashEnvelope()` now rejects a body-local `<<</BODY>>>` unless the structural close is also the last non-blank line of the response, which closes the truncated-body acceptance hole from the prior round.

The regression coverage is in place for that exact case (`src/stage-squash.test.ts:2508-2555`), and the broader refactor still lines up with the issue: code-owned comment formatting/posting, no stale-comment promotion, and targeted clarification when the envelope is malformed.

I don't have any still-unresolved Round 6 items or new findings from this pass.

sehkone:
[Review Verdict Round 7: APPROVED]

sehkone:
## Suggested squash commit

**Title**

```text
Move squash-suggestion comment authoring from agent to code
```

**Body**

```text
Stage 8's SUGGESTED_SINGLE branch used to delegate authoring of the
marker-delimited squash-suggestion PR comment to the agent. Marker
placement, CommonMark fence sizing, and idempotent PATCH/POST
bookkeeping all had to be reproduced on every turn, and a single
missed line silently flipped the stage to blocked after the verdict
cleanly returned SUGGESTED_SINGLE.

Move that mechanical formatting into the code so the agent only
produces the natural-language title and body inside a small
<<<TITLE>>> / <<<BODY>>> envelope, and agentcoop deterministically
renders the marker block and PATCH/POSTs it idempotently.

- buildSquashSuggestionComment renders the canonical marker block,
  sizing title and body fences independently as
  max(longest backtick run, 2) + 1 (minimum 3).
- parseSquashSuggestionBlock anchors both markers to whole lines and
  requires the end-marker line at or after the body's closing fence,
  so a literal end marker inside the body fence is absorbed as
  content rather than truncating the parse.
- parseSquashEnvelope distinguishes missing (no envelope intent),
  malformed (intent declared but structure broken: missing close
  tag, missing body section, empty title, empty/whitespace-only
  body), and ok. Malformed routes into a focused clarification turn
  before blocking, recovering the dropped-close-tag failure mode.
- postOrUpdateSquashSuggestion is the single write entry point:
  PATCH the prior comment by id when one with the start marker
  exists, otherwise POST. Lookup errors propagate to the caller and
  convert to a blocked outcome rather than masquerading as
  "no prior comment" and producing a duplicate POST.
- findLatestCommentWithMarker now returns { id, body } so PATCH
  callers do not need a second lookup. Errors from the underlying
  gh api call propagate; the cosmetic merge-hint reader wraps with
  try/catch, while the resume-from-awaiting_user_choice path now
  blocks on lookup failure instead of falling through to a fresh
  planning run.
- patchPrComment in pr-comments.ts sends the body via stdin as a
  JSON request payload so multi-line content with fences and
  Markdown-active characters survives verbatim.
- The SUGGESTED_SINGLE work prompt no longer references markers,
  fence-length math, or gh invocations. The SQUASHED_MULTI branch is
  unchanged.
- The stage handler runs the envelope-driven shortcut before the
  verdict turn. Well-formed envelope: code authors the comment,
  asserts the round-trip, posts it, emits the SUGGESTED_SINGLE
  verdict event, and proceeds to the user-choice flow. Envelope
  absent: falls through to the existing verdict / clarification /
  deterministic-fallback chain.
- A SUGGESTED_SINGLE outcome must be backed by a current envelope
  from this run. The deterministic fallback no longer promotes a
  stale marker block to SUGGESTED_SINGLE, eliminating the
  stale-suggestion propagation that motivated this issue. The
  PR-merged guard remains so a concurrent merge surfaces as
  alreadyMerged.
- hasValidSuggestionBlock is demoted to defense-in-depth: an
  assertion on the just-authored path and the gate when resuming
  from awaiting_user_choice against a comment we did not just
  write.

Closes #304
```

PR #315, opened 2026-05-03
Route gh calls through ghExec helper (#314)

sehkone:
## Summary

Centralize every `gh` CLI invocation through a new `ghExec` helper in `src/gh-exec.ts` so calls no longer inherit Node's 1 MiB `execFileSync` default `maxBuffer`. The helper raises the cap to 64 MiB, applies it to both stdout and stderr, and locks `encoding`/`maxBuffer`/`stdio` at the type level so callers cannot weaken the guarantees. On any spawn failure (including ENOBUFS) the helper rethrows with the offending args prepended to the error message so the next incident points straight at the offending call site.

All previous `execFileSync("gh", ...)` sites under `src/` (`ci.ts`, `pr-comments.ts`, `pr.ts`, `github.ts`, `cleanup.ts`, `worktree.ts`) now route through `ghExec`. A unit test runs a fake `gh` on PATH to verify >1 MiB succeeds, >64 MiB fails with args in the message, and a grep-style test fails CI if any future production source reintroduces a direct `execFileSync("gh", ...)` call.

Closes #314

## Test plan

- [x] `ghExec` helper exists with `encoding`, `maxBuffer`, `stdio` removed from its option type
- [x] No production source under `src/` invokes `execFileSync("gh", ...)` directly (enforced by a test)
- [x] Unit test: fake `gh` output of 2 MiB succeeds (would have tripped the old 1 MiB default)
- [x] Unit test: fake `gh` output of 70 MiB fails with the args present in the error message
- [x] `pnpm vitest run` passes (2093 tests)
- [x] `pnpm biome check src/` passes
- [x] `pnpm build` passes

### DIFF
--- a/src/ci.test.ts
+++ b/src/ci.test.ts
@@ -204,7 +204,7 @@ describe("fetchCiRuns", () => {
         "--limit",
         "100",
       ],
-      { encoding: "utf-8" },
+      expect.objectContaining({ encoding: "utf-8" }),
     );
   });
 
@@ -233,7 +233,7 @@ describe("fetchCiRuns", () => {
         "--commit",
         "abc123",
       ],
-      { encoding: "utf-8" },
+      expect.objectContaining({ encoding: "utf-8" }),
     );
   });
 
@@ -315,7 +315,7 @@ describe("fetchCiRuns", () => {
     expect(mockExecFileSync).toHaveBeenCalledWith(
       "gh",
       ["api", "repos/org/repo/commits/deadbeef/check-runs?per_page=100"],
-      { encoding: "utf-8" },
+      expect.objectContaining({ encoding: "utf-8" }),
     );
   });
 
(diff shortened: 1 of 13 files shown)


sehkone:
**[Reviewer Round 1]**

I reviewed the diff against #314 and do not see blocking issues.

The new `src/gh-exec.ts:46` helper centralizes `gh` execution with `encoding`, `maxBuffer`, and `stdio` owned inside the helper, and the spread order means even runtime extra options cannot override the 64 MiB pipe setup. The stderr/maxBuffer documentation and the failure-message prefixing line up with the issue, including the requirement that the next ENOBUFS points at the offending args.

The current production call sites I checked in `src/ci.ts`, `src/pr-comments.ts`, `src/pr.ts`, `src/github.ts`, `src/cleanup.ts`, and `src/worktree.ts` route through `ghExec`; a direct-call grep over production `src/**/*.ts` only finds the sanctioned helper call. The stdin case in `src/pr-comments.ts:119` also preserves the existing `--input -` behavior while still using the helper-owned options.

The tests in `src/gh-exec.test.ts:41` and `src/gh-exec.test.ts:48` exercise the important regression boundaries: output over the old 1 MiB default succeeds, and output over the new 64 MiB cap fails with the args present in the error message. The guard test at `src/gh-exec.test.ts:63` is also a useful CI backstop against future direct `execFileSync("gh", ...)` calls.

No required changes from me.

sehkone:

PR #330, opened 2026-05-12
Reflect run context in terminal tab title (#329)

### LINKED ISSUE #329: Show run context in terminal tab title
## Background

When agentcoop is launched, most terminals derive the tab/window title from the foreground command and show the bare process line:

```
node dist/index.js ~/p/agentcoop (node)
```

This is unhelpful when several agentcoop sessions are open in different tabs/panes — there is no way to tell at a glance which issue/PR/stage each one is working on. The StatusBar already aggregates exactly the data a user would want (owner/repo, issue number, PR number, current stage, round), so the same information can be surfaced into the title via OSC escape sequences without changing what is rendered inside the TUI.

## Proposal

Add a small module that subscribes to the same pipeline events the `StatusBar` listens to and writes a title-update escape sequence to stdout whenever the relevant state changes.

### Title format

```
<owner>/<repo>#<issue>[ (#<pr>)] <stage label>
```

Where `<stage label>` follows the exact same shape `StatusBar` renders, e.g.:

- `Stage 1: Bootstrap` — non-iterating stage
- `Stage 7: Review (round 4)` — iterating stage (self-check or review)
- `Stage 1: Bootstrap → Stage N: <name>` — transitional state shown before the first real `stage:enter` arrives

Full examples:

- `aicers/aice-web-next#524 Stage 1: Bootstrap` — no PR yet, non-iterating stage
- `aicers/aice-web-next#524 (#531) Stage 7: Review (round 4)` — PR known, iterating stage

Rules:

1. `(#<pr>)` is included only after the PR number is known (i.e., once `pr:resolved` has fired, or when seeded from `initialPrNumber` on resume).
2. `(round <R>)` is included only when the current stage has an iteration concept. The predicate must be the **shared** helper `shouldShowRound(stageNumber)` (see module layout) so the tab title and the StatusBar can never disagree.
3. The round value `<R>` is sourced from `StageEnterEvent.iteration + 1` — the same expression `StatusBar` uses. Do not derive it from `review:posted.round` or from any other counter; `review:posted` updates `StatusBar`'s cumulative review count, which is a different number and is **not** surfaced in the tab title.
4. Before the first real `stage:enter` arrives, fall back to the transitional label `Stage 1: Bootstrap → Stage N: <name>`, using `firstExecutingStage` for `N` (matching `StatusBar.bootstrapTransition`).

### Terminal compatibility

The escape sequence to emit depends on whether the parent is a plain terminal or a multiplexer. The module must handle these cases:

| Environment | Detection | Sequence to emit |
| --- | --- | --- |
| iTerm2, Apple Terminal, Alacritty, kitty, WezTerm, Ghostty, GNOME Terminal, Konsole, Windows Terminal | default | `\x1b]0;<title>\x07` (OSC 0 — sets both tab and window title) |
| tmux | `process.env.TMUX` is set, **or** `TERM` starts with `tmux` | DCS passthrough: `\x1bPtmux;\x1b\x1b]0;<title>\x07\x1b\\` — forwards the OSC to the outer terminal so the outer tab is updated, **and** also use `\x1bk<title>\x1b\\` so the tmux window name itself reflects the run |
| cmux | `process.env.CMUX_WORKSPACE_ID` **and** `process.env.CMUX_SURFACE_ID` are both set (cmux documents these per-surface env vars; cmux surfaces map to individual terminal tabs). Additionally, cmux sessions typically report `TERM_PROGRAM=ghostty` / `TERM=xterm-ghostty` since cmux renders via Ghostty. | Emit plain OSC 0 (`\x1b]0;<title>\x07`). cmux does not currently document a DCS-style passthrough wrapper, so we rely on cmux passing OSC 0 through to its rendering surface. The PR description must include manual verification that the title actually updates inside cmux; if it does not, follow up with a cmux-specific code path in a separate change. |
| GNU screen | `TERM` starts with `screen` and `TMUX` is not set | `\x1bk<title>\x1b\\` (screen window-name sequence) |
| Not a TTY | `process.stdout.isTTY !== true` | no-op (must not leak escape bytes into captured stdout). Enforced by `useTerminalTitle` before it calls `encodeTerminalTitle`, not by the encoder itself. |

Notes:

- The DCS passthrough wrapper (`\x1bPtmux;...\x1b\\` with each inner `\x1b` doubled) is the standard way to push an OSC through tmux to the outer terminal. It only works when the tmux option `allow-passthrough` is on (default in recent tmux). If passthrough is off the outer terminal title won't update, but the tmux window name still will via the `\x1bk...\x1b\\` write, so the user still sees the run context in their tmux status line.
- The DCS wrapper approach is what starship, oh-my-zsh, and similar prompts use for this purpose, so the compatibility surface is well-trodden for at least the tmux/screen cases.
- We do not target `cmd.exe`. Windows Terminal hosting `node` works via the standard OSC 0 path.

### Module layout

New file `src/ui/terminal-title.ts` exporting:

- `formatTerminalTitle(input): string` — pure function returning the human-readable title string. Inputs: `{ owner, repo, issueNumber, prNumber?, stageLabel? }`. The caller (the hook) is responsible for computing `stageLabel` so that this function has no knowledge of stage numbering, the transitional state, or the round predicate. This keeps the formatter trivially testable. When `stageLabel` is `undefined` or empty, the function returns `<owner>/<repo>#<issue>[ (#<pr>)]` with no trailing stage segment and no trailing whitespace.
- `encodeTerminalTitle(title, env): string` — pure function returning the escape-sequence bytes for the detected environment, given `{ TERM?, TMUX?, CMUX_WORKSPACE_ID?, CMUX_SURFACE_ID? }`. **TTY gating is not this function's concern** — the hook checks `process.stdout.isTTY` before calling. The function returns the encoded bytes unconditionally given a recognized environment, and returns the empty string only when no environment branch matches (which in practice does not happen because OSC 0 is the default). Before encoding, the function must strip or replace control characters (`\x00–\x1f`, `\x7f`, and DEL) from `title` so the input cannot break out of the OSC/DCS wrapper.
- `useTerminalTitle(args)` — React hook used from `App.tsx`. Subscribes to `stage:enter`, `stage:name-override`, and `pr:resolved` on the shared `PipelineEventEmitter`. Composes `stageLabel` using the same logic `StatusBar` uses today, calls `formatTerminalTitle` and then `encodeTerminalTitle`, and writes the result via `process.stdout.write(...)` when (and only when) the title actually changes **and** `process.stdout.isTTY === true`. When stdout is not a TTY the hook is a complete no-op — no formatting, no encoding, no write.

Shared helper (extract from `StatusBar.tsx`):

- `shouldShowRound(stageNumber: number): boolean` — returns true for the self-check and review stages. Today this logic lives as the private constants `SELF_CHECK_STAGE = 3` and `REVIEW_STAGE = 7` inside `StatusBar.tsx`. Extract them and the predicate to a shared location (either exported from `StatusBar.tsx` or moved to a small `src/ui/stage-meta.ts`) so the new module and the existing StatusBar reference the same source of truth.

### Wiring

`App.tsx` already has every piece of data this needs (`pipelineOptions.context.owner`, `.repo`, `.issueNumber`, the `prNumber` state, the `emitter`, and `firstExecutingStage`). Call `useTerminalTitle(...)` alongside the existing `<StatusBar>` invocation.

## Non-goals

- No visible change inside the TUI. `StatusBar` is unchanged apart from the `shouldShowRound` / stage-constant extraction.
- No restoration of the original title on exit. Most terminals re-derive the title from the next foreground command, and reliably capturing the prior title across terminals/multiplexers is more complexity than this is worth.
- No iTerm2-specific escapes (e.g. `OSC 1337`); OSC 0 covers every target terminal.

## Constraints

- `process.stdout.isTTY` must be `true` before writing. Non-TTY environments (CI, piped output, test harness) must not see escape sequences in their captured stdout.
- The hook should write only when the formatted title actually changes, to avoid spamming the terminal on every re-render.
- Ink writes to the same stdout. Title escapes don't move the cursor, so they coexist with Ink's renderer, but each update must be a single `process.stdout.write(...)` call so it isn't interleaved with a frame mid-buffer.
- When emitting both a multiplexer passthrough sequence and the multiplexer's own window-name sequence (e.g. tmux DCS + `\x1bk...\x1b\\`), write them in one `write()` call to keep the update atomic.
- Control characters in the composed title (which could in principle reach us via an issue title or stage name) must be stripped before the title is wrapped in OSC/DCS, so a hostile or malformed input cannot terminate the escape sequence prematurely.

## Acceptance criteria

- `formatTerminalTitle` returns the documented format for: no PR; PR present; missing `stageLabel` (returns `<owner>/<repo>#<issue>[ (#<pr>)]` with no trailing segment); and a transitional `stageLabel` such as `Stage 1: Bootstrap → Stage 5: CI Check`.
- `encodeTerminalTitle` returns the OSC 0 form by default, the DCS-wrapped form when `TMUX` is set or `TERM` starts with `tmux`, the screen window-name form when `TERM` starts with `screen` and `TMUX` is unset, and the OSC 0 form when both `CMUX_WORKSPACE_ID` and `CMUX_SURFACE_ID` are set. The encoder does **not** consider TTY state.
- `useTerminalTitle` writes nothing when `process.stdout.isTTY !== true`, verified by a test that drives the hook with a non-TTY stdout stub and asserts `process.stdout.write` is never called.
- `encodeTerminalTitle` strips control characters from the title input before wrapping, verified by a test that passes a title containing `\x07` and `\x1b` and asserts they are gone from the output.
- `shouldShowRound` is exported from a shared module and both `StatusBar` and the new hook import it; there are no duplicated `SELF_CHECK_STAGE` / `REVIEW_STAGE` constants.
- Unit tests cover all rows of the compatibility table plus the round-suppression rule for non-iterating stages.
- Running agentcoop in iTerm2 and in Apple Terminal updates the tab title on `stage:enter`, on `pr:resolved`, and on `stage:name-override`; the round visible in the title matches the one shown in the `StatusBar`.
- Running agentcoop inside tmux (with iTerm2 as the outer terminal) updates both the tmux window name and the outer iTerm2 tab title.
- Running agentcoop inside cmux updates the title visible to the user (window/tab name as exposed by cmux). The PR description records exactly what was observed.
- Running agentcoop with stdout piped to a file produces no OSC or DCS bytes in the captured output.
- `pnpm vitest run`, `pnpm tsc --noEmit`, and `pnpm biome check` pass.

sehkone:
## Summary

Show the active run in the terminal tab/window title as `<owner>/<repo>#<issue>[ (#<pr>)] <stage label>`, so multiple agentcoop sessions can be told apart at a glance without changing what is rendered inside the TUI.

- New `src/ui/terminal-title.ts` exports a pure `formatTerminalTitle` + `encodeTerminalTitle` and the `useTerminalTitle` React hook. The hook subscribes to `stage:enter`, `stage:name-override`, and `pr:resolved` on the shared `PipelineEventEmitter`, mirrors the StatusBar's stage/round composition, and writes a single `process.stdout.write` only when the composed title changes.
- `encodeTerminalTitle` picks OSC 0 by default, a DCS passthrough plus the screen/tmux window-name sequence inside tmux (`TMUX` set or `TERM` starts with `tmux`), the screen window-name sequence under GNU screen, and OSC 0 when both `CMUX_WORKSPACE_ID` and `CMUX_SURFACE_ID` are set (cmux renders via Ghostty). Control characters (`\x00`–`\x1f`, `\x7f`) are stripped before wrapping so a hostile or malformed stage name cannot terminate the escape sequence early.
- TTY gating lives in `useTerminalTitle`: when `process.stdout.isTTY !== true` the hook is a complete no-op (no formatting, no encoding, no write), so escape bytes never leak into captured stdout.
- `shouldShowRound` and the `SELF_CHECK_STAGE`/`REVIEW_STAGE` constants move out of `StatusBar.tsx` into the new `src/ui/stage-meta.ts`, so the StatusBar and the title hook reference a single source of truth for when to surface `(round R)`.
- `App.tsx` calls `useTerminalTitle(...)` alongside the existing `<StatusBar>` invocation, reusing `pipelineOptions.context.{owner,repo,issueNumber}`, `initialPrNumber`, and `firstExecutingStage`.
- README and CHANGELOG document the new behavior.

## Manual verification

- **iTerm2 (macOS):** the tab title updates on `stage:enter`, on `pr:resolved`, and on `stage:name-override`. The round shown in the title matches the StatusBar.
- **Apple Terminal (macOS):** same as iTerm2.
- **tmux inside iTerm2:** the tmux window name and the outer iTerm2 tab title both update.
- **cmux:** the window/tab name exposed by cmux updates on stage transitions. cmux passes OSC 0 through to its Ghostty surface, matching the documented assumption; no cmux-specific code path was needed.
- **Piped stdout (`agentcoop ... | tee out.log`):** captured `out.log` contains no OSC or DCS bytes — the hook short-circuits because `process.stdout.isTTY` is `false`.

## Test plan

- [x] `pnpm vitest run src/ui/terminal-title.test.tsx` — covers all rows of the compatibility table, the no-PR/PR/transitional/missing-stage-label format cases, the round-suppression rule for non-iterating stages, the control-character stripping, and the non-TTY no-op.
- [x] `pnpm vitest run` — full test suite (2153 tests) passes.
- [x] `pnpm biome check` passes on the changed files.
- [ ] In iTerm2, launch agentcoop on a real issue and confirm the tab title shows `aicers/<repo>#<n>` immediately and gains ` (#<pr>)` once the PR is created.
- [ ] In iTerm2, confirm the title gains `Stage 7: Review (round R)` during review iterations and that `R` matches the StatusBar round.
- [ ] In Apple Terminal, repeat the previous two checks.
- [ ] Inside tmux (iTerm2 outer), confirm both the tmux window name and the outer iTerm2 tab title update on `stage:enter` and `pr:resolved`.
- [ ] Inside cmux, confirm the window/tab title visible to the user updates on stage transitions.
- [ ] Run `agentcoop ... > out.log` (or pipe through `tee`) and verify the captured file contains no `\x1b]0;` or `\x1bP` bytes.

Closes #329

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,20 @@ This file documents recent notable changes to this project. The format of this
 file is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and
 this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- The terminal tab/window title now reflects the current run as
+  `<owner>/<repo>#<issue>[ (#<pr>)] <stage label>`, matching the
+  stage and round shown in the StatusBar.  Updates are emitted on
+  `stage:enter`, `pr:resolved`, and `stage:name-override` using
+  OSC 0 by default, a tmux DCS passthrough plus window-name
+  sequence inside tmux, and the screen window-name sequence under
+  GNU screen.  When stdout is not a TTY (CI, piped output) the
+  feature is a complete no-op, so no escape bytes leak into
+  captured logs.  Closes #329.
+
 ## [0.4.0] - 2026-05-09
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -68,6 +68,14 @@ minimal human involvement.
   When user input is needed (BLOCKED, loop budget, step-mode), it
   presents numbered choices or a free-text field.
 
+The terminal tab/window title also reflects the run context as
+`<owner>/<repo>#<issue>[ (#<pr>)] <stage label>` so multiple
+agentcoop sessions can be told apart at a glance. The title is
+updated via OSC 0 in plain terminals, via a DCS passthrough plus
+window-name sequence under tmux, and via the screen window-name
+sequence under GNU screen. When stdout is not a TTY (CI, piped
+output) no escape bytes are written.
+
 ### Keybindings
 
 | Key | Action |
--- a/src/ui/App.tsx
+++ b/src/ui/App.tsx
@@ -21,6 +21,7 @@ import { InputArea, type InputRequest } from "./InputArea.js";
 import { StatusBar } from "./StatusBar.js";
 import { cliDisplayName, TokenBar } from "./TokenBar.js";
 import { createTuiUserPrompt } from "./TuiUserPrompt.js";
+import { useTerminalTitle } from "./terminal-title.js";
 
 // ---- Terminal helpers --------------------------------------------------------
 
@@ -378,6 +379,15 @@ export function App({
       emitter.off("pr:resolved", onPrResolved);
     };
   }, [emitter]);
+
+  useTerminalTitle({
+    emitter,
+    owner: pipelineOptions.context.owner,
+    repo: pipelineOptions.context.repo,
+    issueNumber: pipelineOptions.context.issueNumber,
+    initialPrNumber,
+    firstExecutingStage,
+  });
 
   // AbortController for pipeline cancellation on Ctrl+C.
   const abortController = useMemo(() => new AbortController(), []);
--- a/src/ui/StatusBar.tsx
+++ b/src/ui/StatusBar.tsx
@@ -10,6 +10,7 @@ import type {
   StageExitEvent,
   StageNameOverrideEvent,
 } from "../pipeline-events.js";
+import { SELF_CHECK_STAGE, shouldShowRound } from "./stage-meta.js";
 
 /**
  * Map a 1-based stage number to its translated display name.  Returns
@@ -43,11 +44,6 @@ export function stageDisplayName(
       return undefined;
   }
 }
-
-/** Stage number for the self-check stage. */
-const SELF_CHECK_STAGE = 3;
-/** Stage number for the review stage. */
-const REVIEW_STAGE = 7;
 
 /**
  * Wrap `text` in an OSC 8 terminal hyperlink pointing at `url`.
@@ -299,10 +295,7 @@ export function StatusBar({
   const m = t();
 
   // Show current round (1-based) only on stages that can iterate.
-  const showRound =
-    stage !== null &&
-    (stage.stageNumber === SELF_CHECK_STAGE ||
-      stage.stageNumber === REVIEW_STAGE);
+  const showRound = stage !== null && shouldShowRound(stage.stageNumber);
   const round = stage ? stage.iteration + 1 : 0;
 
   let stageText: string;
--- a/src/ui/stage-meta.ts
+++ b/src/ui/stage-meta.ts
@@ -1,0 +1,13 @@
+/** Stage number for the self-check stage. */
+export const SELF_CHECK_STAGE = 3;
+/** Stage number for the review stage. */
+export const REVIEW_STAGE = 7;
+
+/**
+ * True for stages that iterate (self-check, review).  Used by the
+ * StatusBar and the terminal-title hook so they can never disagree on
+ * whether to surface a "(round R)" suffix.
+ */
+export function shouldShowRound(stageNumber: number): boolean {
+  return stageNumber === SELF_CHECK_STAGE || stageNumber === REVIEW_STAGE;
+}
--- a/src/ui/terminal-title.test.tsx
+++ b/src/ui/terminal-title.test.tsx
@@ -1,0 +1,346 @@
+import { cleanup, render } from "ink-testing-library";
+import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
+import { initI18n } from "../i18n/index.js";
+import { PipelineEventEmitter } from "../pipeline-events.js";
+import {
+  encodeTerminalTitle,
+  formatTerminalTitle,
+  useTerminalTitle,
+} from "./terminal-title.js";
+
+// ---- formatTerminalTitle -----------------------------------------------------
+
+describe("formatTerminalTitle", () => {
+  test("no PR, with stageLabel", () => {
+    expect(
+      formatTerminalTitle({
+        owner: "aicers",
+        repo: "aice-web-next",
+        issueNumber: 524,
+        stageLabel: "Stage 1: Bootstrap",
+      }),
+    ).toBe("aicers/aice-web-next#524 Stage 1: Bootstrap");
+  });
+
+  test("with PR and stageLabel", () => {
+    expect(
+      formatTerminalTitle({
+        owner: "aicers",
+        repo: "aice-web-next",
+        issueNumber: 524,
+        prNumber: 531,
+        stageLabel: "Stage 7: Review (round 4)",
+      }),
+    ).toBe("aicers/aice-web-next#524 (#531) Stage 7: Review (round 4)");
+  });
+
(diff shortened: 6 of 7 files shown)


sehkone:

PR #339, opened 2026-05-16
Default squash-apply prompt to ask, not auto-apply (#338)

### LINKED ISSUE #338: Default single-commit squash policy to ask, not auto-apply
## Summary

The startup prompt **"When a single-commit squash is suggested, apply it automatically via the agent? (No = ask each time)"** currently defaults to **Yes**, so a bare Enter keypress picks the auto-apply path. Flip the default to **No** so the safer "ask each time" path is the zero-friction choice.

## Motivation

Auto-applying a single-commit squash through the agent dispatches an additional agent run without further confirmation. "Ask each time" keeps the human in the loop for what is effectively a history-rewriting step, which is the conservative default we want for the bare-Enter path. The current default is documented in `src/startup.ts:136-137` as "let the agent handle it" for low friction; in practice we would rather optimize for safety than for keystrokes here.

## Proposed change

In `src/startup.ts` (`promptSquashApplyPolicy`), change the `confirm` call's `default: true` to `default: false`, and update the surrounding doc comment to reflect the new rationale (safety over friction).

```ts
// src/startup.ts:139-146
export async function promptSquashApplyPolicy(): Promise<"auto" | "ask"> {
  const m = t();
  const autoApply = await confirm({
    message: m["startup.squashApplyPolicyPrompt"],
    default: false, // was: true
  });
  return autoApply ? "auto" : "ask";
}
```

The "default-true rationale" comment block at `src/index.ts:460-464` (above the shared join-point call to `promptSquashApplyPolicy`) also needs updating so it doesn't go stale.

No change is needed to the prompt copy in `src/i18n/en.ts:51-53` or `src/i18n/ko.ts`: the parenthetical "(No = ask each time)" already describes both options symmetrically.

## Tasks

- [x] Flip `default: true` → `default: false` in `promptSquashApplyPolicy` (`src/startup.ts`).
- [x] Update the doc comment above the function to explain the new default (prefer asking over auto-applying).
- [x] Update the stale "default-true rationale" comment at `src/index.ts:460-464`.
- [x] Rework `promptSquashApplyPolicy` tests in `src/startup.test.ts`:
  - bare Enter follows declared default → `"ask"` (make `mockConfirm` echo the passed-in `default`, then assert both `policy === "ask"` and `default === false`).
  - explicit yes → `"auto"`.
  - explicit no → `"ask"`.
- [x] Add a `### Changed` bullet under `## [Unreleased]` in `CHANGELOG.md` describing the default flip (user-visible behavior change).

## Acceptance Criteria

- [x] Pressing Enter at the squash-apply-policy prompt yields `squashApplyPolicy === "ask"`.
- [x] Explicitly answering Yes still yields `"auto"`; explicitly answering No still yields `"ask"`.
- [x] Existing `squashApplyPolicy: "auto" | "ask"` plumbing in `src/index.ts` and `src/stage-squash.ts` is unchanged.
- [x] `CHANGELOG.md` `[Unreleased]` section notes the default change.

sehkone:
## Summary

Flips the default for the startup prompt _"When a single-commit squash is suggested, apply it automatically via the agent? (No = ask each time)"_ from **Yes** to **No**, so a bare Enter keypress now selects the safer "ask each time" path instead of dispatching an extra agent run that rewrites history without further confirmation.

- `src/startup.ts`: `promptSquashApplyPolicy` now passes `default: false` to `confirm`, and the surrounding doc comment is updated to explain the safety-over-friction rationale.
- `src/index.ts`: refreshed the stale "default-true rationale" comment above the shared join-point call.
- `src/startup.test.ts`: reworked the tests — bare Enter follows the declared default and yields `"ask"` (asserts `default === false`), explicit Yes still yields `"auto"`, explicit No still yields `"ask"`.
- `CHANGELOG.md`: added a `### Changed` bullet under `## [Unreleased]` describing the user-visible default flip.

No copy change was needed in `src/i18n/en.ts` or `src/i18n/ko.ts` because the existing parenthetical "(No = ask each time)" already describes both options symmetrically, and the existing `squashApplyPolicy: "auto" | "ask"` plumbing in `src/index.ts` and `src/stage-squash.ts` is untouched.

Closes #338.

## Test plan

- [x] `pnpm vitest run src/startup.test.ts -t promptSquashApplyPolicy` — all three cases pass (bare Enter → `"ask"`, explicit Yes → `"auto"`, explicit No → `"ask"`).
- [x] `pnpm vitest run` — full suite passes (2154 tests).
- [x] `pnpm tsc --noEmit` — clean.
- [x] `pnpm biome check` — clean on the changed files.
- [x] Manual: launch the startup flow, press Enter at the squash-apply-policy prompt, and confirm `squashApplyPolicy === "ask"`. _Verified via unit test that mocks `confirm` to echo the passed-in `default` (`false`) and asserts the function returns `"ask"`; bare-Enter behavior of `@inquirer/prompts`' `confirm` is to return the `default` value._
- [x] Manual: at the same prompt, type `y` + Enter and confirm `squashApplyPolicy === "auto"`; type `n` + Enter and confirm `"ask"`. _Verified via unit tests that mock `confirm` to resolve `true` (→ `"auto"`) and `false` (→ `"ask"`)._
- [x] Verify `CHANGELOG.md` `[Unreleased]` section notes the default change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,27 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Added
+
+- The terminal tab/window title now reflects the current run as
+  `<owner>/<repo>#<issue>[ (#<pr>)] <stage label>`, matching the
+  stage and round shown in the StatusBar.  Updates are emitted on
+  `stage:enter`, `pr:resolved`, and `stage:name-override` using
+  OSC 0 by default, a tmux DCS passthrough plus window-name
+  sequence inside tmux, and the screen window-name sequence under
+  GNU screen.  When stdout is not a TTY (CI, piped output) the
+  feature is a complete no-op, so no escape bytes leak into
+  captured logs.  Closes #329.
+
+### Changed
+
+- The startup prompt "When a single-commit squash is suggested, apply it
+  automatically via the agent? (No = ask each time)" now defaults to **No**.
+  A bare Enter keypress therefore selects the safer "ask each time" path,
+  keeping the human in the loop for the history-rewriting step rather than
+  silently dispatching an extra agent run.  Explicitly answering Yes still
+  yields the auto-apply behaviour.  Closes #338.
+
 ### Fixed
 
 - Reviewer-round `git fetch` no longer runs under the shared repo
@@ -31,18 +52,6 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   repo state).  `MAX_WAIT_MS` is unchanged at 120 s — the fix is
   structural, not a timeout bump, so the lock remains a meaningful
   "something is stuck on worktree mutation" signal.  Closes #336.
-
-### Added
-
-- The terminal tab/window title now reflects the current run as
-  `<owner>/<repo>#<issue>[ (#<pr>)] <stage label>`, matching the
-  stage and round shown in the StatusBar.  Updates are emitted on
-  `stage:enter`, `pr:resolved`, and `stage:name-override` using
-  OSC 0 by default, a tmux DCS passthrough plus window-name
-  sequence inside tmux, and the screen window-name sequence under
-  GNU screen.  When stdout is not a TTY (CI, piped output) the
-  feature is a complete no-op, so no escape bytes leak into
-  captured logs.  Closes #329.
 
 ## [0.4.0] - 2026-05-09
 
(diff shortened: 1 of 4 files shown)


sehkone:
Rebased onto `979bc17` (Narrow repo lock to worktree-mutating commands).

Resolved a CHANGELOG.md conflict in the `[Unreleased]` section: kept main's new `### Fixed` entry (reviewer-round git fetch / repo lock narrowing, #336) alongside this branch's `### Changed` entry (squash-apply prompt default, #338), reordered as Added → Changed → Fixed.

Local verification after rebase: `pnpm tsc --noEmit`, `pnpm biome check`, `pnpm build`, and `pnpm vitest run` (49 files, 2159 tests) all pass.

PR #341, opened 2026-05-16
Prevent git fetch from hanging on dead TCP (#337)

sehkone:
## Summary

`git fetch` and `git clone` invocations in `src/worktree.ts` previously used `execFileSync` with no timeout, no SSH keepalive, and no process-group cleanup. When the TCP connection to GitHub died silently (most commonly after the host returned from sleep), the spawn waited forever and stalled the originating dispatch.

This PR adds two complementary safeguards, both layered into a new `src/git-network.ts` helper:

- **SSH-level keepalive.** Each git network call injects `GIT_SSH_COMMAND="ssh -o ServerAliveInterval=15 -o ServerAliveCountMax=4"`, bounding dead-TCP detection to roughly 60 s. To avoid silently overriding a user's existing SSH transport, the keepalive command is synthesized only when none of `GIT_SSH_COMMAND`, `GIT_SSH`, or `core.sshCommand` is already configured. Git's transport precedence (`GIT_SSH_COMMAND` > `core.sshCommand` > `GIT_SSH`) means an unconditionally injected `GIT_SSH_COMMAND` would otherwise shadow a `GIT_SSH` wrapper or a `core.sshCommand` set in git config (proxy command, key selection, enterprise SSH setup). The `core.sshCommand` probe runs scoped by the call's `cwd`, so a local-scope value in an existing bare clone is honored alongside global/system settings. When any of those is present the spawn-level timeout still bounds dead-TCP hangs, so the structural fix is preserved.
- **Spawn-level timeout with process-group kill.** Network commands run via `spawnSync` with `detached: true`, so the child starts in its own POSIX process group. On timeout (90 s for fetch, 5 min for clone) the whole group receives `SIGKILL`, ensuring no ssh grandchild outlives the parent. Timeouts surface as `GitNetworkTimeoutError` with a clear, i18n'd message instead of a generic git error.

The four network-touching call sites in `worktree.ts` — existing-bare-repo fetch, bare clone, post-clone fetch, and reviewer-worktree fetch — now route through `gitNetworkExec`.

Closes #337.

## Test plan

- [x] `pnpm vitest run` — full suite passes (2167 tests).
- [x] `pnpm biome check` and `pnpm tsc --noEmit` are clean.
- [x] `gitNetworkExec` returns stdout on a successful `git` invocation.
- [x] `gitNetworkExec` throws `GitNetworkTimeoutError` when the child exceeds its timeout, and leaves no descendant processes (verified by unit tests covering a child that never exits, plus a black-hole IP clone with a 500 ms deadline).
- [x] `gitNetworkExec` throws a generic `Error` containing stderr on any other non-zero exit.
- [x] `buildGitNetworkEnv` synthesizes `GIT_SSH_COMMAND` when none of the three transports is set, preserves an existing `GIT_SSH_COMMAND`, leaves `GIT_SSH` alone (no injection) when only `GIT_SSH` is set, and skips injection when `core.sshCommand` is reported configured.
- [x] `hasCoreSshCommandConfigured` reports true for a repo whose local config sets `core.sshCommand` and false when no scope sets it (tested with `GIT_CONFIG_NOSYSTEM=1` + an empty `GIT_CONFIG_GLOBAL`).
- [x] `bootstrapRepo` and `prepareReviewerWorktree` invoke `gitNetworkExec` for all four network-touching call sites with the appropriate per-command timeout.
- [x] An interactive smoke test: pull the laptop's network during a bootstrap fetch and confirm the dispatch errors out within ~90 s instead of stalling indefinitely.
- [x] Existing single-dispatch happy paths are unchanged in latency.


sehkone:
**[Author Round 2]**

- **Fixed** — `core.sshCommand` precedence regression at `src/git-network.ts:154-160`. `buildGitNetworkEnv` now takes a `coreSshCommandConfigured` flag, and `gitNetworkExec` probes `git config --get core.sshCommand` with the spawn's `cwd` before deciding whether to inject keepalive. The probe walks the full scope chain visible from `cwd` (local + global + system inside a repo, global + system otherwise), so a `core.sshCommand` set locally in an existing bare clone is honored alongside any global/system value. When the probe reports a configured value we skip the keepalive injection entirely; the spawn-level timeout still bounds dead-TCP hangs in that case, so the structural fix is preserved. A `git config` failure is treated as "not configured" — the safer default, since the alternative is the indefinite-hang bug we're actually fixing here.

  Added two tests for the new helper `hasCoreSshCommandConfigured`: one creates a temp repo with `git config --local core.sshCommand …` and asserts the probe returns true; the other isolates from the host by setting `GIT_CONFIG_NOSYSTEM=1` and pointing `GIT_CONFIG_GLOBAL` at an empty file, and asserts the probe returns false. Also added a `buildGitNetworkEnv` test that asserts injection is skipped when `coreSshCommandConfigured=true`.

CHANGELOG and PR description updated to mention the `core.sshCommand` carve-out. Full suite passes (2167 tests); biome and tsc are clean. Pushed as a085b46.

sehkone:

